Issue #542, opened 2026-07-20
Unicode correctness: Arabic, Hindi, and the scripts that break ASCII assumptions


hjosugi:
Kofun should handle the world's writing systems correctly, not just Latin and
CJK. Arabic, Hebrew, Hindi, Thai, and the Indic scripts each break assumptions
that ASCII-shaped languages quietly bake in.

## Source text

The lexer is already UTF-8 and accepts Unicode identifiers (`let 合計 = 40 + 2`
appears in `docs/SYNTAX.md`). That is the easy half.

- **Identifier rules** should follow UAX #31 rather than an ad-hoc codepoint
  range, so what counts as an identifier is specified, not accidental.
- **Normalisation.** `é` can be one codepoint or two. Two identifiers that
  render identically must be the same identifier or a hard error — never two
  distinct bindings that look the same. NFC at lex time, with confusables
  rejected.
- **Bidi is a security issue, not a rendering one.** Trojan Source
  (CVE-2021-42574) uses RTL override characters to make source read differently
  to a human than to the compiler. Bidi control characters must be rejected.

## String semantics

The decision that matters most, and it must be made before `Text` operations
(#9) are implemented, because changing it later is extremely expensive:

- does `len(text)` count bytes, codepoints, or grapheme clusters?
- what does `text[i]` index?

`नमस्ते` is 6 codepoints but 4 user-perceived characters. Slicing by codepoint
cuts graphemes in half and produces mojibake.

Recommendation: **grapheme clusters** (UAX #29) by default, with explicit
`.bytes` and `.codepoints` views. Swift made this choice and it is the one
people stop tripping over. It costs performance, which is exactly why it should
be chosen deliberately rather than by default.

Also needed: correct case mapping (Turkish dotless ı, German ß), locale-aware
collation, and comparison that does not pretend byte order is alphabetical order.

## Output and tooling

`rt.print_str` writes UTF-8 bytes to fd 1, which stays correct. Terminal width
does not: CJK and emoji are double-width, combining marks zero-width. The
progress bar in `tui.py` assumes one column per character — already wrong, and
tracked with the TUI work.

Diagnostics should be localisable with English fallback, and column numbers
should count grapheme clusters so the caret lands under the right character.

## Acceptance criteria

- identifiers follow UAX #31, NFC-normalised, confusables rejected
- bidi control characters in source are rejected; Trojan Source is impossible
- `len` and indexing are grapheme-based, with byte and codepoint views explicit
[...]

Issue #33, opened 2026-07-19
Native bridge: call one Text-returning compiler-shaped helper


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation
- Kind: implementation
- Area: Native backend / x86-64
- Priority: P1
- Roadmap milestone: M1-bootstrap native-independence track
- Size: M
- Related parent: #32 (parallel direct-native track; not a B4/B5 prerequisite)

## Outcome

Lower one compiler-shaped Text helper through the direct x86-64 backend: Text parameters, a direct/forward helper call, Text concatenation/result, an immutable local Text binding, and `print(Text)`.

Produce deterministic static ELF without generated C, assembler, linker, or libc, and compare exact observations with an audited C reference.

This advances direct-native compiler independence. The first semantic C11 fixed point follows #618–#622/#271/#272 and does not wait for this issue.

## Canonical fixture

```kofun
fn declaration_label(kind: Text, name: Text) -> Text {
    return kind + " " + name
}

fn main() {
    let label: Text = declaration_label("fn", "main")
    print(label)
}
```

Expected stdout is `fn main\n`, stderr empty, status 0. A multibyte UTF-8 companion proves the existing byte-length Text ABI rather than ASCII-only copying. A helper defined after `main` proves declaration collection.

## Scope

- Text parameters/results for direct x86-64 functions;
- up to two Text arguments in ordinary SysV integer registers;
- frame-backed immutable Text locals;
- direct/forward calls, return, concatenation, and print;
- existing `[byte_length:i64][UTF-8 bytes]` pointer ABI;
- checked `rel32` fixups and deterministic ELF/provenance hashes;
- exact observable comparison with audited C11 reference;
- explicit rejection of unsupported bodies/signatures before artifact commit.

## Ownership/runtime contract

- Text parameters are immutable views valid for the call.
- Results are an existing immutable pointer or one newly `mmap`-allocated exact Text object.
- Concatenation checks length overflow and OOM through the existing stable runtime path.
- Process-lifetime allocation is explicit; this slice does not claim general reclamation.
- Arguments evaluate left-to-right exactly once.
- The returned pointer may be stored in one immutable caller frame slot.

## Required negative cases

- wrong arity;
- Int passed to Text parameter;
- declared/result type mismatch;
- missing Text return;
- mutable Text local;
- `List[Text]` parameter/result;
- unsupported loop/file operation;
- forced OOM with exact status/stderr and no partial stdout;
[...]

Issue #112, opened 2026-07-20
Shadowing: allow ancestor bindings and reject same-scope duplicates


hjosugi:
## Metadata

- State: ready
- Tracker role: curated
- Kind: implementation
- Area: Types/Resolution
- Priority: P1
- Roadmap milestone: M1-bootstrap
- Size: S
- Generated range: KOFUN-01521–KOFUN-01540
- Active rows: KOFUN-01523 8129c185f201; KOFUN-01525 5b0b6cad1f76; KOFUN-01526 56b89754b853; KOFUN-01529 f2c5bf36d995

## Summary

Implement the already accepted shadowing contract from spec/syntax/FOUNDATIONS_AND_CONTROL.md:

- the same scope cannot declare the same name twice;
- a child lexical scope may shadow an ancestor binding;
- uses resolve to the nearest BindingId;
- leaving the child scope restores the ancestor binding.

No new shadow keyword or warning policy is introduced.

## Prerequisite

Satisfied by #110 through #592 (`002cde7`). This issue must operate on the merged resolved scope identities, not token-text scans.

## Exact rules

- Function parameters occupy one scope; duplicate parameter names are rejected.
- The function body is a child scope. Decide in implementation docs whether a top-level body let may shadow a parameter; recommendation: allow because it is a child scope, with an optional lint outside this compiler-error issue.
- Same-scope let/let, let/pattern, and pattern/pattern collisions are errors.
- Separate if branches and separate match arms may reuse a name because neither scope contains the other.
- A nested child may shadow an outer immutable or mutable binding.
- Assignment and ownership operations apply to the nearest resolved BindingId.
- After the child exits, the ancestor binding and its state are restored; changes to a shadow binding never mutate the ancestor.
- Type differences between outer and shadow bindings are allowed.
- Generated bindings participate with hygienic identities; textual coincidence alone is not a collision.
- Top-level declaration and import namespace collisions remain #298/#290, not this local rule.

## Required fixtures

Positive:

- nested block shadows Int with Bool/Text and outer value is restored;
- then/else branches reuse the same local name;
- match arms reuse a pattern name;
- inner mutable shadow changes without changing outer binding;
- three levels resolve nearest/parent/root BindingIds correctly.

Compile-fail:

- duplicate parameter;
- duplicate let in one scope;
- let collides with an existing pattern binding in the same arm;
- duplicate names in one pattern;
- a same-scope duplicate after intervening statements;
- resolver budget boundary.

[...]

Issue #110, opened 2026-07-20
Lexical scopes: resolve outer bindings across nested blocks


hjosugi:
## Metadata

- State: ready
- Tracker role: curated
- Kind: implementation
- Area: Types/Resolution
- Priority: P0
- Roadmap milestone: M1-bootstrap
- Size: M
- Generated range: KOFUN-01501–KOFUN-01520
- Active rows: KOFUN-01503 3a8bf9f927a4; KOFUN-01505 75af07efc2de; KOFUN-01506 60249fa49779; KOFUN-01508 dd9e4bea07b8

## Summary

Replace the bounded block-local declaration scan with a lexical scope tree so nested if/match blocks can read and, when permitted, assign bindings from enclosing scopes without leaking branch-local names.

The accepted rule already exists in spec/syntax/FOUNDATIONS_AND_CONTROL.md: a binding is invisible in its own initializer, may shadow an outer binding, and may not duplicate a name in the same scope. #112 owns the focused shadowing diagnostics; this issue builds the resolver foundation.

## Current behavior

Stage 2 executes immutable/mutable Int bindings in the bounded Core. Assignment to an outer binding from inside if or match is intentionally rejected with E2S22 because the lowerer has no complete lexical resolver. Function prototypes make top-level call order irrelevant, but that declaration scan is not a local scope tree.

## Scope model

Create an explicit scope tree with stable ScopeIds for:

- function parameter scope;
- function body block;
- each if then/else block;
- each match arm, with pattern bindings visible to that arm guard and body only;
- each nested ordinary block already accepted by the frontend;
- loop header/body scopes when existing loop syntax is structurally represented, without claiming loop execution.

Each binding receives BindingId, defining ScopeId, declaration span, name, mutability, type slot, ownership mode, and initialization state.

Resolution walks the current scope then lexical parents. It never searches sibling/child scopes and never depends on source storage address or hash-map iteration order.

## Normative rules

- Parameters are visible in the function body and child scopes.
- A let initializer resolves before the new BindingId enters its scope.
- A binding is visible from the following statement through the end of its scope.
- Each if branch is a distinct child; names do not cross between branches or escape after the if.
- Each match arm is distinct; pattern bindings are visible in that arm's guard and body only.
- A mutable outer binding may be assigned in a child scope after normal type/ownership checks.
- An immutable outer binding produces the same stable assignment diagnostic as a local immutable binding.
- Reading an uninitialized/self binding is rejected at the use span with the declaration span.
[...]

PR #544, opened 2026-07-20
Expand Kofun bootstrap and native runtime

hjosugi:
## Summary

- keep the implementation Kofun-only and Python-free while rebasing onto the current native List/Text, AArch64, DWARF, C ABI, Rust-shim, and Frost build checkpoints on `main`
- extend the Kofun-authored Stage 2 checkpoint with typed Core control flow, bounded recovery, borrowed-List Copy/move diagnostics, immutable `List[Int]` literal/index lowering, and a narrow immutable Int/Bool value-position `if`
- add a small Issue #547 syntax checkpoint: `=>` arrow forms are tokenized and structurally projected deterministically, while semantic lowering rejects them explicitly
- enforce value-position `if` semantics: mandatory `else`, Bool condition, same-type Int/Bool branch unification, selected-branch-only evaluation, and exact diagnostics
- retain Kofun-authored native List/Text execution and stdlib Text/Bytes/List reference gates
- keep `.kofun` as the source extension and preserve explicit unsupported boundaries

## Why

This advances executable semantic and runtime checkpoints while keeping the bootstrap honest: every new compiler behavior has canonical Kofun source, an audited seed, deterministic fixtures, and a verification gate.

Advances #8, #9, #33, #39, #43, #47, #74, #484, #485, #542, and #547. This PR does not close their full lifecycle scopes.

## Validation

- `make clean && make verify`
- numeric corpus: 9/9
- native List corpus: 13/13
- native UTF-8 Text corpus: 9/9
- Stage 2 ownership, recovery, List, value-position-if, and arrow-token gates
- value-position-if positive output: `42\ntrue\n`
- value-position-if negative fixtures: exact `E2S15`/`E2S32`, exit 1, no C artifact
- arrow structural fixtures: expression, multi-argument, pipeline, and trailing-lambda forms round-trip; C lowering emits exact E2S10
- native x86-64, AArch64 ELF/hash/instruction, DWARF/gdb, C ABI, Rust shim, build-system, stdlib, and syntax checkpoints
- Stage 2 verified with GCC and Clang
- all JSON parsed with `jq`; all shell files passed `sh -n`; `git diff --check`
- no Python implementation, bytecode cache, legacy `.kf`, or assembly source
- AArch64 runtime differential is explicitly skipped locally because `qemu-aarch64` is unavailable

## Honest boundary

Issue #8 remains open for shared-source native/interpreter List parity, generic `List[T]` wiring, callbacks, heap growth, and ownership/GC. Issue #74 remains a narrow expression checkpoint: value blocks with declarations, value `else if`, List-valued branches, typed IR, recovery/fuzz/portability/tooling lifecycle work remain open. Issue #547 remains structural-only: lambda inference, closures, braced bodies, trailing-lambda execution, and higher-order native lowering remain open. Full semantic self-recompilation and general backend integration remain open.


hjosugi:
Added #112 lexical shadowing checkpoint in f4f3dfe: positive shadow/restore and negative duplicate fixtures, exact diagnostics, deterministic Stage2 gate. 50-issue triage comments are being posted with explicit partial boundaries.

hjosugi:
Consolidation audit (2026-07-21): this draft is now 57 commits behind `main` with 5 branch-only commits and is not safely mergeable as a unit.

The working portions have either already landed through current, narrower checkpoints (native List/Text, value-position `if`, ownership/diagnostics, enum/match, wasm/browser, package roots, and visibility) or have since changed substantially on `main`. Merging/rebasing this branch wholesale would conflict with and risk removing the modern Stage 2, wasm, LSP, stdlib, enum, and visibility work.

Remaining ideas are preserved as issue-sized follow-ups rather than discarded:
- arrow/lambda syntax and execution: continue from #547 without importing the stale frontend;
- lexical shadowing: implement only on #110's stable `ScopeId`/`BindingId` resolver, then #112 (no token-text re-resolution);
- bounded recovery and Bytes/Text reference work: port as focused slices against current `main` when their owning issues are scheduled.

Closing this superseded integration PR as part of the branch consolidation. The source branch is intentionally retained for archaeology.

Issue #546, opened 2026-07-20
Nominal heterogeneous records for structured compiler and application data


hjosugi:
## Metadata

- State: ready
- Tracker role: curated spec-design
- Kind: spec-design
- Area: Types / Syntax / Native layout
- Priority: P1
- Roadmap milestone: M2-alpha
- Size: M
- Parent UX tracker: #624

## Outcome

Define nominal heterogeneous records that are concise to construct/read, statically typed, ownership-aware, and directly laid out by native backends. A Token-shaped `Int`/`Text` record is the first executable proof.

Records are required for a maintainable structured full compiler, application domain models, and ergonomic APIs. They are **not a hard prerequisite for the first minimal string-scanning B4/B5 fixed point** tracked by #618–#622 unless the frozen compiler profile explicitly adopts them.

## Current gap

Kofun currently has no implemented general record declaration/construction/field-access path. Homogeneous `List[T]` cannot represent a token, AST node, diagnostic, symbol entry, or typed configuration with fields of different types. Tuple syntax #52 is separately specified but is not a replacement for readable named compiler data.

## Syntax decision to reconcile

Existing docs specify nominal records through `type Name = { ... }`, while an old experimental branch explored `record Name { ... }`. That branch is stale, depends on a removed implementation, and is prior art only.

The accepted design must choose one canonical form after the #624 usability corpus compares at least:

```kofun
type Token = {
    kind: TokenKind
    text: Text
    start: Int
}

let token = Token(kind: Identifier, text: "name", start: 0)
```

against any `record Token { ... }` alternative. Map literals, blocks, ADT constructors, and record construction must remain lexically/grammatically unambiguous. Only one construction spelling is accepted.

## Required semantic contract

- nominal identity and module/package visibility;
- source field order, duplicate/missing/unknown/wrong-type diagnostics;
- named construction and typed field access;
- immutable fields by default, with any field mutation explicit and ownership-checked;
- `read`/`edit`/`take` behavior for whole records and projections;
- move/borrow/drop behavior with partial moves either defined or rejected in v1;
- generic-independent first implementation, followed by generic records later;
- target-neutral semantic layout facts separated from target ABI layout;
- deterministic x86-64 and AArch64 field layout gates;
- safe pattern matching/destructuring only after ownership effects are specified.

## First implementation child

Implement one Token-shaped record with:

- `TokenKind` plus `Text` and `Int` fields;
- construction, pass, return, and typed access;
[...]

PR #559, opened 2026-07-20
Make Int functions executable across backends

### LINKED ISSUE #549: User-defined function calls do not work — the top of the dependency graph
The Kofun toolchain on `main` cannot call a user-defined function.

```kofun
fn f(n: Int) -> Int {
    return n * 2
}

fn main() {
    print(f(21))
}
```

```
$ ./bin/kofun run prog.kofun
error[R011]: unknown Core binding

$ ./bin/kofun build prog.kofun -o prog && ./prog
error[R011]: unknown Core binding
```

Both paths fail. Arithmetic inside `main` works; calling anything declared
alongside it does not.

## Why this is the top of the dependency graph

Every conformance case in the corpus is `main`-only. The two files that declare
several functions are not counterexamples: `tests/conformance/numeric/expectations.kofun`
is a manifest read by the runner rather than a program that executes, and the
syntax case is checked for structure rather than run.

`examples/fibonacci_native.kofun` fails with the same error.

This means:

- **Self-hosting (#32, #33) cannot begin.** A compiler is a few hundred
  functions calling each other. Not one of them can be called today.
- **Records (#546) do not help on their own.** A record whose fields cannot be
  read by a function is not useful; the tokeniser that proves records work has
  to call `describe(token)`.
- **Every framework issue (#24, #25, #27) is downstream.** An HTTP server is
  functions.
- **The native backend cannot be corpus-tested** beyond `main`-only cases, so
  `native-x86_64` reporting UNSUPPORTED will persist.

It is a single blocker sitting under roughly two thirds of the open work, and
it is not currently tracked as an issue.

## Scope

Whatever is needed for `fn name(params) -> T { ... }` to be callable:

- a symbol table for top-level declarations, so a name resolves to a function
  rather than being reported as an unknown Core binding
- a calling convention in the emitted code, including argument passing and
  return values
- recursion, and forward references, so declaration order does not matter
- arity and type checking at the call site

## Suggested split

These are separable and can land one at a time, each with its own corpus case:

1. call a zero-argument function, ignoring its result
2. call with arguments
3. return a value and use it
4. recursion, and mutual recursion
5. forward references
6. calls in the native backend, not only the C11 path

## Acceptance criteria

- `fn f(n: Int) -> Int { return n * 2 }` is callable from `main`
- recursion works: a `fib` case in the conformance corpus
- forward references work, so declaration order is irrelevant
- the corpus grows cases that declare more than one function, and they run under
  both backends
- `examples/fibonacci_native.kofun` runs

hjosugi:
## Summary

- extend the Kofun-authored Stage 2 checkpoint with bounded multi-function Int C11 lowering
- add direct x86-64 lowering for up to six Int arguments, results, forward and mutual recursion, guarded returns, checked arithmetic, and signed Int64 output
- register one shared five-case function corpus under both the C11 and direct-native adapters
- run the real `examples/fibonacci_native.kofun` program as a native acceptance gate
- keep unsupported AArch64/debug/type combinations explicit and artifact-free

## Why

Issue #549 is the top execution blocker: a language cannot be practically used when top-level functions cannot call each other. This PR makes the bounded Int slice executable through both active compiler paths without routing the direct-native target through C, an assembler, or a linker.

## User impact

Programs can now define and call Int functions, pass and return values, recurse, use forward references, and print arbitrary signed Int64 results. `fib(20)` produces `6765` under both the ordinary C11 CLI path and the direct static x86-64 ELF path.

## Validation

- `make verify`
- C11 function corpus: 5/5
- direct x86-64 function corpus: 5/5
- exact negative gates for unknown functions, wrong arity, arithmetic overflow, AArch64, and debug-mode boundaries
- existing numeric, List, Text, AArch64, DWARF, C ABI, Rust shim, HTTP, CLI framework, stdlib, build-system, syntax, and repository gates all pass

## Honest boundary

The function profile currently accepts Int parameters/results, up to six register arguments, comparison-guarded early returns, and direct calls. General local declarations/control flow, Text/List function signatures, AArch64 function calls, and multi-function DWARF remain open. Heterogeneous records remain the next P0 blocker in #546.

Closes #549


hjosugi:
CI is green. The two evidence-portability commits are temporarily stacked here and isolated for review in #561; after #561 lands, they can disappear from this PR on rebase.

PR #560, opened 2026-07-20
Add must-fail diagnostics and deterministic fuzz gates

### LINKED ISSUE #548: Automated test strategy: must-fail diagnostics, backend coverage, fuzzing
How production language implementations test themselves, and what Kofun should
adopt.

## What the established implementations actually do

**Rust — `compiletest`.** The suite is a corpus of programs under `tests/ui/`,
each paired with a `.stderr` file holding the *exact* expected diagnostic
output. Errors are also annotated inline:

```rust
let x: i32 = "hello";  //~ ERROR mismatched types
```

Categories are explicit: run-pass, compile-fail, and ui (compiles, but the
diagnostic text itself is the thing under test). Blessing a change regenerates
the `.stderr` files, so a diagnostic wording change shows up as a reviewable
diff rather than a silent drift.

Beyond that, **crater** rebuilds a large share of crates.io against a candidate
compiler and reports regressions. That is ecosystem-scale differential testing.

**Go — directive comments.** Files in `test/` carry a header saying how they
should be treated:

```go
// run          -- must compile and run, output compared
// errorcheck   -- must fail, with expected errors marked inline
// compile      -- must compile, not run
```

Expected errors are marked at the point they occur with `// ERROR "regexp"`.
The standard library is tested separately with ordinary `go test`.

**Java — jtreg and the TCK.** Tests carry `@test`, `@compile`, `@run` tags that
the harness interprets. Separately, the Technology Compatibility Kit is a
normative conformance suite: an implementation is only "Java" if it passes.
That is the strongest version of a spec-as-tests discipline.

**Common to all three, and the part worth copying:**

1. **Golden output.** Compile, run, compare bytes. No assertions about
   internals.
2. **Must-fail tests are first-class.** A compiler that accepts bad programs is
   as broken as one that miscompiles good ones, and this is the half most
   hobby implementations skip.
3. **Diagnostics are tested, not just exit codes.** The error code, the message,
   and the span are the product.
4. **Differential testing.** Backend against backend, or against a reference
   implementation.
5. **Fuzzing.** Csmith for C, and randomised program generation for Rust and Go.
   Finds the cases humans do not think of.
6. **Self-compilation is the integration test.** A compiler that compiles itself
   and reproduces its own binary has exercised nearly everything at once.

## Where Kofun already stands

Better than it might look. There is a conformance corpus with `# expect:`
directives, a `kofun.backend-differential/v1` contract, and bootstrap gates
that check stage reproduction. That is points 1, 4 and 6 in some form.

Measured today: 46 conformance cases; the numeric corpus reports
`9 passed; 0 failed; coverage 9/9 by c11-stage1`, with `native-x86_64` reported
`UNSUPPORTED`.

## Gaps, in the order they matter

**1. Must-fail tests with expected diagnostics.** The corpus checks programs
that work. There is no systematic suite asserting that bad programs are
rejected *with the right error*. Every diagnostic the type checker can emit --
E001 through E334, R009 through R012 -- should have a case pinning its code,
message and span. This is Rust's `.stderr` files and Go's `// ERROR`, and it is
the single biggest gap.

**2. Native backend corpus coverage.** `native-x86_64` currently reports
UNSUPPORTED for the numeric corpus. Backend-differential testing only has value
when both backends actually run the cases; an UNSUPPORTED backend proves
nothing. The coverage number should be tracked and required to rise.

**3. Fuzzing.** Two kinds worth having:
   - *grammar fuzzing* — generate random token sequences and require the parser
     to produce a diagnostic rather than crash or hang
   - *semantic fuzzing* — generate valid programs and require the interpreter
     and every backend to agree, which is Csmith's approach and finds
     miscompiles nothing else will

**4. Property tests over the law system.** The law checker already verifies
algebraic properties. The same machinery should be pointed at the language's
own primitives: integer division and modulo identities, Text concatenation
associativity, list index round-trips.

**5. Diagnostic quality as a tracked metric.** Not just presence. Does the span
point at the right token? Does the message suggest a fix? Rust treats this as a
product surface, and it is a large part of why people tolerate its strictness.

## Acceptance criteria

- a must-fail corpus where each case declares its expected error code, message
  and span, and the harness compares all three
- every diagnostic code the implementation can emit has at least one case
- backend coverage is reported as a number per corpus, and native-x86_64 is no
  longer UNSUPPORTED
- a grammar fuzzer runs in CI and no input crashes or hangs the parser
- a semantic fuzzer compares interpreter and backend output on generated
  programs
- expected-output files can be regenerated with a single command, so a
  deliberate diagnostic change is a reviewable diff

hjosugi:
Advances #548.

## Summary

- add a Stage 2 must-fail corpus with exact paired `.stderr` goldens
- require every diagnostic code in the canonical Stage 2 source to have an executable case
- validate status, code, exact message, declared span, empty internal stderr, and absence of a rejected C artifact
- add one-command golden regeneration with `tests/diagnostics/stage2/bless.sh`
- add deterministic grammar fuzzing (128 token streams, two-second watchdog)
- add deterministic arithmetic differential fuzzing (48 programs compared against an independently computed value, C11, and direct x86-64)
- put both gates in `make verify`, which is already the CI entry point

## Verification

- `make verify`
- `sh tests/diagnostics/stage2/bless.sh`
- `sh tests/diagnostics/stage2/run.sh`
- `git diff --check`

All pass locally.

## Measured coverage and remaining boundary

The Stage 2 slice is 12/12 emitted codes: 9 precise spans and 3 explicit span debts (E2S04, E2S20, E2S21).

This intentionally does not close #548 yet. Diagnostics outside Stage 2 still need their own suites, the numeric corpus still reports native-x86_64 as unsupported on main, and the semantic fuzz gate uses C11 as the active reference because the Python-free toolchain has no general interpreter yet.


hjosugi:
CI is green. The two evidence-portability commits are temporarily stacked here and isolated for review in #561; after #561 lands, they can disappear from this PR on rebase.

PR #561, opened 2026-07-20
Make CI evidence gates portable across clean toolchains

hjosugi:
## Summary

- keep the recorded HTTP benchmark commit as provenance metadata
- verify the four recorded HTTP source SHA-256 values instead of requiring historical Git objects
- make CLI runtime reproduction byte-stable within one host toolchain, not falsely identical across compiler versions
- compile and execute a program through the actively reproduced CLI runtime prefix

## Root causes

GitHub Actions uses a shallow checkout, so the HTTP gate could not resolve the benchmark's historical commit even though the artifact already contained exact content hashes.

The checked CLI runtime prefix was produced by GCC 16, while Actions uses a different compiler version. Different versions legitimately choose different instructions and can move the page-aligned config segment. Exact cross-toolchain ELF equality was therefore not a valid invariant.

## Validation

- `sh tests/http/check.sh`
- `sh framework/cli/check.sh`
- `CC=clang sh framework/cli/check.sh`
- shell syntax checks
- `git diff --check`

The HTTP gate now passes in GitHub Actions. The updated complete CI run is in progress.

### DIFF
--- a/framework/cli/README.md
+++ b/framework/cli/README.md
@@ -20,9 +20,12 @@ The checked runtime prefix is generated from the auditable `runtime.c`,
 `start.S`, and `runtime.ld` sources. `template_to_inc.c` locates the config
 anchor, removes the section table from the product, and produces
 `runtime_template.inc`. `SHA256SUMS` binds those sources to the checked prefix.
-The acceptance gate rebuilds the prefix twice, compares it byte for byte, and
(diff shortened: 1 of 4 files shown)


hjosugi:
GitHub Actions is now green on #559, #560, #561, and the dependent #562 branch. The HTTP hash gate and cross-toolchain CLI reproduction both pass on the clean Actions runner.

Issue #687, opened 2026-07-25
`/` on Int: the spec says Float, every implementation truncates — decide


hjosugi:
## Metadata

- State: ready
- Tracker role: curated
- Kind: spec-design
- Area: Semantics / Numerics
- Priority: P1
- Milestone: M1-bootstrap
- Size: S
- Generated backlog ID: none
- Fingerprint: none

## Summary

The specification says `/` on `Int` returns `Float`. Every implementation that
has an opinion truncates toward zero and returns `Int`. `Float` does not exist
anywhere in the compiler. Decide which is true and make the documents and the
backends agree.

## Evidence

Documents that say `/` yields `Float`:

- `spec/semantics.md:25`
- `docs/TYPE_SYSTEM.md:80,86`
- `docs/SYNTAX.md:142` — `7 / 2 # 3.5`

What the implementation does:

- `Float` is not a type anywhere. The lexer terminates a numeric token at `.`,
  so `3.5` lexes as three tokens, and `Float` in a signature is a hard error —
  the frozen fixture `bootstrap/selfhost/frontend/reject_parameter_type.hir`
  expects `E2S15 | type is outside the frozen profile`.
- `docs/tour/check.mjs:83-87` runs `print(7 / 2)` and asserts the output is
  `3`. It is gated by `make tour`, which `make verify` runs.
- The Stage 1 seed and the wasm32 backend both lower `/` as truncating signed
  64-bit division, each with its own zero and `INT64_MIN / -1` diagnostics.
- `bootstrap/selfhost/driver/corpus_answer.c` — the C the compiler built from
  the frozen `S` emits — implements `/` as truncating with `R010` diagnostics,
  and `corpus_answer.kofun` relies on it: `(answer - 2) / 8` is `5`.
- The direct native backends now lower `/` as truncating, to agree with that
  compiler.
- The C11 Stage 1 *emitter* does not accept `/` at all: `emit-c` on a source
  containing it fails with `E2S14`.
- `tests/conformance/numeric` has no `/` case. All nine cases are `//`, `%`,
  `+`, `-`, `*`, and unary `-`.
- `tests/kofun/arithmetic.kofun` (`# expect: 3.5`) and
  `tests/kofun/numeric.kofun` (`assert_eq(7 / 2, 3.5)`) assert the `Float`
  reading, but nothing runs them: no Makefile target or CI script references
  either file, and `bootstrap/native/README.md:252` records that the acceptance
  path they belonged to no longer exists.

[...]

Issue #666, opened 2026-07-24
Conformance: declare backend capabilities and execute numeric corpus on x86-64


hjosugi:
## Metadata

- State: ready
- Tracker role: curated
- Kind: test-quality
- Area: Testing, Security, and Performance
- Priority: P1
- Milestone: M1-bootstrap
- Size: M
- Generated backlog ID: none
- Fingerprint: none

## Summary

Replace the conformance adapters' free-form corpus claims with one checked backend-capability manifest, then make the direct x86-64 backend execute the complete nine-case numeric corpus. The result makes coverage claims deterministic and turns a supported-to-unsupported regression into a failing gate.

This is the backend-conformance child of #596. It is a scoped current-repository successor to the generated long-range catalogue in #175, not a replacement for that catalogue.

## Current behavior and evidence

On `origin/main` at `25a97098f3f06b6132ab61544261209ad84d42eb`:

- `tests/conformance/backends/*.sh` declare whitespace-separated `BACKEND_CORPORA` strings independently.
- `tests/conformance/run.sh` trusts those strings and prints corpus-level `UNSUPPORTED` without checking a central capability contract.
- `tests/conformance/numeric/expectations.kofun` names only `c11-stage1` and `wasm32-node`, so backend declarations and the executable manifest can drift.
- `./bin/kofun test tests/conformance/numeric` passes 9/9 for C11 and Wasm, but reports `UNSUPPORTED [native-x86_64] corpus numeric` and the same corpus-level result for AArch64.
- The direct native backend already executes checked arithmetic in focused native/function gates; the missing shared-corpus claim therefore needs an explicit implementation-and-evidence decision rather than another backend-specific fixture set.

## Why now

M1 requires the compiler to never silently accept unsupported backend behavior and to keep conformance evidence honest. A central manifest makes backend coverage reviewable, while x86-64 numeric execution closes the concrete native gap carried forward from #548/#596.

## Goal

Every registered backend/corpus pair has one canonical, validated state and evidence path. The x86-64 backend executes all numeric cases with byte-exact stdout, stderr, and exit status instead of reporting the corpus as unsupported.

### In scope

- Add one checked manifest, preferably `tests/conformance/capabilities.tsv`, with backend, corpus, state, and evidence-or-reason fields.
- Validate adapter identity and claims against the manifest; remove independent free-form `BACKEND_CORPORA` truth.
- Reject missing, duplicate, unknown, or contradictory backend/corpus entries.
[...]

PR #688, opened 2026-07-25
native: lower the self-host success corpus to a static ELF (#686)

### LINKED ISSUE #686: Native: widen the direct Core so the self-host success corpus reaches a native ELF
## Metadata

- State: ready
- Tracker role: curated
- Kind: implementation
- Area: Native Code Generation / Self Hosting
- Priority: P1
- Milestone: M1-bootstrap
- Size: M
- Generated backlog ID: none
- Fingerprint: none

## Summary

Widen the direct native backend far enough that the self-host driver's success
corpus — the program `A1` already compiles on the C11 path — also reaches a
static ELF through the direct backend, and turn
`bootstrap/selfhost/native/`'s negative evidence about that corpus into a
positive two-path differential.

Today the native half of the self-host track is a single expression:
`print((6 + 1) * 6)`. The self-host success corpus is refused, and the gate
records the refusal. That is honest, but it is also the narrowest possible
statement of "the self-host Core reaches native output".

On `main` after #684:

- `bootstrap/selfhost/native/corpus_core.kofun` is 37 bytes and prints `42`.
  `check-native-corpus.sh` lowers it two independent ways and compares.
- The same gate asserts that `bootstrap/selfhost/driver/corpus_answer.kofun`
  is *refused* by the native backend with `unsupported Core`, while the C11
  self-host path accepts it and prints `42 / 7 / -4 / 1 / 5`.
- `core_compiler.c` dispatches to the function profile only when
  `function_count > 1`, so every single-`main` program goes to the aggregate
  parser, which accepts exactly one `print` of a known value in `10..99`.
- The function profile accepts multiple `print` statements and full signed
  Int64 output, but rejects a `let` without a type annotation, allows only one
  local per function, requires that local to be `Text`, and has no `//`, `%`,
  or `/` at all — its product level is `*` only.

So the corpus is out of reach of both front ends for unrelated reasons, and the
function profile is much closer than the aggregate one.

## Why now

`make selfhost-native` is the only gate that ties self-hosting to direct native
production, and it currently ties it to one expression. Every piece needed to
widen it already exists and is measured: the function profile has the checked
arithmetic, the Int64 printer, the register allocator (#665), and constant-stack
returned calls (#684). What is missing is a dispatch decision, three operators,
and Int locals.

## Goal

`bootstrap/selfhost/driver/corpus_answer.kofun` compiles to a deterministic
static ELF on x86-64 and AArch64 through the direct backend, executes, and
prints exactly what the self-host C11 path prints.

### In scope

- Fallback dispatch: a single-`main` program the aggregate Core refuses is
  retried through the function profile. The aggregate path runs first and is
  untouched, so no currently accepted program changes a byte.
- `let` in the function profile: infer `Int` from the initializer when no
  annotation is written, accept `Int` locals, and allow more than one.
- `//`, `%`, and `/` in the function profile at the `*` precedence level, on
  both targets, with the floor semantics `docs/SEMANTICS.md` defines and
  `kofun_floor_div`/`kofun_floor_mod` implement.
- Runtime checks for a zero divisor and for the non-representable quotient,
  branching to a trap in the same style as the existing checked arithmetic.
- Retarget the two gates that assert the refusal this change removes.
- Widen `bootstrap/selfhost/native/` to the real success corpus, keeping the
  two-path differential and the determinism, path-independence, and ELF-shape
  checks.

### Out of scope

- Canonical `error[R010]` runtime diagnostics on the native backend. The native
  trap messages stay in their existing bounded `kofun:` form, so the numeric
  conformance corpus stays unsupported for the native adapters; that alignment
  is #666's.
- Full Int64 literals, which the function profile still caps at 65535 and which
  this corpus does not need.
- Any claim of self-application: `A1` still compiles an ordinary input, exactly
  as in the driver gate.
- The aggregate Core's own limits, and the C11 fixed point, which stays `cc`
  based and native independent.

## Contract and constraints

- `//` floors toward negative infinity; `%` takes the divisor's sign; `/`
  truncates toward zero. A zero divisor is a runtime error on all three, and
  `INT64_MIN // -1` and `INT64_MIN / -1` are runtime errors while
  `INT64_MIN % -1` is `0`.
- Divisor checks read runtime values, never only literal zeros.
- Every pinned image, digest, and byte signature in `bootstrap/native/check.sh`
  and `SHA256SUMS` stays exactly as it is.
- Two clean builds of one source stay byte-identical on both targets, and the
  image stays independent of the build directory.

## Acceptance criteria

- [ ] `corpus_answer.kofun` builds on both targets and prints `42 7 -4 1 5`,
      identical to the self-host C11 path's output.
- [ ] `//`, `%`, and `/` agree with the C11 backend on every sign combination,
      executed on both native targets.
- [ ] A zero divisor and a non-representable quotient each produce a bounded
      diagnostic and exit 1 on both targets, from a runtime value.
- [ ] No currently accepted program changes a byte; every existing pin holds.
- [ ] The two gates asserting the old refusal are retargeted, not deleted, and
      the native backend still refuses something it genuinely cannot lower.
- [ ] `make verify` passes.

## Validation

| Check | Command or artifact | Expected result |
|---|---|---|
| Self-host native parity | `make selfhost-native` | The success corpus lowers identically through both paths |
| Native regression | `sh bootstrap/native/check.sh` | Every existing pin holds; new operator gates pass |
| Function differential | `./bin/kofun test tests/conformance/functions` | C11, x86-64, and AArch64 agree |
| Full regression | `make verify` | Passes |

## Dependencies

- Blocked by: none.
- Related: #666 (canonical diagnostics and the numeric corpus, which this
  deliberately does not claim), #618/#622 (the C11 self-host fixed point, which
  this does not touch), #665, #684.

## Documentation and compatibility impact

`bootstrap/selfhost/native/README.md` currently documents the bounded surface
and the recorded refusal; that section becomes the widened surface and whatever
refusal remains. `docs/NATIVE_BACKEND.md` and `bootstrap/native/README.md` gain
the operator and dispatch contract. No source-language change is intended: these
operators already exist in the language and on other backends.

hjosugi:
Closes #686.

The native half of the self-hosting track was one expression, `print((6+1)*6)`,
and `bootstrap/selfhost/native/` recorded as *negative* evidence that the
driver's own success corpus could not reach a native image:

```kofun
fn main() {
    let six = 2 * 3
    let answer = six * 7
    print(answer)
    print(1 + 2 * 3)
    print(-7 // 2)
    print(-7 % 2)
    print((answer - 2) / 8)
}
```

It reaches one now, on x86-64 and AArch64, and prints exactly what the
self-host C11 path prints: `42 / 7 / -4 / 1 / 5`.

## Three unrelated things were in the way

**Dispatch.** Only a file declaring more than one function reached the function
profile, so every single-`main` program went to the aggregate Core — which
accepts exactly one `print` of a known value in `10..99`. A single-`main`
program the aggregate Core refuses now falls through to the function profile.
The aggregate path runs first and is untouched, so the two accepted sets are
disjoint by construction. That is checked, not asserted: rebuilding the whole
fixture, conformance, example and benchmark corpus on both targets produces
**152 byte-identical images**. A `-g` build never falls through, because `-g`
is the aggregate Core's feature. When both front ends refuse, the verdict is
the Core's, so the diagnostic keeps the stable `unsupported Core` wording —
which is also what the conformance adapters read to report an explicit skip —
and carries the function profile's more specific reason.

**Locals.** A body was allowed exactly one local and it had to be annotated
`Text`. A body may now bind as many locals as it has statements, an unannotated
local takes its initializer's type, and a written annotation is still checked
against it. Two `Text` locals in one body were a *refusal* before this change,
so they are gated positively now rather than merely no longer refused, and a
new fixture pins the annotation-mismatch rejection.

**Division.** `//`, `/`, and `%` join `*`'s precedence level. `//` floors and
`%` takes the divisor's sign, matching `docs/SEMANTICS.md` and the C11
backend's `kofun_floor_div`/`kofun_floor_mod`; `/` truncates.

Both divide instructions need guarding, for opposite reasons. x86-64's `idiv`
*faults* on a zero divisor and on the one non-representable quotient, and a
fault is a signal rather than a diagnostic. AArch64's `sdiv` never faults: it
silently returns zero for a zero divisor and `INT64_MIN` for `INT64_MIN / -1`,
so an unguarded divide there is a wrong answer instead of a crash. A `-1`
divisor is the whole of the second case and is answered without dividing — the
quotient is `-left`, whose overflow the negate reports in the overflow flag,
and the remainder is always zero, which is why `INT64_MIN % -1` is `0` rather
than an error.

The sequences were validated independently before being written in: an
exhaustive harness compared them against the C11 backend's implementation over
8,051,335 x86-64 comparisons and 844,720 AArch64 comparisons (the four
documented sign cases, every pair in `[-64,64]²`, a 29-value extreme set
crossed with itself, and 4.4M pseudorandom pairs biased toward `-1`, `0`, and
`INT64_MIN`), with zero failures, plus ten deliberate mutations to prove the
harness was sensitive rather than vacuous.

## Two limits, stated rather than papered over

- A native zero divisor reports `kofun: division by zero`, not the canonical
  `error[R010]` line, so the native adapters still do **not** claim the numeric
  conformance corpus.
- `INT64_MIN` is unreachable in this profile — literals stop at 65535 and every
  other route overflows first — so the non-representable-quotient guard is
  emitted and reviewed but executed by no fixture.

Both belong to #666.

`/` is left out of the conformance corpus deliberately: the C11 Stage 1 emitter
does not accept it, and its meaning is unsettled — the spec says it returns
`Float`, while every implementation that has one truncates and `Float` does not
exist anywhere in the compiler. The native backends match the compiler built
from the frozen `S` so the self-host differential means something, and **#687**
records the divergence instead of settling it here.

## Cleanup in the same pass

- Deleted `bootstrap/native/fixtures/core_list_{len_42,oob,positive_42}.kofun`
  — referenced by nothing since `tests/conformance/list` took over the same
  shapes on both adapters.
- Removed a stale claim that `tests/kofun` no longer exists; it does, and
  `./bin/kofun test tests/kofun` runs it.
- Factored the two per-target runtime-trap blocks into `x64_function_trap` /
  `a64_function_trap` (each now resolves its own message fixup), and the
  repeated three-way operator test into `function_divides`. Proved byte-neutral
  over 164 images across both targets.

Two orphans found and deliberately **not** deleted, because they are intent
rather than dead code — `tests/compile-fail/` (three unwired must-fail cases
for `E310`/`E323`/`E330`, the only concrete statement of three planned
ownership diagnostics) and `stdlib/tests/errno_value.kofun` (a Stage 2
value-level contract whose Stage 1 slice `errno_core.kofun` does run). Say the
word and they go.

## Validation

- `make verify` (serial) — passes
- `sh bootstrap/native/check.sh` — passes, including the new truncating-divide
  and three zero-divisor gates on x86-64, with the AArch64 images built twice
  and audited unconditionally and executed under qemu in CI
- `make selfhost-native` — both corpora lower identically through the A1/C11
  and direct-native paths on both targets
- `./bin/kofun test tests/conformance/functions` — 12/12 on c11-stage1 and
  native-x86_64 locally; native-aarch64 needs qemu and runs in CI


hjosugi:
Superseded: this branch's content landed on main as 5916fdc. Closing so the stack is clean; #691 carries the follow-on Int64 literal work rebased onto main.

PR #698, opened 2026-07-25
native: write canonical per-operator R010 diagnostics (#690)

### LINKED ISSUE #690: Native diagnostics: move message bytes to the data page, then emit canonical per-operator R010
## Metadata

- State: ready
- Tracker role: curated
- Kind: implementation
- Area: Native Code Generation / Diagnostics
- Priority: P1
- Milestone: M1-bootstrap
- Size: M
- Generated backlog ID: none
- Fingerprint: none

## Summary

Move the native backends' runtime diagnostic messages out of the one-page RX
segment, then give them the canonical per-operator `error[R010]` text. With
that, both direct native backends execute all nine `tests/conformance/numeric`
cases; today they compute every one correctly and differ only in the message.

After full Int64 literals landed, the direct x86-64 backend already produces
byte-exact stdout and exit status for all nine numeric cases. The only
remaining difference is stderr:

| Case | Native writes | Corpus requires |
|---|---|---|
| `overflow_add` | `kofun: integer overflow` | ``error[R010]: integer overflow in operator `+` `` |
| `overflow_negate` | `kofun: integer overflow` | ``error[R010]: integer overflow in unary operator `-` `` |
| `divide_by_zero_variable` | `kofun: division by zero` | ``error[R010]: operator `//` failed: division by zero`` |
| `modulo_by_zero_variable` | `kofun: division by zero` | ``error[R010]: operator `%` failed: division by zero`` |

A working implementation of the per-operator form exists on the branch
`wip-r010-traps` and passes all nine cases on `native-x86_64`. It is not
merged, because it does not fit.

## Why it does not fit, measured

`elf_image` refuses a text section over `PAGE_SIZE - TEXT_OFFSET` = **3920
bytes**, and the diagnostic message bytes are inside that text. Naming the
operator means one message per operator, so a body that can fail three ways
carries three messages instead of one shared 24-byte string. The canonical
lines are roughly 45 bytes each.

Eight stdlib fixtures go over with the per-operator form:
`clock`, `csv`, `decimal`, `json`, `random`, `regex`, `toml` checkpoints, and
`set/projection`. `set/projection` is the tightest: it emits 3897 bytes today,
**23 bytes below the ceiling**.

Shrinking the trap *code* cannot absorb this. Replacing each block's
write/exit sequence with a shared tail saves about 21 bytes per kind but the
messages add about 20 bytes per kind over the single shared string, so a
three-way body still grows by well over 100 bytes.

## Goal

Both native backends write the canonical per-operator `error[R010]` line, and
`tests/conformance/numeric` is a supported pair for `native-x86_64` and
`native-aarch64`.

### In scope

- Move diagnostic message bytes from the RX text into the existing RW data
  page. `elf_image` writes a fixed three-byte data page today
  (`0, 0, '\n'` at `DATA_ADDRESS`) with a full page of memory reserved, so the
  room is already mapped; what is missing is a way for a backend to contribute
  bytes to it and get their addresses back.
- One trap block per message the language defines, emitted only for the kinds a
  program can reach, with the shared write/exit tail so the per-kind code cost
  is a few instructions.
- The AArch64 word-alignment requirement between trap blocks (branch immediates
  count whole words, so a message of arbitrary length must not leave the next
  block unaligned).
- Register `numeric` for both native adapters and add them to
  `tests/conformance/numeric/expectations.kofun`.
- Re-pin the `kofun:`-form expectations in `bootstrap/native/check.sh`.

### Out of scope

- Changing `IMAGE_BASE`, `DATA_ADDRESS`, or the two-segment layout. The data
  page exists and has room; growing the RX segment past one page would move
  every pinned address.
- New numeric semantics or new corpus cases.
- The aggregate single-`main` Core's own diagnostics, unless the same data-page
  mechanism makes them free to move.

## Contract and constraints

- Every currently accepted program must still compile, including
  `stdlib/set/tests/projection.kofun`, and no image may exceed the RX page.
- Two clean builds stay byte-identical on both targets.
- stdout, exit status, and every existing rejection stay unchanged; only stderr
  text moves, and only to the canonical form.

## Acceptance criteria

- [ ] All nine numeric cases execute on `native-x86_64` with byte-exact
      stdout, stderr, and exit status, and on `native-aarch64` under
      `qemu-aarch64`.
- [ ] No fixture, conformance case, example, benchmark, or stdlib test exceeds
      the RX page; `set/projection` in particular still builds.
- [ ] The message bytes are in the data segment and the text section shrinks
      for programs that carry diagnostics.
- [ ] `make verify` passes.

## Validation

|---|---|---|
| Numeric differential | `./bin/kofun test tests/conformance/numeric` | 9/9 on both native adapters |
| Native regression | `sh bootstrap/native/check.sh` | Every pin holds |
| Page ceiling | build every `stdlib/*/tests/*.kofun` | none refused |
| Full regression | `make verify` | Passes |

## Dependencies

- Blocked by: none. Int64 literals landed separately.
- Related: #666 (the capability manifest half, which this completes the
  execution side of), #686, #687.

## Notes

`wip-r010-traps` has the per-operator trap table, the emitter changes for both
targets, the adapter registration, and the re-pinned gate expectations. It is
correct and passes 9/9; it just needs the messages moved off the RX page first.

hjosugi:
Closes #690. Completes the execution half of #666.

Both direct native backends now write the same `error[R010]` line the C11 and wasm32 backends write, naming the operator, and execute **all nine `tests/conformance/numeric` cases** with byte-exact stdout, stderr, and exit status. They already agreed on every value; only the message differed.

## Why the earlier attempt did not fit, and what changed

Naming the operator means one message per operator. The first attempt kept those messages in the one-page RX segment and eight stdlib fixtures went over the 3920-byte ceiling — `set/projection` sits **23 bytes** under it. Two changes make it fit, and both are things the emitter should have been doing anyway.

**The messages are data.** They move to the RW page, which has always started with three fixed bytes and reserved a whole page for them. A backend's blob is appended after those three, so every address baked into existing code is unchanged and a backend with nothing to add emits exactly what it did before. Naming the operator now costs the RX segment nothing.

**The write-and-exit half of a trap is identical for every kind.** It is emitted once, and each kind reduces to loading its message and branching to it — three instructions, against a whole sequence per kind. Only the kinds a program can actually reach are emitted, so a body with no arithmetic carries no trap at all.

Checked against main: **the set of sources that exceed a page limit is exactly the same before and after**, on both targets.

## Also restored

`5916fdc` landed #686 partially — the trap factoring and three superseded `core_list_*` fixtures were dropped. Both are back, together with `function_divides`.

## Validation

- `make verify` (serial) — passes
- `./bin/kofun test tests/conformance/numeric` — 9/9 on `c11-stage1` and `native-x86_64`; `native-aarch64` runs the same corpus under qemu in CI
- `sh bootstrap/native/check.sh` — passes with every diagnostic expectation re-pinned to its per-operator form
- Page-limit sweep over 198 sources on both targets — no regression


hjosugi:
Superseded by 6f21b39 on main, which lands the same design independently: message bytes in the RW page via `elf_image(..., data)`, per-operator `error[R010]` traps, and `native-x86_64` executing 9/9 numeric. Verified on main: `coverage: 9/9 cases executed by native-x86_64`.

Closing rather than resolving a 32-conflict rebase to re-land duplicate work. The branch `native-r010-data-page-690` stays pushed in case any detail here is worth diffing against — notably the shared trap tail (one write-and-exit sequence, each kind reduced to three instructions) and the page-limit sweep showing the over-page set is unchanged before and after.

Issue #703, opened 2026-07-25
Lambdas: bind parameters, then accept the arrow form decided in #547


hjosugi:
## Metadata

- State: ready
- Tracker role: curated
- Kind: implementation
- Area: Syntax / Parser / Inference
- Priority: P1
- Decision owner: #547 (syntax), #305 (parameter inference)

## Summary

Make lambdas work. The syntax was decided in #547; this is the implementation, and it is larger than that issue's title suggests because **no lambda binds its parameter today**.

## Current behavior and evidence

The parser accepts the lambda shape and never binds the parameter, so any lambda whose body mentions it is rejected:

```sh
printf 'fn main() {\n    let f = fn(x: Int) => x * 2\n    print(f(3))\n}\n' > /tmp/l.kofun
./bin/kofun check /tmp/l.kofun
# error[E2S35]: unknown lexical binding `x` at byte 27
```

The same `E2S35` appears for `map(xs, fn(x) => x * 2)`, `map(xs, x => x * 2)` and `map(xs, (x: Int) => x * 2)` — always at the body's `x`, never at the syntax. A named function in the same position compiles and runs.

The form is documented in `docs/SYNTAX.md:61`, present in `editor/tree-sitter-kofun/grammar.js:431`, and implemented in the playground's TypeScript interpreter (`app/kofun-runtime.ts:459-465`), which is the only place it executes.

Three shipped examples use it and **none compiles**, each stopping earlier at a `law` declaration, and none is referenced by any runner:

| Example | `bin/kofun check` |
|---|---|
| `examples/lawful_list_monad.kofun` | `error[E2S02]` at byte 491 |
| `examples/broken_list_monad.kofun` | `error[E2S02]` at byte 310 |
| `examples/proven_optional_bool_monad.kofun` | `error[E2S02]` at byte 219 |

So there is no executable evidence for lambdas in either direction, and the documentation currently describes a feature the compiler does not provide.

### In scope

- Bind lambda parameters in their body's scope, so `fn(x: Int) => x * 2` type-checks and runs.
- Accept the forms #547 decided: `x => expr` (single unannotated parameter, no parentheses) and `(x, y) => expr`. Reject `x: Int => expr` with a diagnostic that suggests `(x: Int) => expr`.
- Reject `it` and `_` as implicit parameter names, per #547.
- Lower lambdas so they execute, with the C11 backend as the reference observation.
[...]

PR #706, opened 2026-07-25
stage2: bind arrow-lambda parameters in the scope-HIR (#703)

hjosugi:
`fn(x: Int) => x * 2` reported `error[E2S35]: unknown lexical binding `x``
at the parameter's own declaration. That diagnostic was false: `x` is not an
unknown binding, it is a binding the scope-HIR never made. Pass 4 of
`build_scope_hir_mode` is a free-identifier scanner, and it had no idea an
anonymous `fn` declares anything, so it treated the parameter name as a
reference to something that did not exist.

Stage 2 has no lambda code path at all. Of the ten `"=>"` sites in
`compiler.kofun`, ten are `match`-arm machinery or the lexer's two-character
pairing; `lambda`, `closure` and `Fn[` appear nowhere. So this is the first
slice: make the *scope* correct, and leave lowering to report honestly that it
is unsupported.

An arrow lambda now contributes a `lambda-parameters` scope keyed by its `(`,
spanning to the end of its arrow body, parented to the enclosing block scope.
Its parameters become bindings in that scope, and the identifier pass resolves
through it. A named function has an identifier after `fn`, so requiring `(`
there is what separates `fn(x) => x` from `fn name(x) { }`.

The body is delimited by `expression_end`, the same expression grammar the Core
lowers. A body this Core cannot parse yields -1 and the lambda contributes no
scope, so nothing that used to be reported stops being reported. That is also
why nested lambdas are out: `expression_end` cannot parse an inner `fn`, so the
outer one is not delimited and neither gets a scope.

Measured, on the four cases that distinguish a real fix from a suppressed
diagnostic:

| Source | Before | After |
|---|---|---|
| `fn(x: Int) => x * 2`, then `f(3)` | `E2S35` on `x` | resolves; `E2S16: unknown Core function `f`` |
| `let n = 10` then `fn(x: Int) => x + n` | `E2S35` on `x` | resolves, capture included; `E2S12` from lowering |
| `fn(x: Int) => y + 1` | `E2S35` on `x` | `E2S35` on `y` — still rejected, nothing suppressed |
| `fn(x: Int, x: Int) => x + 1` | not detected | `E2S47: duplicate binding `x`` |

The third row is the one that matters: the change resolves declarations, it
does not make the scanner lenient.

`tests/conformance/syntax/issues_35_47/run.sh` still asserts
`expect_stage2_unsupported` on `unsupported_lambda.kofun`, and it still holds —
exit 1, no `.c` emitted, `^error[E2S` on stdout. Only the message improved, from
a false claim about a binding to `E2S12: invalid Int expression`, which is true.

Gated by `tests/diagnostics/stage2/e2s47_duplicate_lambda_parameter`, which can
only pass if the lambda parameter scope exists.

`compiler.c` is a hand-audited transliteration, not a generated artifact, so the
same change is written twice and both `SHA256SUMS` entries are regenerated.

`make -j1 verify` passes.


hjosugi:
Superseded by #707, which contains this commit as its first of four and fixes the CI failure. That failure was a hard-coded corpus size in `tests/typed-sidecar/stage2-events.sh` — `expected all 33 Stage 2 diagnostic fixtures, saw 34` — which no message connects to adding a diagnostic fixture. Local `make -j1 verify` passed here only because it ran before the fixture was added.

Issue #721, opened 2026-07-26
Decimal slice 2: arbitrary-precision runtime representation, canonicalization, and the versioned resource profile


hjosugi:
Child of #710, slice 2 of its ordered list: *arbitrary-precision runtime
representation, canonicalization, versioned resource profile, and stable
resource diagnostics.*

#710 is the curated tracker and states it is not ready for a single
implementation PR. This is the second independently reviewable slice split out
of it. Slice 1 landed as #717 (PR #718) and its follow-ups landed with #703 and
#552.

## Current behavior and evidence

Slice 1 delivers the token contract and nothing beyond it. A well-formed
Decimal or Float literal that reaches lowering is now refused by name:

~~~sh
printf 'fn main() {\n    print(1.5)\n}\n' > /tmp/d.kofun
./bin/kofun check /tmp/d.kofun
# error[E2S99]: Decimal literal at byte 22 has no runtime representation yet (#710 slice 2)
~~~

`E2S99` exists specifically to point at this issue, and it is the thing this
slice removes. Until then there is no value to lower a literal *to*: the token
retains its digits and the compiler stops.

The bounded `stdlib/decimal` checkpoint from #545 is **not** this. It proves
exact arithmetic over a signed Int64 significand with scale `0 .. 18`. Frozen
decision 1 requires an arbitrary-precision binary significand, and decision 6
says the checkpoint's runtime scale field is not static scale safety.

## Goal

A canonical `(significand, scale)` value with an arbitrary-precision binary
significand, plus the versioned resource profile that bounds it. **No typing
and no arithmetic** — those are slices 3 and 4. A Decimal that reaches an
operator must still fail explicitly, naming its successor slice.

### In scope

- The canonical `(significand, scale)` representation, arbitrary precision.
- Construction from a slice-1 token: remove underscores, build the integer
  significand, apply the exponent to scale, normalize. No host `double` on the
  path, per `docs/DECIMAL.md`.
- Canonicalization, so that two constructions of one value are
  indistinguishable. Small `i64`/`i128` storage is permitted **only** as an
  observationally invisible optimization (frozen decision 1).
- A **versioned** resource profile, cross-backend observable (frozen decision
  8). Exceeding a limit fails explicitly and never clamps, wraps, rounds,
  lowers through Float, or silently changes representation.
- Stable, source-located diagnostics for each resource failure.
- Binary64 `Float` construction for the `f64` suffix, kept distinct.

### Out of scope

[...]

Issue #717, opened 2026-07-26
Decimal slice 1: numeric token contract and parser, with byte-exact literal and malformed-token fixtures


hjosugi:
Child of #710, slice 1 of its ordered list: *numeric token contract and parser,
with byte-exact Decimal/`f64` literal and malformed-token fixtures.*

#710 is the curated tracker and states it is not ready for a single
implementation PR. This is the first independently reviewable slice split out
of it, sized M.

## Current behavior and evidence

Measured on `main` at `17075de`:

~~~sh
printf 'fn main() {\n    print(1.5)\n}\n' > /tmp/d.kofun
./bin/kofun check /tmp/d.kofun
# error[E2S13]: expected `)` at byte 23

printf 'fn main() {\n    let r = 1..2\n    print(3)\n}\n' > /tmp/r.kofun
./bin/kofun check /tmp/r.kofun
# error[E2S10]: unsupported Core statement at byte 25
~~~

No lexer in the tree produces a fractional, scientific, or `f64` token today.
The `.` simply terminates the integer token, which is why the diagnostic points
at the character after `1` rather than at the literal. So there is no token to
type, lower, or preserve digits from — every later slice depends on this one.

## Goal

Produce and preserve numeric tokens exactly as `docs/DECIMAL.md` specifies,
**without** yet giving them a type or a runtime. A `Decimal` or `Float` literal
that reaches lowering must fail explicitly, naming the unimplemented slice —
never lower through `Int`, never round, never convert through a host `double`.

### In scope

- Lex the normative grammar from `docs/DECIMAL.md:97-107`: `digits` with
  `_` separators between digits, `exponent`, `decimal`, and `float64`.
- Maximal munch **with the range exception**, exactly as specified:

  | Source | Tokens |
  |---|---|
  | `1.0` | one Decimal token |
  | `1e3` | one Decimal token |
  | `1f64` | one Float token |
  | `1..2` | `Int(1)`, `range(..)`, `Int(2)` |

- The token retains the **original digit sequence** and the positions of the
  decimal point, exponent, and suffix. No host `double` on the path.
- Lexical errors, each with a stable source-located diagnostic and no artifact:
  `1.`, `.5`, malformed exponents (`1e`, `1e+`, `1.0e`), and `_` outside a
  position between two digits (`_1`, `1_`, `1._0`, `1.0_`).
- `f64` is part of the numeric token, not an identifier — `42f64` is one token,
  and `42 f64` is not.
[...]

Issue #710, opened 2026-07-26
[Umbrella] Deliver compiler-native Decimal, Float, and Fixed across declared backends


hjosugi:
## Metadata

- State: blocked
- Tracker role: planning migration umbrella
- Kind: planning
- Area: syntax / type system / compiler / code generation
- Priority: P1
- Size: L (multi-backend program delivered through bounded children)
- Parent planning umbrella: #482
- Normative design: `docs/DECIMAL.md`
- Blocked by: #725, #1402, #1403, #1404, #1405
- Audited against: `origin/main@78b409c7ba69ea8fe21bdaea911197430f0e05b8`

## Goal

Deliver the accepted arbitrary-precision Decimal, explicit binary64 Float, and static-scale Fixed semantics truthfully on every backend/profile that the conformance matrix declares, without borrowing another backend's compiler or silently narrowing values.

## Delivered semantic foundation

The original semantic slices are complete on the bounded C11 Stage 2 profile:

| Slice | Owner | Status/evidence |
| --- | --- | --- |
| exact numeric token contract | #717 | complete |
| arbitrary-precision carrier/resource profile v1 | #721 | complete |
| same-type operators/explicit conversions | #722 | complete |
| exact arithmetic and checked exact division | #723 | complete |
| five-mode rounding, rounded division, parse/format, ledger example | #724 | complete |
| Decimal-backed Fixed static scale | #725 | open planning umbrella; children #1249-#1253 |
| law and bounded differential evidence | #726 | complete for the declared C11 Stage 2 profile |

`task decimal` and `decimal-arithmetic` prove `0.1 + 0.2 == 0.3`, canonical arbitrary precision, exact/inexact/zero division, all rounding modes, resource codes D001-D007, Float contrast, sanitizers, and the ledger/tax bytes on C11 Stage 2.

## Corrected backend evidence

The matrix truthfully leaves c11-stage1, native x86-64, native AArch64, and wasm unsupported. Two prerequisites must precede implementation:

- #1254 (closed): the adapter called `c11-stage1` currently invokes `bin/kofun emit-c`, which tries Stage 2 first. It must invoke only Stage 1 before any Stage 1 capability result is credible.
- #1255 (closed; the decision is recorded as accepted RFC-0015): direct native emits linker-free standalone ELF, while wasm splits bare numeric `wasm32` from arena-backed `wasm32-hostabi1`. The Decimal target identity/ABI cannot be inferred without changing one of those contracts.

[...]

Issue #551, opened 2026-07-20
Law declarations: make Monad a library law, not a compiler keyword


hjosugi:
## Metadata

- State: ready
- Tracker role: curated spec-design
- Kind: spec-design
- Area: Laws / Type system / Syntax
- Priority: P1
- Roadmap milestone: M2-alpha
- Size: M
- Parent UX tracker: #624

## Measured current state

The repository contains `law monad` examples, law-evidence schema/artifacts, and design documentation, but the active Python-free Stage 1/2 compiler does **not** parse or check them.

Measured on current main:

```text
kofun check examples/lawful_list_monad.kofun
  -> error[E2S02]: expected top-level `fn` or `type`

kofun check examples/broken_list_monad.kofun
  -> the same parser rejection; no counterexample is computed
```

The former Python checker referenced by the old issue body has been removed. Static JSON evidence is historical/design evidence, not a currently executable compiler-integrated law gate. README/docs must remain honest until an executable Kofun implementation lands.

## Outcome

Define ordinary source-level algebraic law declarations whose equations are data checked by a generic finite-model engine. Monad, Functor, Applicative, Semigroup, and Monoid become library declarations/instances rather than hard-coded compiler families or permanent keywords.

The beginner experience does not require Monad terminology. Ordinary errors use `Result`, match, pipelines, and the sequencing decision in #626. Law declarations are an advanced assurance feature.

## Concrete-first v1

Do not block executable laws on full higher-kinded types. V1 should support concrete finite instantiations such as:

- `Optional[Bool]` Monad;
- `List[Int]` over a declared bounded carrier;
- `Int` monoid under a bounded model;
- one user-declared deliberately broken law.

The design may describe future generic `Monad[M[_]]`, but `proven-finite` is granted only when the compiler certifies a complete finite carrier and complete enumerated function space. Bounded samples remain `bounded-exhaustive`; neither becomes a universal proof.

### Declaration and instance grammar

Choose one canonical grammar for:

- typed operation requirements;
- quantified variables and their domains;
- named equations;
- an implementation/instance supplying operations;
- a model/evidence request naming values/functions/equality/budget.

`monad` must not remain a special token. `Monad` may be an ordinary declaration name.

### Equality

Law equality is typed observational equality over the declared finite model. Function equality is never assumed decidable; functions are compared extensionally only over a compiler-certified finite input space, otherwise assurance is bounded.

### Evaluation safety

[...]

PR #727, opened 2026-07-26
stage2: pass functions as arguments, and diagnose the notations that were removed (#703, #552, #717)

### LINKED ISSUE #552: Function type notation is inconsistent: Fn[A,B] vs -> — decide currying first
Kofun currently uses two unrelated notations for the same concept.

A function **type** is written with brackets:

```kofun
fn list_bind(values: List[Int], next: Fn[Int, List[Int]]) -> List[Int]
```

A function **declaration's** return type is written with an arrow:

```kofun
                                                          -> List[Int]
```

Both describe "a function from Int to List[Int]", and they look nothing alike.
That is the actual problem — not whether `->` is worth two characters.

## Answering the question directly: is there evidence for `->` over Go's style?

**No. There is no empirical evidence either way.** No controlled study shows
`fn f(a: Int) -> Int` is more or less readable than `func f(a int) int`. This
is a consistency and taste decision, and it is more honest to say so than to
construct a justification after the fact.

The arguments that are real:

**For `->`:**
- It marks unambiguously where the return type begins. With return types
  optional — as they already are in Kofun — a parser reading
  `fn add(a: Int, b: Int) Int {` must decide whether `Int` opens the body or is
  a type. Go avoids this because its grammar is tighter about where types can
  appear. Workable either way, but `->` is free of the question.
- It is scannable: `-> ` is greppable in a way a bare type name is not.

**Against `->`:**
- Two characters on every annotated signature.
- Return types are already inferred when omitted, so the annotated case is the
  minority, which weakens the ergonomic argument on both sides.

Neither list is decisive. What *is* decisive is the third question.

## Currying and arrow functions: `->` earns its keep only if function types use it

Measured today:

```kofun
let inc = add(1)      // error[E321]: `add` expects 2 arguments, found 1
```

Partial application does not work. Closures do:

```kofun
fn adder(n: Int) -> Fn[Int, Int] {
    return fn(x) => x + n
}
```

So currying is not a thing Kofun does, and `Fn[A, B]` is the notation for
function types. Under those facts, `->` in a declaration buys almost nothing:
it does not compose with `Fn[...]`, and there is no curried form for it to make
readable.

Arrows only pay for themselves when the function type is *itself* an arrow:

```
Int -> Int -> Int          reads as: takes Int, returns (Int -> Int)
Fn[Int, Fn[Int, Int]]      the same thing, unreadable
```

That is the whole case for arrows, and it depends entirely on wanting currying.

## Two coherent choices; the current state is neither

**Package A — arrows throughout.** Function types become `Int -> Int`,
declarations keep `-> T`, currying becomes natural to write and read. ML,
Haskell, F#. Costs: currying has to actually work, and the parser must handle
right-associative arrows.

```kofun
fn list_bind(values: List[Int], next: Int -> List[Int]) -> List[Int]
```

**Package B — brackets throughout, Go-style declarations.** Function types stay
`Fn[Int, Int]`, declarations drop the arrow and put the return type after the
parameters. Consistent, lighter, and honest about not being a curried language.

```kofun
fn list_bind(values: List[Int], next: Fn[Int, List[Int]]) List[Int]
```

**Recommendation: decide currying first, then let syntax follow.**

- If currying and partial application are wanted, take Package A. The arrow
  stops being decoration and starts carrying meaning.
- If they are not — and nothing in `docs/LANGUAGE_VISION.md` asks for them —
  take Package B. It is smaller, and `Fn[...]` already exists.

Given the stated goal of being approachable to TypeScript and Python users,
neither of which curries by default, Package B is the more consistent fit. But
this is a genuine fork and should be decided deliberately rather than by
inheriting Rust's syntax without Rust's reasons.

## Acceptance criteria

- a decision recorded on whether currying and partial application are supported
- one notation for function types, used in both positions
- `docs/NAMING.md` or `docs/SYNTAX.md` records the decision and the rejected
  alternative, including that the readability question has no empirical answer
- if Package A: `add(1)` returns a function, and a test covers it
- if Package B: `->` is removed from declarations, and the grammar in `spec/`
  is updated with it

### LINKED ISSUE #703: Lambdas: bind parameters, then accept the arrow form decided in #547

hjosugi:
Closes #703. Closes #552.

Four acceptance criteria were written down and not met. This meets all four,
each with an executable gate.

## #703 criterion 3 — a lambda passed as an argument

Nothing could be passed. Measured on `main`, all three pieces were missing:

~~~
apply(x => x * 2, 3)             E2S17: expects 2 arguments, got -1
apply(dbl, 3)                    E2S35: unknown lexical binding `dbl`
fn apply(g: Int -> Int, v: Int)  E2S15: expected Core parameter name
~~~

A callable parameter now lowers to a plain C function pointer, which works
because every value in this Core is an `int64_t` and needs no environment.
That is also what makes the boundary honest: a **capturing** lambda is lifted
with its captures as trailing parameters, so its address is not the declared
callable. Passing one is refused as `E2S96` rather than lowered into something
whose observations would not match — closure conversion is #116/#370, which
#703 puts out of scope.

The gate computes each value **twice**, once through a named function passed as
an argument and once through a lambda at the call site, so identical
observations are what is measured rather than "a lambda argument compiles".

Two things fell out of it, both measured rather than reasoned:

- the scope-HIR parameter walk stepped over one type token, so a multi-token
  type desynchronised it and every parameter after a callable went unbound,
  surfacing as `unknown lexical binding 'left'`;
- a bare function name is a value only in argument position. Allowing it
  everywhere let `double + 1` reach the Int expression grammar and surface as a
  **C** type error instead of a Kofun diagnostic.

## #703 criterion 4 — a shipped example that runs

`examples/lambdas.kofun`, gated through the public `kofun run`, so what is
pinned is the output a reader copying it gets. The three existing lambda
examples still stop at `law` (#551) and are not that evidence.

## #552 — the notation, and the rewrite it promised

The decision was already recorded in four normative documents by `d5910f0`;
the issue was open because three things it required were missing.

- **tree-sitter could not parse a callable type at all.** Not theoretical:
  `examples/lambdas.kofun` failed the repository corpus until the grammar
  learned it. Newlines may not precede the `->` — they are significant, and
  letting a type consume them made `extern "C" fn add(...) -> CInt` swallow its
  terminator and fail at EOF.
- **three shipped examples still used `Fn[...]`**, which the spec calls
  invalid. Migrated, with a gate that fails if one returns.
- **the targeted rewrite all four documents mandate did not exist.** `E2S97`
  computes it: `Fn[Int, List[Int]]` → `Int -> List[Int]`, `Fn[A, B, R]` →
  `(A, B) -> R`, `Fn[R]` → `() -> R`. `Fn` stays an ordinary identifier.

`docs/NAMING.md` also now records the rejected alternatives and states plainly
that the readability question has no empirical answer, which #552 asked for by
name.

## #717 follow-ups

`1.`, `.5`, `1e`, `1e+`, `1_`, `1._0`, `1.0_`, `1.0e` were not lexical errors.
`1_` was worse: it **compiled**, as `1`, and wrote an artifact. All are now
`E2S98`, reported before a token tape exists.

They are diagnosed from the token sequence rather than by tightening
`token_end`, because separating the range operator from a fraction inside the
scanner needs exactly the lookahead the range exception exists to avoid — the
concern recorded on #717 was right. After lexing, `..` and a lone `.` are
already distinct tokens, so the rule is exact. `1..2` is gated beside them.

`_1` is deliberately **not** included, departing from #717's acceptance list: it
is a well-formed identifier, not a numeric literal, and banning an identifier
spelling on the strength of a rule about numbers is the wrong trade. The
reasoning is recorded in `docs/DECIMAL.md`.

Separately, a **well-formed** Decimal or Float literal reaching lowering reused
`E2S12`, "invalid Int expression" — which says the literal is wrong when what is
true is that the compiler is unfinished. Now `E2S99`, naming #710 slice 2
(#721), which removes it.

## Also

`compiler.kofun` was missing the bare-lambda fix `compiler.c` got in #720. The
`.kofun` copy took `balanced_end` from a non-`(` token, ran on to a later
unbalanced `)`, and would have bound `=>` as a parameter name. Nothing executes
the `.kofun`, so no gate caught the divergence.

## Verification

`make verify` passes, **808 PASS** (baseline on `main`: 798).

Both hard-coded corpus counts in `tests/typed-sidecar/stage2-events.sh` were
re-measured by running the gate, not predicted: 35 → 39 fixtures and 134 → 138
repository error companions. `E2S98` is registered alongside `E2S01` as a
pre-token code, since it aborts in the lexer and emits no event stream.

One caveat worth recording: a first verify run failed in the `native` target
with a missing `.elf`. That was the known shared-`build/` race — I had run
`bootstrap/native/check.sh` by hand while the parallel verify was in flight. A
clean re-run is green.


hjosugi:
CI green on `836e825` ([run 30194930551](https://github.com/hjosugi/kofun/actions/runs/30194930551)).

The two earlier red runs were not the compiler change. `d1a5580` and `19b4ac9` both died in the `syntax` target with `Syntax error: ")" unexpected` — unescaped backticks in one new failure message, which are command substitution and which dash rejects while *parsing* the gate script. The whole gate never ran; the log showed no failures because nothing executed.

Worth recording why local `make verify` reported 808 PASS three times and never caught it: this machine's `/bin/sh` is bash, and the line sat on the right of `||` where the `grep` succeeded, so the string was never evaluated. Neither shell catches it by running the happy path — only a parse check does.

Fixed in `836e825`; every other backtick in that file was already escaped or inside single quotes.

Issue #747, opened 2026-07-26
selfhost: accept the Text type, literals and concatenation in S (#622)


hjosugi:
## Metadata

- State: blocked
- Tracker role: curated
- Kind: implementation
- Area: Self Hosting / Stage 1 seed
- Priority: P0
- Roadmap milestone: M1-bootstrap
- Size: M
- Parent: #622

## Outcome

`S` accepts the `Text` type: literals with escapes, `+` concatenation, `==`
comparison, `Text` parameters, results and bindings.

## Why this changes the emitted program

Every value the Core emits today is an `int64_t` held in a C local. `Text`
introduces a second representation and the first heap allocation in an emitted
program, so the runtime shim the emitted C carries has to grow with it —
`kofun_rt_text_concat`, `kofun_rt_text_equal` and their allocation discipline —
rather than the emitter alone.

## Scope

- `"..."` literals, including `\n`, `\"` and `\\`;
- `+` on two `Text` values, refused when mixed with `Int`;
- `==` and `!=` on `Text`;
- `Text` in parameter, result and binding position;
- `print` of a `Text`.

## Acceptance criteria

- Both seeds accept the form and refuse `Text`/`Int` mixing in both orders.
- `corpus_answer.c` is unchanged.
- A `Text` program emits C that compiles under the gate's flags with no leak of
  the user's source into the emitted program beyond its own literals.
- `bootstrap/stage1/compiler.c` is hand-ported and differentials byte for byte.
- `make selfhost-profile bootstrap stage2 check` pass.

## Dependencies

Blocked by the comparison/`Bool` slice for `==`.

hjosugi:
`blocked` → `ready`。本文が名指ししている唯一のブロッカーが解消されています。

> Blocked by the comparison/`Bool` slice for `==`.

その slice は #744 で、2026-07-27 14:15:04Z に completed で close 済み、`main` の `ecbb0e3`（"selfhost: lower comparisons and Bool in Stage 1"）です。`==` `!=` `<` `<=` `>` `>=` と `Bool` 値、`!` / `&&` / `||` が `S` に入っており、`Bool` を算術に混ぜることは拒否されます。

#622 チェーンの他3件は blocked のままです（それぞれのブロッカーが未解決）:

- #748 ← この issue（`Text` slice）
- #749 ← `Text` と `List[Text]` の両方
- #751 ← #622 の他のすべての言語 slice

[...]

PR #776, opened 2026-07-28
codegen: give every native target one shared register allocator (#770)

### LINKED ISSUE #770: Codegen contract: stop re-implementing register allocation and lowering per target
## Metadata

- State: ready
- Tracker role: curated
- Kind: spec-design
- Area: Codegen
- Priority: P1
- Milestone: M1-bootstrap
- Size: M
- Parent: #554

## Summary

Give the direct native backends one explicit contract, so a target supplies the
facts only it knows — register file, calling convention, instruction encoding —
and everything derivable from those facts is written once. #554 already decided
"codegen placed behind an interface, so a second backend stays possible", as a
single acceptance bullet with no design behind it. This is that design.

## Measured current duplication

`bootstrap/native/core_compiler.c` carries **59 `x64_*` and 92 `a64_*` symbols,
of which 29 are the same concept spelled twice**. They are not all the same kind
of duplicate, and the difference decides what may be shared.

Normalising the `X64_`/`A64_` prefixes and diffing:

| Layer | Functions | x64 / a64 lines | Diff after renaming |
|---|---|---|---|
| Register allocation | `take_scratch_register` | 9 / 9 | **0** |
| | `take_value_register` | 13 / 13 | **0** |
| | `take_eval_register` | 10 / 10 | **0** |
| | `take_call_safe_register` | 9 / 9 | **0** |
| Operand materialisation | `value_operand` | 12 / 12 | 6 |
| | `eval_operand` | 17 / 17 | 6 |
| | `move` | 16 / 18 | 16 |
| Lowering | `function_compare` | 18 / 23 | 31 |
| | `function_divide` | 97 / 60 | 129 |
| | `function_expression` | 201 / 190 | 239 |

The first four are **41 lines of identical register allocation, written twice**.
The only differences are the type and constant names: `X64RegisterFile` versus
`A64RegisterFile`, `X64_SCRATCH_REGISTERS` versus `A64_SCRATCH_REGISTERS`,
`X64_NO_REGISTER` versus `A64_NO_REGISTER`.

The remaining 19 pairs are `call_safe_registers`, `expression`,
`function_call`, `function_declaration`, `function_epilogue`,
`function_epilogue_block`, `function_guard_inverse`, `function_layout`,
`function_print_runtime`, `function_print_text_runtime`, `function_program`,
`function_tail_call`, `function_trap_runtime`, `function_trap_stub`,
`register_operand`, `runtime`, `scratch_registers`, `slot_operand`, `text`.

## Why now

Every queued codegen item is currently priced per target, and the multiplier is
about to grow:

- **#278** ARM Cortex-M (thumb2) would be a third target;
- **#26** extends wasm32;
- **#710** is explicitly "across all backends";
- **#158–#176** are 19 separate codegen features — SIMD intrinsics, atomics,
  thread-local storage, unwind tables, stack maps, position-independent code,
  Windows calling convention, sanitizer and coverage instrumentation,
  floating-point and overflow modes, tail calls, static and dynamic linking.

At the present structure that is 19 features times N targets of hand-written
lowering, and the 29 pairs show the shape it takes.

## What TB does, concretely

[TB (Tilde Backend)](https://github.com/RealNeGate/Cuik/tree/master/tb), the
backend of [Cuik](https://github.com/RealNeGate/Cuik), is the closest useful
reference: a from-scratch, non-LLVM backend with four targets (`x64/`,
`aarch64/`, `mips/`, `wasm/`) that are roughly equal in size, which is the
property to copy.

Its per-target contract lives in `tb/codegen.h` as callbacks on one `Ctx`
struct. A target provides:

| Hook | What the target declares |
|---|---|
| `MatchRule` | bottom-up rewrite of generic nodes into machine-friendly forms |
| `NodeConstraint` | register-class constraints and input/output requirements per node |
| `NodeConstraintKill` | which registers an operation clobbers |
| `TB_2Addr` | shared input/output edges for two-address operations, as x86 needs |
| `NodeRemat` | whether a node may be rematerialised instead of spilled |
| `CallingConv` | stack/frame/return-address registers, caller-save (`C`) versus callee-save (`c`), and parameter and return conventions per register class |

What stays generic, in `tb/` rather than any target directory:

1. **GCM/dataflow** — SSA scheduling that preserves memory aliasing;
2. **register allocation** — `briggs_ra.c`, `rogers_ra.c`, `ra_split.c`, solving
   the constraints the target *declared* rather than each target allocating;
3. **emission** — `node_emit` writes bytes from the flattened graph.

`abi.c` is shared. `tb.dsl` declares the IR node hierarchy once as
S-expressions with flags (`PINNED`, `CTRL`, `TERMINATOR`, `MEMORY_IN/OUT`,
`CONSTANT`) and `parent=` inheritance, with platform-specific nodes included
separately — so node kinds are data, not per-target boilerplate.

The transferable idea is one sentence: **a target describes its register file as
constraints; a shared allocator solves them.** Kofun currently has each target
allocate for itself, which is why those four functions are identical.

## Proposed contract

Mirror TB's split, sized to Kofun's actual profile rather than adopting a graph
IR:

- `TargetRegisterFile` — register classes, counts, scratch/value/call-safe roles,
  and the no-register sentinel, as **data**. The four `take_*_register`
  functions become one shared implementation reading it.
- `TargetCallingConvention` — argument and return registers, caller/callee-save
  sets, frame layout, in the spirit of TB's `CallingConv`.
- `TargetEmitter` — the only hand-written per-target code: encode a move, an
  arithmetic op, a compare, a branch, a call, prologue/epilogue. These are the
  `x64_mov_*` / `a64_*` families that legitimately differ per ISA.
- Shared lowering drives the emitter, so `function_layout`, `function_epilogue`,
  `function_guard_inverse`, `function_tail_call` and the runtime/trap stubs are
  written once.

## Relationship to DD-022 — this is the hard part

DD-022 says redundancy that a gate turns into evidence must not be shared. The
x86-64/AArch64 pair *is* differential evidence: the native gate runs the same
program on both targets and requires the same observable behaviour and the same
`R010` diagnostic bytes. So this issue must not be read as "share everything".

The separating argument is the diff column above:

- **The diff-0 pairs already provide no differential value.** Identical code
  cannot disagree, so the existing gate proves nothing about those four
  functions today. Sharing them loses no evidence and gains one tested
  implementation. Stage this first.
- **The diverged pairs do provide value.** `function_expression` is 201 versus
  190 lines with 239 differing lines — two genuinely independent lowerings that
  agree on the corpus. Sharing those *would* reduce evidence, and the gate would
  keep passing if the single shared lowering were wrong on both targets at once.

So the design is staged, and the second stage needs its own decision: either
keep two lowerings and accept the cost, or share and replace the lost
differential with something else — a lowering-level oracle, or property tests
over the emitter contract. That decision is out of scope here and should be its
own issue; this one must not silently make it.

### In scope

- The three descriptor types above, and the shared register allocator that reads
  `TargetRegisterFile`.
- Migrating the four diff-0 `take_*_register` functions to the shared
  implementation, for both targets.
- Recording the contract in `docs/NATIVE_BACKEND.md`, and recording in
  `docs/DESIGN_DECISIONS.md` that a new target implements `TargetEmitter` and
  declares its register file, and adds nothing else.

### Out of scope

- Adopting a graph or sea-of-nodes IR. TB's `Ctx` model is the reference for the
  *split*, not for its IR.
- Sharing the diverged lowering pairs. Staged deliberately, per DD-022 above.
- A third target. #278 stays deferred; this issue is what makes it cheaper.
- Any change to emitted bytes.

## Acceptance criteria

- [ ] One shared register allocator; no `x64_take_*` / `a64_take_*` pair remains.
- [ ] A target's register file, calling convention, and emitter are the only
      per-target artifacts, and `docs/NATIVE_BACKEND.md` says so.
- [ ] **Emitted bytes are unchanged for every native fixture** — this is a
      refactor, and the existing goldens are the proof.
- [ ] A mutation fixture: perturbing the shared allocator fails the native gate
      on both targets, so the shared path is genuinely exercised.
- [ ] `docs/DESIGN_DECISIONS.md` records what a new target must supply, so the
      next backend cannot quietly re-add a duplicate.
- [ ] DD-022 is cited where the staging decision is recorded, and the diverged
      pairs are explicitly listed as *not* shared by this issue.

## Validation

| --- | --- | --- |
| Byte-identical output | `make native` | every fixture's emitted bytes unchanged |
| Both targets | `sh bootstrap/native/check.sh` | x86-64 and AArch64 pass, AArch64 under `qemu-aarch64` |
| Function profile | `sh tests/conformance/functions/run.sh` | pass |
| Duplicate check | `grep -c 'x64_take_\|a64_take_' bootstrap/native/core_compiler.c` | 0 |
| Negative | perturb the shared allocator | native gate fails on both targets |
| Regression | `make verify` | pass |

## Dependencies

- Parent: #554, whose "codegen behind an interface" acceptance bullet this
  implements.
- Makes cheaper, does not block: #278, #26, #710, #158–#176.
- Constrained by: DD-022.

## Risks and open questions

- The staging boundary is the whole risk. Sharing the diverged lowering pairs
  because they look like the diff-0 ones would delete real evidence while every
  gate kept passing. The diff column is the criterion, and it should be
  re-measured rather than assumed at implementation time.
- `a64_expression` is 25 lines against `x64_expression`'s 468, handling only
  literals, add and multiply. Whether that is a narrower path with the function
  profile carried elsewhere, or genuine capability drift between targets, is
  unresolved here and worth settling before touching the lowering layer.

hjosugi:
Closes #770.

## What changed

A native target now declares only the facts its ABI decides. It brings a
`TargetRegisterFile` — its caller-saved scratch class, its call-safe class, and
the value meaning "no register" — and the allocation policy is written once
against that descriptor.

```c
typedef struct {
    const unsigned *scratch;
    size_t scratch_count;
    const unsigned *call_safe;
    size_t call_safe_count;
    unsigned no_register;
} TargetRegisterFile;
```

`x64_take_*` and `a64_take_*` are gone — `grep -c 'x64_take_\|a64_take_'` is 0.
`RegisterFile` is indexed by architectural register number and sized for the
widest file any target declares, so one type serves a 16-register and a
32-register target without either reading the other's bounds.

## Why these four and not the rest

The issue's diff column, re-measured rather than assumed. After normalising the
`X64_`/`A64_` prefixes the four `take_*_register` pairs were **identical**.
Two spellings of one algorithm cannot disagree, so the native gate proved
nothing about them: sharing loses no evidence and gains one tested
implementation.

The diverged lowering pairs are **not** shared, deliberately.
`function_expression` is 201 lines against 190 with 239 differing — two
genuinely independent lowerings the gate requires to agree on observable
behaviour and on `R010` diagnostic bytes. There the agreement *is* the
evidence, and sharing them would delete it while every gate kept passing.
DD-023 cites DD-022 for exactly this and lists the 25 pairs left alone, so the
staging boundary is recorded rather than silently crossed.

## Emitted bytes are unchanged

This is a refactor and the goldens are the proof. Every artifact the native
gate emits was hashed before and after:

```
baseline artifacts: 149
after artifacts:    149
RESULT: every emitted artifact is byte-identical
```

Both targets included. The only digests that moved are the two that pin
*sources I edited*: the provenance row for `core_compiler.c`, and the
release-evidence row for `docs/NATIVE_BACKEND.md`. The provenance `output` row
— the emitted ELF — still reads `fff4720e…`, unchanged.

## The shared path is exercised by both targets

Reversing the scratch order inside the shared allocator changes the pinned leaf
prologue on **each** target, so `function_register_allocation` fails the gate
twice rather than leaving one target silently unexercised:

| target | unmutated | mutated |
|---|---|---|
| x86-64 | `… 4c 8b d3` | `… 4c 8b db` (r10 → r11) |
| AArch64 | `… ee 03 00 aa ec 03 0e aa` | `… ed 03 00 aa ef 03 0d aa` (x14 → x13/x15) |

Both pins hold on the unmutated compiler and break under the mutation. The
gate itself was also run with the mutant in place and fails.

## Validation

| Check | Result |
| --- | --- |
| `make native` | pass |
| `make test` (incl. `tests/conformance/functions`) | pass, 12/12 |
| `make rfc-registry` | pass |
| `make release-claims` | pass |
| `grep -c 'x64_take_\|a64_take_'` | 0 |
| emitted bytes, both targets | 149/149 byte-identical |
| perturbed allocator | fails on both targets |

`make verify` fails one target, `tree-sitter`, for an environment reason
unrelated to this change: npm's `allowScripts` policy blocks `tree-sitter-cli`'s
install script, so `node_modules/tree-sitter-cli/tree-sitter` is absent and
`npm test` dies with `spawn … ENOENT`. This diff touches five files and no
grammar file among them. `repository-check`, which also fails on this machine
when stale agent worktrees are present, passes here.

## Not in scope

AArch64 execution is skipped locally — no `qemu-aarch64` on this machine — so
the AArch64 evidence above is emitted-byte evidence, which is what the
acceptance criterion asks for. CI runs the execution half.

`a64_expression` being 25 lines against `x64_expression`'s 468 is left
unresolved, as the issue asks: whether that is a narrower path or genuine
capability drift is worth settling before anyone touches the lowering layer.

---
_Generated by [Claude Code](https://claude.com/claude-code)_


hjosugi:
`origin/main` moved (#777 landed the #747 Text slice) and this PR went `CONFLICTING` on `artifacts/release-evidence/index.json` — a generated file both changes touch.

Resolved by taking `main`'s version and re-running `make release-evidence`, so every digest is recomputed from the merged tree rather than hand-merged.

Re-verified against the new `main`:

| Check | Result |
| --- | --- |
| `make native` | pass |
| `make release-claims` | pass |
| `grep -c 'x64_take_\|a64_take_'` | 0 |
| emitted bytes, both targets | **149/149 still byte-identical to the pre-change baseline** |

The byte-identity claim is unchanged: the same 149 artifacts hash the same as they did before this branch existed, now on top of the Text slice.

---
_Generated by [Claude Code](https://claude.com/claude-code)_

Issue #794, opened 2026-07-31
Triage sweep 2026-07-31: all 500+ open issues classified by how quick they are


hjosugi:
## Metadata

- State: ready
- Tracker role: curated
- Kind: maintenance
- Priority: P3
- Size: S

## Why this exists

Every open issue was read on 2026-07-31 against `18bdc57` and classified by how quickly it could be finished. The point is that **the tracker does not answer "what can I pick up right now"** — 85% of it is a generated catalogue, and several of the remaining issues carry states that stopped being true when a blocker landed. This records the answer so the next person does not re-derive it.

## The shape of the tracker

| bucket | count | actionable? |
|---|---:|---|
| `planning` (umbrella parents only) | 11 | Yes. #624, #605, #575, #543, #532, #298, #289, #32, #31, #30 and one more — real parent issues with state and sizing, not catalogue rows. |
| `curated` | 69 | Yes, but see below. |

The 435 generated catalogue placeholders — every one of them carrying no State, Priority, Size, or Kind — have since been closed as `not_planned`, leaving 80 open issues, and every open issue now carries a milestone (`M0-spec`, `M1-bootstrap`, `M2-alpha`, `M3-beta`, `M4-1.0`) where before this 509 of 514 had none.

Of the curated set, the great majority is `blocked`, `deferred`, `needs-detail`, or `needs-decision`. **Genuinely quick work is rare and mostly consists of correcting states that a landed blocker has already invalidated.**

## Quick list, in the order I would take them

State corrections already applied in this sweep are marked ✅.

| # | issue | what makes it quick | status |
|---|---|---|---|
| 1 | #609 | All four gates in its own Validation table pass on `main`; the last comment claiming unmerged local work is stale | ✅ → `verification-pending` with evidence |
| 2 | #606 | Names #609 as its **only** blocker | ✅ → `ready` |
| 3 | #312 | Names #70 as its only blocker; #70 landed as `1e29395` | ✅ → `ready` |
| 4 | #550 | Labelled `in-progress` with zero work on `main` and no open PR; `E2S19` still fires | ✅ → `ready` |
| 5 | #314 | Two of three stated blockers (#70, #120) landed; body says "all three are still absent" | ✅ commented, stays `blocked` — one real blocker remains |
| 6 | #605 | Parent tracker of #609; body names #609 as the sole current blocker | ✅ → `ready`, `blocked` dropped; State line and the #609 delivery-plan entry synced to match |
[...]

Issue #783, opened 2026-07-28
records: lower nominal record construction and field reads to the Stage 2 C11 slice (#644)


hjosugi:
## Metadata

- State: ready
- Tracker role: curated
- Kind: implementation
- Area: Native Code Generation (Stage 2 C11 slice)
- Priority: P2
- Roadmap milestone: M2-alpha
- Size: M
- Parent: #644
- Generated backlog ID: none
- Fingerprint: none

## Summary

Lower nominal record declaration, labelled construction, pass, return, and
typed field reads for `Int` and `Bool` fields through the Stage 2 C11 slice, so
a record value executes on a real backend instead of only inside the record
frontend's own evaluator. This is item 1 of the bounded producer artifact
#644 requires.

## Current behavior and evidence

Records are frontend-only, and the accepted specification says so directly.

- `spec/records-v1.md` opens with: the bounded frontend under
  `tests/conformance/records/` implements and gates the document, and **no
  production compiler path, module system, or backend lowers records yet**.
- `docs/MVP_IMPLEMENTED.md:27` records the same boundary: bounded typed
  frontend only, with no module, generic, or backend lowering.
- `tests/conformance/records/token_pipeline.run` is a `kofun-record-run/v1`
  transcript emitted by `bootstrap/stage2/record_frontend.c`. It is the record
  frontend's own evaluator output, not a program executed by a backend, so it
  cannot satisfy the "no host-language substitution" requirement in #644.
- The negative fixtures in `tests/conformance/records/` — `unknown_field`,
  `wrong_field_type`, `field_assignment`, `use_after_move`, `partial_move` —
  already fix the diagnostics this issue must leave byte-identical.

## Why now

#644 requires construction, pass/return, and typed field reads to be proved in
Kofun source through a checked source → typed HIR → executable backend path.
A frontend evaluator transcript is exactly the host-only substitute that issue
rules out. Records also remain named P1 work in `docs/ROADMAP.md`, and this is
the first slice that turns the accepted `spec/records-v1.md` design into
executable evidence.

## Goal

A `.kofun` fixture declares a record of `Int` and `Bool` fields, constructs it
with labelled arguments, passes it to a function, returns it, reads its fields,
and prints results under the Stage 2 C11 slice, gated by a `.stdout` golden.

### In scope

- `type Name = { field: Type, ... }` declaration collection in the Stage 2
  compiler path, for `Int` and `Bool` field types;
- `Name(field: value, ...)` labelled construction with free written field order;
[...]

Issue #757, opened 2026-07-27
Build entry point: replace the Makefile with a Taskfile, keeping every gate byte-identical


hjosugi:
## Metadata

- State: ready
- Tracker role: curated
- Kind: maintenance
- Area: Build system
- Priority: P2
- Milestone: M1-bootstrap
- Size: M
- Generated backlog ID: none
- Fingerprint: none

## Summary

Replace the `Makefile` with a `Taskfile.yml` driven by [go-task](https://taskfile.dev), so the repository's 60 gates keep their exact commands and observations while gaining declared dependencies, described task listing, and per-task parallelism without `.PHONY` bookkeeping. No gate's behaviour changes; only the entry point does.

## Current behavior and evidence

`make` is the single entry point. Measured on the current tree:

- the `Makefile` declares **60 targets**;
- **306 real `make <target>` invocations** appear across **43 files** outside the `Makefile` — 27 Markdown, 7 shell, 4 JSON, 3 `.mjs`, 1 YAML, 1 `.kofun`;
- `VERIFY_TARGETS` is a hand-maintained 55-name string, and `.PHONY` a parallel 56-name string. Neither is checked against the actual target set, so a target can be added to one and forgotten in the other;
- `make verify` re-invokes `$(MAKE) -j$(VERIFY_JOBS)` to get parallelism, then runs `lsp` and `roadmap` serially afterwards with a comment explaining why;
- `make help` is a hand-written 58-line `printf` block listing target descriptions, which drifts from the targets independently of both lists above.

Two machine-checked manifests now resolve these commands, and both would need to follow:

- `release/claims.json` — `tests/release/validate-claims.mjs` resolves `make <target>` against the `Makefile` for every positive gate and reproduction command;
- `rfcs/index.json` — `tests/rfc/validate-registry.mjs` does the same for every implementation gate.

Nothing in the frozen bootstrap artifacts is affected. `bootstrap/stage2/compiler.{kofun,c}`, `bootstrap/selfhost/driver/S.c` and `bootstrap/selfhost/frontend/S.hir` contain the word "make" only inside English prose comments, so no digest, byte-span pin or hand-audited transliteration pair is touched by this change.

## Why now

[...]

Issue #725, opened 2026-07-26
[Umbrella] Deliver Decimal-backed Fixed[S] on the bounded C11 Stage 2 profile


hjosugi:
## Metadata

- State: blocked
- Tracker role: planning umbrella
- Kind: planning
- Area: Decimal / type system / C11 Stage 2
- Priority: P2
- Size: L (delivered through bounded children)
- Parent: #710
- Blocked by: #1250, #1251, #1252, #1253
- Audited against: `origin/main@28e1e6213533446fc646fe6a035dd24ff3ab420b`

## Goal

Replace the truthful interim `runtime-scale/v1` boundary with one explicitly named Decimal-backed `Fixed[S]` profile on C11 Stage 2, while preserving plain Decimal canonicalization and keeping every other backend refusal honest for #710.

## Delivered foundation

| Boundary | Owner | Status |
| --- | --- | --- |
| runtime-only Decimal scale profile | Part A / PR #804 | complete |
| literal integer const-generic identity/specialization | #916 / PR #1010 | complete |
| executable identity-vs-product boundary | #1079 / PR #1085 | complete |
| truthful docs distinguishing identity from Fixed semantics | #1102 / main `96e1b01a` | complete |
| native Decimal exact/rounded runtime | #721-#724 | complete on C11 Stage 2 |

Current `task decimal` correctly says `Fixed[2]` and `Fixed[3]` are distinct type identities while `stdlib/decimal` has no Decimal-backed Fixed product.

## Corrected evidence

Two claims in the former body were stale or wrong:

1. Arithmetic result rules are not open. `docs/DECIMAL.md` already fixes same-scale exact `+/-` and `Fixed[A] * Fixed[B] -> Fixed[A+B]`, with explicit cross-scale conversion.
2. A Stage 2 lowering refusal is required to leave no C/backend artifact, but the compiler deliberately retains parsed IR/token diagnostic checkpoints for some refusals. Current E2S151 produces no C and deterministically retains a 446-byte IR plus 1392-byte token tape. Removing those globally would violate the pinned optional-coalescing contract; the children preserve and describe the distinction.

What is genuinely undecided is the executable Fixed carrier/API/ownership/failure profile, isolated in #1249.

## Delivery DAG

```text
#1249 decide Fixed[S] executable profile
   |
   v
#1250 carrier + construction + scale identity
   |\
   | +--> #1251 explicit Decimal <-> Fixed conversion
   | +--> #1252 exact Fixed arithmetic
[...]

PR #806, opened 2026-08-01
Parallel pickup from the delivery roadmap: #550, #312

hjosugi:
Picks up the `ready` issues on the delivery roadmap that no open PR and no `in-progress` label already claims, worked in parallel and collected on one branch.

## What this branch does not overlap with

| Issue | Already claimed by |
|---|---|
| #638, #640 | PR #805 (merged as `e2b5242`) |
| #757, #794 | PR #789 (merged as `bb9e950`) |
| #783 | PR #807 (merged as `b5ea9ec`) |
| #606 | PR #808 (merged as `f89629c`) — dropped from this branch after the coordination note below |
| #725 | PR #804 |
| #637 | `in-progress` |
| #723 | `ready`, but its own lowering (`E2S99`, #710 slice 4) is not there yet |

That leaves **#550** and **#312**, both of which are now implemented here.

## #550 — a final `if`/`else` is the function's result

Slice 1 (#799) made a final *simple* expression an implicit checked return. This does the same for a final `if`/`else`, routing it to the *existing* `validate_value_if` + `emit_value_into` — the same join and lowering that `return if c { a } else { b }` already used, so the final position gains no rules of its own.

It is a widening, not a change of meaning: before this, a non-`main` Int-returning body ending in an `if` **always** reached `E2S19`, so no program that compiled before compiles differently now.

The `else`-less form is refused by name: `error[E2S27]: a final `if` needs an `else`; its false path yields no Int`. `E2S27` is reused rather than minted — a final `if` *is* a value-position `if` — with a registered code and an executable fixture, so the registry gate is not satisfied vacuously.

**Lambda/named parity is partial, and the boundary was measured rather than assumed.** The gate now compares the two forms pairwise and they agree, but Stage 2 Core has only *arrow* lambdas (`fn(x) => e`) whose body already is an expression — there is no block whose last statement could differ. Two forms the issue's prose implies are not covered: a braced lambda body `fn(x) { x * 2 }` is not in this Core's grammar at all, and a value-`if` as a lambda body gives `E2S35` because `lambda_parameters_end` measures the body with `expression_end`. Both are lambda-body grammar, which #550's own triage assigns to #547.

Stage 2 only, measured against `sign`: native reports `native Core if body must be `{ return Int }`` and wasm32 supports only `let`/`print`/`return`/`if` statements. Unlike slice 1's bare expression there is no existing native shape to rewrite into, so the gate sits in `bootstrap/stage2/check.sh` as slice 1's did.

**A record-shaped hole in the same guard, closed here.** #807 added nominal records, but the final-expression guard excluded only `main` and enums, so a record-returning function took the Int path and Stage 2 emitted `int64_t kofun_result = k_b1;` — a struct assigned to an integer, which no C compiler accepts. Reproduced against a pristine `b5ea9ec` build, so it is a defect that was already on `main`; it is the same guard this slice extends, so it is closed and gated for both forms.

## #312 — Optional narrowing without leaking refinements

The issue body's "Blocking artifact" paragraph is stale, as the #794 sweep said: `TYPE_OPTIONAL`, contextual `null`, and spans are all in `bootstrap/stage2/optional_frontend.c`, and `tests/conformance/optional/run.sh` exists.

The full contract, not a slice: four recognized shapes (`x != null` / `null != x` refine the true edge, `x == null` / `null == x` the false edge) over a direct local binding declared `Optional(T)`, matched syntactically with the block's `{` required to follow so a longer expression containing one cannot be mistaken for it; early-return guards in both directions; per-branch environments merging by intersection at joins; and every invalidation rule — assignment (with the replacement still checked against declared `Optional(T)`), mutable-binding-passed-to-a-call, immutable `let` retaining, and loop backedge.

The IR carries the analysis rather than implying it: `refinement` rows print declared beside refined and name the edge, `narrowed-use` marks each use typed as `T`, and `refinement-discarded` records every discard with `reason=assignment|call|loop-backedge`.

Two new registered codes: **E2S142** (property/index paths, unrecognized `null` comparisons) and **E2S143** (assignment to an immutable binding — without it the retain-across-calls rule is unsound).

Deliberately out: narrowing from a `while` header. The loop backedge is deliberately coarse — it discards any *mutable* binding the body mentions, read or written; sound and imprecise was chosen over precise and fragile. Ownership modes are refused by this frontend, so "edit/own/unknown-effect" collapses to "every call has unknown effects", stated in the docs rather than papered over. Compound conditions, aliases, captured variables, interprocedural summaries, `match`, safe navigation, truthiness, user-defined equality, union narrowing, and runtime representation are all refused or non-narrowing, never optimistically assumed.

This lane lives entirely in `optional_frontend.c`, so the `compiler.c`/`compiler.kofun` pair, `SHA256SUMS`, and `function_text_provenance.txt` are untouched by it.

**Gate strength was measured, not assumed.** Seven mutations of the analysis — inverted edge, dropped call/assignment/backedge invalidation, shared then/else environments, union-instead-of-intersection joins, universal block termination — each fail the gates.

## Combining the two lanes

One conflict: the hand-maintained `repository_error_cases` pin in `tests/typed-sidecar/stage2-events.sh`. Base 215, #550 +1, #312 +13, resolved to **229**; `diagnostic_cases` stays at 47 since #312 adds no Stage 2 diagnostic fixture. `artifacts/release-evidence/index.json` pins SHA-256 of `Taskfile.yml`, `docs/TYPE_SYSTEM.md`, and `tests/diagnostics/registry.tsv`, all edited here — `task release-evidence` regenerates it with no diff, so it is consistent after combining.

Two unrelated issues share one branch because this session is pinned to `claude/issues-project-check-n6v0ms`.

## Verification

Two local caveats worth recording so nobody re-derives them:

1. Two container dependencies were missing and had to be installed before any local gate result meant anything: `xxd` (without it `task diagnostics` fails with `FAIL: xxd is required` on an unmodified tree) and `qemu-aarch64` (CI installs `qemu-user`). Local results taken before that are not evidence.
2. `task repository-check` fails locally whenever an agent worktree is live under the checkout: its `find` prunes `./build` but not `./*/build`, so it picks up `build/cli-test/legacy.kf`, a file the test suite itself generates. Nothing to do with tracked content, and CI has no nested worktrees.

Results:

- CI green at `b7cd4b3` (`main` + one `.gitignore` line) — the baseline.
- CI green at `d30ffef` (#550 landed).
- Combined branch, local: `stage2-events` passes at the resolved pin of 229; `tests/conformance/optional-narrowing/run.sh` green; `sh bootstrap/stage2/check.sh` and `sh tests/diagnostics/stage2/run.sh` exit 0 (47/47 Stage 2 codes). Full `task verify` was still running when this was pushed — CI on `758c6f5` is the authoritative result and is reported below.


hjosugi:
Coordination note: #606 was claimed in the issue at 2026-08-01 02:17 UTC and is now implemented in focused PR #808 (`d5c7eedf`), with local full/focused gates passed and CI running. Please drop #606 from this combined branch to avoid duplicate LSP work. I am not touching this PR's #550/#312 lanes.

hjosugi:
Dropped. #606 is out of this branch — the agent working it has been stopped and its branch and worktree deleted, so nothing from that lane will reach this PR. #808 owns it.

This branch is now #550 and #312 only, both still in the Stage 2 compiler pair and both rebased onto `b5ea9ec` so they sit on top of #807's record lowering.

---

Issue #864, opened 2026-08-01
law evidence: optional-bool-monad.evidence.json records a source hash that no longer matches, and has no producer


hjosugi:
## Metadata

- State: needs-decision
- Tracker role: curated
- Kind: implementation
- Area: Quality and Evidence
- Priority: P2
- Roadmap milestone: M2-alpha

## Summary

`artifacts/optional-bool-monad.evidence.json` asserts a `proven-finite` law
result over 264 cases for `examples/proven_optional_bool_monad.kofun`, binds
itself to that file by `source.sha256`, and the hash has not matched since
2026-07-30. Nothing compares the two, and nothing can regenerate the artifact.

Measured on `9923169`.

## Evidence

~~~
$ node -e "console.log(require('./artifacts/optional-bool-monad.evidence.json').source)"
{ path: 'examples/proven_optional_bool_monad.kofun', sha256: '0adb5fea24c34657e63598c4ad01cbf7a9d38206e16eeeada839278a3009ece6' }
$ sha256sum examples/proven_optional_bool_monad.kofun
2e9da000501d9b99fbbfd4cfea3aa483f1d7db68e14a6be8f7c012f87194739a  examples/proven_optional_bool_monad.kofun
~~~

The artifact still reports:

| Field | Value |
|---|---|
| `status` | `passed` |
| `required_assurance` | `proven-finite` |
| `reports[0].cases_checked` | 264 of 264 |
| `compiler.version` | `0.2.0-bootstrap` |

The repository is on `0.3.32-seed`. Both files last changed in `cd1d19b`, a
snapshot merge, which is where the two fell out of step.

## Why nothing caught it

It is the only law-evidence artifact in the tree that records a
`source.sha256` at all:

| Artifact | Binds to a source by hash |
|---|---|
| `artifacts/optional-bool-monad.evidence.json` | yes — and it mismatches |
| `stdlib/decimal/tests/law-evidence.json` | no |

`stdlib/decimal/tests/validate_law_evidence.mjs` is the only law-evidence
checker, and it validates the JSON against a schema only — it never opens the
source. So the one artifact that made a falsifiable claim about a file was the
one nothing falsified.

## Why it cannot simply be regenerated

There is no producer on `main`. `bin/kofun` has no `law` or `prove` command:

~~~
$ ./bin/kofun law
kofun: unknown command: law
~~~

DD-035 states that laws are library declarations checked by a finite-model
engine; the engine is not reachable from the CLI. So the recorded result
[...]

PR #872, opened 2026-08-01
examples: bind every example to the check that owns it (#864)

hjosugi:
## The question this started from

"Is `examples/` using old syntax?"

**No.** Every construct in all twenty files traces to a live entry in
`docs/DESIGN_DECISIONS.md` — `fn` (DD-001), `null` and `T?` (DD-002), `else if`
(DD-003), `read`/`take`/`own` modes (DD-005, DD-006), `|>` (DD-011), `type`
records (DD-021), `law` (DD-035) — and `repr(C) struct` is its own grammar
production for the C ABI profile, not a retired record spelling. `task
tree-sitter` parses all twenty with no `ERROR` or `MISSING` node.

What *had* rotted was everything the parser does not look at.

## What was actually stale

`docs/REPOSITORY_GUIDE.md` says an example is owned by "the example plus its
nearest check". **Eleven of the twenty had no check at all.** The Tree-sitter
corpus gate proves a file parses; nothing proved anything else. An example
could stop compiling, stop running, or cite a diagnostic that no longer exists,
and no gate would say so.

Two things had already happened in that gap.

**A retired diagnostic code.** `ownership.kofun` told the reader that
uncommenting a line "produces E330: value was already taken". No `E3xx` code has
ever been in `tests/diagnostics/registry.tsv` — the whole numbering scheme is
gone. The comment now names the condition, says the general ownership pass is
open, and points at `E2S123` in
`tests/conformance/records/use_after_move.kofun`, where the rejection is
actually executable.

**A law-evidence artifact that had drifted from its subject.** It recorded
`source.sha256 = 0adb5fea…` for `proven_optional_bool_monad.kofun`, whose bytes
hashed to `2e9da000…`, and still reported `status: passed`, `proven-finite`,
264 of 264 cases. It was the only law-evidence artifact binding itself to a
source by hash, and the only law-evidence checker validated the JSON against a
schema without ever opening the source — so the one artifact making a
falsifiable claim was the one nothing falsified. That became #864, and **#875
resolved it** by reducing the artifact to what is still true.

## What this adds

`examples/README.md` is the missing binding; `examples/check.sh` (`task
examples`) enforces it. Every `.kofun` appears exactly once and every row names
a file that exists, so an example cannot be added without declaring how it is
checked.

| Status | Count | What the gate does |
|---|---:|---|
| `owned` | 9 | asserts the named check still names the example |
| `runs` | 3 | builds it, runs it, compares stdout to a committed `.expected` |
| `illustrative` | 8 | asserts `bin/kofun check` still refuses it with the stated code |

`illustrative` is the load-bearing one: it makes "not implemented yet" a
falsifiable claim. When the slice grows the gate **fails** and asks for the
example to be promoted, so the table cannot quietly describe a past version of
the compiler.

The three `runs` files — `hello`, `native_answer`, `project/src/bench` — all
worked and had simply never been executed by anything.

Plus a registry sweep over every diagnostic code cited by an example, so the
`E330` class cannot return, and a hash check over every example-backed evidence
artifact — which also fails when it finds *no* such artifact, because a hash
check over an empty set passes without checking anything.

## The gate is not decorative

Each of its own claims, broken:

```
$ # remove a row from README.md
FAIL: examples/README.md and examples/ disagree
  present but unlisted:
    hello.kofun
$ # change an expected output
FAIL: native_answer.kofun: output differs from native_answer.expected
$ # restore the retired code
FAIL: an example cites unregistered diagnostic code `E330`; see tests/diagnostics/registry.tsv
```

## Reconciled with #875

#875 landed at the same path from the other side of #864 while this was open.
Both checks belong and they check different things, so each got a path rather
than one winning:

| Path | Owns | Task |
|---|---|---|
| `examples/check.sh` | the manifest over all twenty examples | `examples` |
| `examples/check-law-evidence.sh` | #875's checker, moved unchanged | `example-law-evidence` |

An earlier revision of this branch carried the drifted artifact as a named debt,
because no producer existed to reproduce its result and rewriting the hash would
have asserted a proof that was never re-run. #875 made that unnecessary: the
artifact now records the current hash and states `unverified`, so the general
net here passes on it honestly instead of printing a NOTE.

`artifacts/release-evidence/index.json` follows because it digests
`Taskfile.yml`, which gained the `examples` task.

## Validation

| Check | Result |
|---|---|
| `task examples` | green; falsified three ways above |
| `task example-law-evidence` | green |
| `task tree-sitter` | green — all twenty parse |
| `task release-claims` | green |
| `task verify` | EXIT=0, 956 PASS, 0 FAIL (both gates run inside it) |

https://claude.ai/code/session_01M6ybgRb3RtipMmBd9Ehvu2


hjosugi:
Reconciled with #875, which landed at the same path from the other side of #864 while this was open.

Both checks belong and they check different things, so each got a path rather than one winning:

| Path | Owns | Task |
|---|---|---|
| `examples/check.sh` | the manifest over all twenty examples | `examples` |
| `examples/check-law-evidence.sh` | #875's checker, moved unchanged | `example-law-evidence` |

The debt carve-out is gone. #875 repaired the binding it existed for — the artifact now records the current hash and states `unverified` instead of a `proven-finite` result nobody could reproduce — so the general net here passes on it honestly instead of printing a NOTE:

```
PASS: every example-backed evidence artifact matches its source (1)
```

It also now fails when it finds **no** example-backed artifact at all. A hash check over an empty set passes without checking anything, which is the same shape as the gap that let this drift for two days.

`task verify`: EXIT=0, 956 PASS, 0 FAIL on the merged tree.

---

Issue #648, opened 2026-07-23
Time-zone data v1: versioned tiny tzdb reader with gap and fold resolution


hjosugi:
## Metadata

- State: blocked
- Tracker role: curated parent
- Kind: implementation
- Area: Core Standard Library
- Priority: P2
- Size: M
- Contract: #639 (completed), [`docs/stdlib/date-time.md`](https://github.com/hjosugi/kofun/blob/main/docs/stdlib/date-time.md)
- Executable producer child: #878
- Evidence audit: `origin/main@59a75a60fd6bbe407f29929555d907abf06cfa0a`

## Goal

Implement the smallest versioned time-zone database path that proves deterministic named-zone lookup, transition handling, and explicit local-time ambiguity resolution before packaging the full IANA database.

## Current evidence and blocker

The normative date/time contract exists and #639 is completed. Current `main` also has the bounded date/time producer documentation, but it has no `tests/stdlib/tzdb/check.sh` and no executable tzdb corpus.

The sole named hard blocker is now #878, which owns the bounded Stage 2 producer artifact: minimum value shapes, a closed resolution sum and typed error result, an injected-`Bytes` reader, deterministic transition lookup, provenance, malformed/limit fixtures, and one executable backend observation.

Historical broad dependencies are not blockers:

- #645 remains open and blocked, but the tzdb child needs only a bounded shared Instant/LocalDateTime/UtcOffset subset, not the full calendar/RFC 3339 implementation.
- #492 is closed as not planned. Ambient file access is not part of v1; any later file adapter requires a separate explicit child.
- #408 and #421 were closed as planning items and do not own this artifact.
- #546 and #639 are completed design inputs, not implementation blockers.

## Scope

- define the bounded `ZoneId`, `Tzdb`, zoned value, transition, and shared date/time shapes needed by #878;
- read one committed, versioned, bounded fixture format from injected `Bytes`;
- record and expose database version and content digest;
- resolve local date/time as `Unique`, `Ambiguous`, or `Nonexistent` without silently choosing an offset;
- require explicit earlier/later/reject or shift policy when a caller resolves a fold or gap;
- serialize enough zone/version/digest provenance to detect rule drift;
- keep an ambient `File` adapter separate from the pure injected-Bytes core.

## Out of scope

- downloading or updating tzdb at runtime;
- ambient host zoneinfo discovery;
- locale formatting, CLDR, leap-second tables, or historical-calendar variants;
- shipping the complete IANA database in the first slice;
[...]

PR #889, opened 2026-08-01
stdlib: a bounded tzdb producer that reads injected bytes and never guesses (#878)

### LINKED ISSUE #878: Stage 2 tzdb producer: bounded Bytes reader and explicit gap/fold resolution
## Metadata

- State: ready
- Tracker role: curated child
- Kind: implementation
- Area: C11 backend / Core Standard Library
- Priority: P2
- Size: M
- Parent: #648
- Normative input: #639 (completed)
- Audited against: `origin/main@59a75a60fd6bbe407f29929555d907abf06cfa0a`

## Goal

Land one bounded Stage 2 producer artifact, written in Kofun source and executed through a supported backend, that reads injected versioned time-zone bytes, performs deterministic transition lookup, and returns explicit normal/gap/fold results with provenance.

This child owns the missing executable artifact identified by #648. It does not require the whole date/time core, ambient file I/O, or a packaged IANA database.

## Scope

The bounded artifact must provide:

1. Minimum value shapes for `ZoneId`, `Transition`, `Tzdb` metadata, `Instant`, `LocalDateTime`, and `UtcOffset`, including construction, pass/return, and typed reads.
2. A closed resolution sum with payloads for `Unique`, `Ambiguous`, and `Nonexistent`.
3. A closed typed error result for malformed or unsupported versions, digest mismatch, invalid zones, truncated/trailing bytes, arithmetic overflow, oversized input, and deterministic limit exhaustion.
4. A deterministic bounded reader for one committed fixture format whose core input is injected `Bytes`.
5. Binary search, or another explicitly bounded deterministic transition lookup, exercised for one normal instant, one gap, and one fold. No path may silently choose an offset.
6. Canonical output carrying the tzdb version and content digest so rule-version drift is observable.
7. One checked source -> typed HIR -> executable backend path; a host-language-only oracle is insufficient.

## Out of scope

- Full completion of #645. Only the minimum shared Instant/LocalDateTime/UtcOffset subset is needed.
- Ambient `File` access. #492 is closed as not planned; any later file adapter needs its own explicit child and must preserve the injected-Bytes core.
- Runtime download/update, host zoneinfo discovery, locale/CLDR, leap-second tables, historical calendars, or the full IANA database.
- Silent earlier/later selection for ambiguous or nonexistent local time.
- Changes to the active #647 clock-adapter lane. Before implementation, re-audit open PRs/worktrees and claim issue-specific paths.

## Acceptance criteria

- [ ] The bounded Kofun value/sum/error/Bytes producer executes through Stage 2 and one supported backend.
- [ ] Normal, gap, and fold fixtures return the exact intended `Unique`, `Nonexistent`, and `Ambiguous` payloads.
- [ ] Malformed, unsupported-version, digest-mismatch, invalid-zone, truncated, trailing, overflow, oversized, and limit-exhaustion inputs fail with distinct typed results.
- [ ] Repeated lookup and serialization over the same fixture bytes are byte-identical.
- [ ] Serialized results bind zone ID, instant/local value as applicable, selected offset(s), tzdb version, and content digest.
- [ ] The gate runs without network access, host time-zone files, locale state, ambient clock state, or ambient file access.
- [ ] Reference/oracle and executable-backend observations agree, and the oracle is not substituted for the Kofun producer.
- [ ] `task verify` passes after the focused gate is integrated.

## Validation

| --- | --- | --- |
| Focused corpus | `sh tests/stdlib/tzdb/check.sh` | normal/gap/fold and every malformed/limit/provenance vector pass |
| Determinism | focused gate repeated from a clean temporary directory | byte-identical canonical output |
| Stage 2 producer | focused gate's source/HIR/backend audit | Kofun source is the producer; no host substitution |

## Dependency and close rules

- #639 is a completed normative input, not a blocker.
- #645 and #492 are not hard blockers for this injected-Bytes bounded artifact.
- #648 stays open and `blocked` on this child until this executable gate lands and the parent's remaining acceptance is re-audited.
- Landing a schema, host oracle, frontend transcript, or documentation alone does not complete this child.

hjosugi:
Closes #878.

The executable artifact #648 was missing. `task tzdb` now runs one: time-zone
rules read from bytes a caller hands over, a bounded transition lookup, and gap
and fold results that say so rather than picking an offset.

## Resolution is a closed sum

```
local 12000 (gap)    kind=Nonexistent  primary=0     secondary=3600
local 21000 (fold)   kind=Ambiguous    primary=3600  secondary=0
local 16000          kind=Unique       primary=3600
instant 20000        kind=Unique       primary=0
```

`Ambiguous` carries both offsets, earlier first; `Nonexistent` carries the ones
either side of the gap. Neither hands back one as if it were the answer —
"pick the earlier one" is a policy a caller may hold and a library may not. An
instant is always `Unique`, including at the instant a transition takes effect,
where the new offset is the one in force. That difference is printed rather
than argued for in a comment.

Both intervals are half-open at the top and the gate reads **both ends of
both**. An off-by-one there is the classic way an hour goes missing.

## The lookup

A sum of four comparisons against the four local-time boundaries the two
transitions create. The comparison count and the control flow are the same for
every input, so a replay cannot take a different path through the table —
there is only one path. The issue permits a binary search over those
boundaries; this is strictly more bounded than one.

No branch in the rank selects an offset, and the direction of a jump is read
from the transition, so a gap and a fold cannot be swapped by editing the
boundary order.

## Nine ways the bytes can be wrong, nine constructors

Bad magic, non-zero reserved byte, unsupported version, unknown zone, count
past the bound, digest mismatch, truncated, trailing, oversized, and an
addition that would leave the `Int` range. Every vector is the good fixture
with exactly one byte or one length changed, so the error is attributable to
that change. The gate counts the distinct codes rather than eyeballing them, so
a collapsed pair cannot pass by looking plausible. No path returns UTC because
it could not decide.

Validation order — size, shape, version, zone, bounds, digest — is the
contract, and the gate reads it: a digest computed over bytes whose length was
never checked would be a digest over whatever the buffer held.

## Ambient independence is executed, not grepped for

```
identical under TZ=Pacific/Kiritimati, TZ=Sao_Paulo+tr_TR locale,
and an empty environment
```

The empty environment is there because a program that needed `TZ` would more
likely fall back than fail, and a fallback would not show up as a difference
between two hostile settings.

The emitted C is still searched for the symbols that would reach host time or
time-zone state. The **source** searches now read it with comment lines
stripped: both files explain at length what they do not reach for, and a grep
over the whole text cannot tell the comment saying "does not consult TZ" from a
line that consults it. Not hypothetical — it failed on the first run, on my own
comment.

## Verified in both directions

| break | what the gate said |
|---|---|
| swap Nonexistent/Ambiguous, regenerate the golden | `a local reading inside the gap names no instant: expected '7 12000 3 0 3600 1 101 ', got '7 12000 2 …'` |
| canonical `Ambiguous` drops its second offset | `canonical Ambiguous no longer carries both offsets` |
| two error constructors share a code | `an input that runs long is trailing bytes, not ignored: expected '-6 18 ', got '-5 18 '` |
| append a byte to the empty-environment run | `output changed with an empty environment` |

Each names the rule that changed rather than reporting that a golden moved.

## Scope

Two spellings differ from the canonical surface, and the producer names both:
`Bytes` is projected as a fixed-capacity record of byte slots plus a length
(the reader still refuses a length it did not expect), and a two-offset payload
is a record beside a one-`Int` constructor, because the Core's ADT carries a
single `Int`. The sum stays closed and is still reached by `match`.
`stdlib/tzdb/tzdb.kofun` holds the canonical `Bytes`, `List`, multi-payload and
`Result` surface, and the gate pins that it still stops at the documented
compiler boundary.

`typed_hir.kofun` carries the same six value shapes and both closed sums through
the semantic sidecar, which the producer is too large to fit.

The capability row stays `specified`. This is the bounded artifact, not the
capability: the IANA database, a file adapter, and the full date/time core
remain #648, which stays open and blocked on the parent re-audit.


hjosugi:
Superseded by #888, which landed the same issue while this was in flight. Not
reopening as a competing implementation: I checked #888's resolver against mine
and it is correct where it matters, including the half-open boundary rule —
`local >= gap_start && local < gap_end`, and the same shape for the fold.

I should have re-audited for a concurrent claim on #878 immediately before
pushing rather than only at the start; the issue text asks for exactly that and
I did it once, too early.

Four things in this branch are not in #888, and I will bring them as a focused
follow-up on top of it rather than as a rewrite:

- **Boundary coverage.** #888's golden never reads an edge. Its resolver gets
  them right, so this is a regression guard rather than a fix, but an
  off-by-one at a gap or fold edge is the classic way an hour goes missing and
  nothing currently pins it.
- **Type-separation fixtures.** Nothing yet proves the toolchain refuses an
  `Instant` where a `LocalDateTime` belongs, or an epoch read off a local
  reading. A local reading and an instant are both one number, and keeping them
  apart is the whole reason the resolution sum has anywhere to live.
- **Ambient independence, executed rather than grepped.** The producer runs
  identically under `TZ=Pacific/Kiritimati`, under a `Sao_Paulo` zone with a
  `tr_TR` locale, and under `env -i`. The empty environment matters because a
  program that needed `TZ` would more likely fall back than fail, and a
  fallback does not show up as a difference between two hostile settings.
- **The canonical surface.** `stdlib/tzdb/tzdb.kofun`, holding the `Bytes`,
  `List`, multi-payload and `Result` shapes the Core cannot lower yet, pinned
  the way `stdlib/clock/adapters.kofun` is — including that it must still stop
  at the documented compiler boundary. The `time-zone-data` row in
  `stdlib/capabilities.tsv` also still points only at #648.

One observation for its own issue rather than a silent change here:
`resolve_fixed_offset` returns `ResolutionFailed(code_arithmetic_overflow())`,
so the payload is a copy of the discriminant and carries nothing about which
operand overflowed. The arithmetic guard itself is right — I checked both the
positive and negative offset branches.

PR #892, opened 2026-08-01
tzdb: pin the interval edges, the type separation, and the canonical surface (#878)

hjosugi:
Follow-up to #888. Four things its gate does not yet hold, added on top of it
rather than as a competing implementation — I had a parallel implementation in
#889 and closed it; this brings only the parts #888 does not have.

**The producer's existing 35 golden lines are unchanged.** The new observations
are appended, so every assertion #888 wrote still reads the line it was written
for.

## Interval edges

#888's golden never reads an edge. Its resolver gets them right —
`local >= start && local < end` for both the gap and the fold — so these are
regression guards, not fixes. Worth having anyway: an off-by-one at an interval
edge is the classic way an hour goes missing.

Verified by making the gap's high edge inclusive and regenerating the golden:

```
FAIL: the gap excludes its high edge: expected '1', got '3'
```

## Local readings and instants cannot be confused

Both are one number, and keeping them apart is the whole reason the resolution
sum has anywhere to live. Nothing proved the toolchain enforces it.

- `mixed_local_instant.kofun` hands an `Instant` to a resolver taking a
  `LocalDateTime`
- `local_wall_seconds_on_instant.kofun` reads an instant field off a local
  reading

Both must be refused by `check` **and** by `build`, and the gate asserts the
reason as well as the refusal. Verified by making the first one legal:

```
FAIL: mixed_local_instant passed `kofun check`; the separate tzdb types did not stop it
```

## Ambient independence, executed rather than grepped

The greps prove the symbols are not named; these runs prove the behaviour:
identical bytes under `TZ=Pacific/Kiritimati`, under a `Sao_Paulo` zone with a
`tr_TR` locale, and under `env -i`.

The empty environment matters because a program that needed `TZ` would more
likely fall back than fail, and a fallback would not show up as a difference
between two hostile settings. Verified by appending a byte to that run:

```
FAIL: output changed with an empty environment
```

## Canonical surface

`stdlib/tzdb/tzdb.kofun` holds the `Bytes`, `List`, multi-payload and `Result`
shapes the Core cannot lower yet, pinned the way `stdlib/clock/adapters.kofun`
is: its declarations must survive, its `Ambiguous` and `Nonexistent` must keep
carrying both offsets, and it must still stop at
`error[E2S02]: expected top-level fn or type`. Its field names follow the shapes
#888 landed, so the two files describe one design rather than two.

The `time-zone-data` row in `stdlib/capabilities.tsv` pointed only at #648 and
now names the gate that exists. The state stays `specified` — this is the
bounded artifact, not the capability.

## Not changed here

`resolve_fixed_offset` returns `ResolutionFailed(code_arithmetic_overflow())`,
so that payload is a copy of the discriminant and says nothing about which
operand overflowed, and `check.sh` pins it as `-7`. The arithmetic guard itself
is correct — I checked both the positive and negative offset branches. Changing
the payload changes #888's error design and its golden, so it is filed
separately rather than folded in here.


hjosugi:
## The gate was never running

Pushed a second commit that matters more than the rest of this PR.

#888 defined a `tzdb` task and did not add it to the list `verify` invokes, so
`task verify` has never run `tests/stdlib/tzdb/check.sh`. I found it by
grepping this branch's own CI log for the gate's output:

```
gh run view 30722475413 --log | grep -ci tzdb
0
```

Sixteen minutes of green `Kofun verification`, and the tzdb gate is not in it.

#878's acceptance says "`task verify` passes after the focused gate is
integrated". It passed because the gate was not integrated.

It is also worse than an unrun test. `release/claims.json` carries a claim whose
positive gate is `sh tests/stdlib/tzdb/check.sh`, and
`artifacts/release-evidence/index.json` records it.
`tests/release/check-claims.sh` validates that a claim joins its evidence and
that its digests are current — it does not execute the command. So a published
claim was backed by a gate nothing ran.

The fix is one line, `tzdb \`, beside `benchmark-summary` in the `task -C` list.

I missed this in my own parallel branch for the same reason a reviewer would:
I had added the enrolment myself, so every run I looked at did execute the gate.
Checking the log for the gate's own output, rather than checking that CI went
green, is what surfaced it — and it is the same distinction the rest of this PR
is about.

Issue #895, opened 2026-08-01
tzdb: an explicit earlier/later/reject/shift policy for folds and gaps


hjosugi:
## Metadata

- State: ready
- Tracker role: curated child
- Kind: implementation
- Area: Core Standard Library
- Priority: P2
- Size: M
- Parent: #648
- Normative input: #639, #878 (both landed)

## Goal

Give a caller holding an `Ambiguous` or `Nonexistent` a supported way to state
what they want, and record that choice.

## Current evidence

#648's Scope asks to "require explicit earlier/later/reject or shift policy
when a caller resolves a fold or gap". Nothing implements it. Over
`tests/stdlib/tzdb/tzdb.kofun` and `stdlib/tzdb/tzdb.kofun`:

```
grep -nE "fn .*(earlier|later|reject|policy|prefer|shift)"
```

matches only `pack_instants`, which packs both instants into one integer so
they can be printed.

Returning the closed sum is the right foundation and #888 did it. The missing
half is the call site. A caller who has an `Ambiguous` today must invent the
policy themselves, which is exactly the situation the sum was introduced to
prevent: the library stopped guessing, and the guess moved one level up where
nothing records it.

## Scope

- A closed policy value: `Earlier`, `Later`, `Reject` for a fold; `ShiftForward`
  and `Reject` for a gap.
- Applying a policy to a resolution yields either the instant it selected or a
  typed refusal, never a silent choice.
- No policy is the default. A caller who supplies none does not get one.
- The applied policy is recorded in the serialized result, so a value carries
  not only which instant was chosen but on what grounds.
- Gap policy is stated separately from fold policy: shifting forward over a gap
  and taking the earlier of a fold are different decisions and a caller may
  want one without the other.

## Out of scope

- Locale or CLDR-driven defaults.
- Any policy that reads ambient state to decide.
- Changing the resolution sum itself.

## Acceptance criteria

- [ ] A fold resolved under `Earlier` and under `Later` yields the two distinct
      instants, and the gate reads both.
- [ ] `Reject` yields a typed refusal for both a fold and a gap, distinct from
      every existing reader error.
- [ ] A gap resolved under `ShiftForward` yields the first instant after the
      gap, and the gate reads it.
- [ ] Applying a policy to a `Unique` result is a no-op that still records the
      policy, rather than a special case.
- [ ] The serialized result names the policy that was applied.
- [ ] The gate proves that no policy is the default: a call site without one
[...]

PR #897, opened 2026-08-01
tzdb: bind zone, version, and digest into each serialized resolution (#894)

### LINKED ISSUE #894: tzdb: bind zone, version, and digest into each serialized resolution
## Metadata

- State: ready
- Tracker role: curated child
- Kind: implementation
- Area: Core Standard Library
- Priority: P2
- Size: S
- Parent: #648
- Normative input: #639, #878 (both landed)

## Goal

Make a single serialized resolution attributable to the rules that produced it.

## Current evidence

`tests/stdlib/tzdb/tzdb.kofun` prints the version and the content digest once,
as fields 2 and 3, from `tzdb_metadata(db)`. Every resolution printed after
that carries neither:

```kofun
print(resolution_kind(fold))
print(resolution_payload(fold))
```

Two numbers, with nothing binding them to `metadata.content_digest`. The run as
a whole is attributable; an individual value is not. #648's criterion is
"Serialized zoned values carry enough provenance to detect rule-version drift",
and that is the half that is missing.

This matters beyond tidiness. Rule drift is the failure this design set out to
make observable: an offset computed under 2024b and one computed under 2025a
are different answers to the same question, and a caller holding one value has
no way to tell which it has.

## Scope

- A serialized resolution binds the zone, the format version, and the content
  digest alongside the kind and its payloads.
- The provenance comes from the `Tzdb` the resolution was computed against, not
  from a parameter a caller can pass independently — a result cannot claim
  provenance it does not have.
- `tests/stdlib/tzdb/check.sh` reads the bound fields per result rather than
  only at the top of the run.

## Out of scope

- Changing the resolution sum or its payloads.
- Any new input format or fixture.

## Acceptance criteria

- [ ] Every serialized resolution carries zone, version, and digest.
- [ ] The gate asserts them on a normal, a gap, and a fold result, not only on
      the header.
- [ ] Changing one fixture byte changes the digest carried by every result, and
      the gate observes it.
- [ ] `task verify` passes.

## Validation

| --- | --- | --- |
| Focused corpus | `sh tests/stdlib/tzdb/check.sh` | per-result provenance read on normal, gap, and fold |
| Drift | edit one fixture byte | every result's digest moves, and the gate says so |

hjosugi:
Closes #894. Stacked on #892 — review that first; this branch contains its
commits.

## The gap

Version and digest were printed once, as fields 2 and 3. Every resolution after
that carried neither, so the *run* was attributable and an individual *answer*
was not. #648 asks that serialized zoned values carry enough provenance to
detect rule-version drift.

## Provenance cannot be claimed

`resolve_local_provenanced` reads the provenance from the database it resolved
against, inside the same call. There is no second call to hand a different
database to, so a result cannot carry provenance it does not have.

## Drift, demonstrated

`fixture` is generalised to carry the first transition's instant — existing call
sites delegate with the value they had, and the output is **byte-identical
across that refactor**. So a second rule set can exist, and the same local
reading is asked of both:

| rules | digest | local 1050 |
|---|---|---|
| shipped, transition at 1000 | `475286615` | `Nonexistent`, payload 1000 |
| moved, transition at 1200 | `473962913` | `Unique`, payload 1050 |

Both answers are correct for their own rules. Without the digest travelling
with them they are two numbers that cannot be told apart.

## Verified in both directions

| break | what the gate said |
|---|---|
| provenance taken from a different database than the one resolved against | `moving a transition changes the digest every result carries: expected anything but '475286615', got '475286615'` |
| a result stops carrying the digest | `a normal resolution carries zone, version, and digest: expected '7 1 475286615', got '7 1 0'` |


hjosugi:
Rebased onto main now that #892 is merged, so this is a single commit again — the stacked history is gone and the diff is only the provenance change.

hjosugi:
Superseded by #913, which was rebased on current main after the #892 squash merge, preserves the per-result provenance and drift checks for #894, and also lands the dependent explicit fold/gap policy contract for #895. #897 now re-presents the squashed #892 ancestry and conflicts with current main, so it is being closed without merge.

PR #958, opened 2026-08-02
feat(testing): add kotest unit-test framework and executable stdlib samples

hjosugi:
## Why

A verification pass over the standard library found two things:

1. **28 of 30 canonical stdlib sources do not compile.** Only `date_time.kofun` and `decimal.kofun` pass `bin/kofun check`; the rest stop at `E2S02`/`E2S21`/`E2S32`/`E2S35`, many before the type checker.
2. **15 of 17 module gates assert that failure as expected behaviour**, then execute a separately hand-written Int-Core projection. No test file imports a canonical module, so the green lines came from re-implementations rather than the modules. The projections can drift: `json.kofun` sets `JSON_MAX_NESTING = 64`, its projection tests depth `4`, and the gate prints `bounded nesting contract: PASS`.

The READMEs describe this honestly, but there was no way to write a unit test against real code — `kofun test` existed as usage text with no implementation.

This PR adds the missing layer: a unit-testing framework, a runner, and executable samples that the framework actually tests. It does not change the canonical sources or their existing gates.

## kotest framework — `stdlib/testing/kotest.kofun`

Pure Kofun that compiles and runs through the Stage 2 C11 path today.

- vitest-order `expect_*` assertions for Int and Text, JUnit-order `assert_equals_*` aliases, `fail_now`, and a predicate assertion.
- `starts_with`/`contains` are host builtins outside this lowering slice, so the prefix/fragment predicates are implemented on `len` and `text_slice`.
- `kotest_selfcheck` runs inside every generated test binary, so a framework regression surfaces as a failure in the run itself.

**Mocking** follows Mockito's constructor-injection discipline made explicit, because the executable slice has no mutable capture:

- **stubs** — collaborators arrive as callable parameters and tests script them with lambdas;
- **`stub_unreachable`** — a forbidden collaborator whose call prints a diagnostic *and* returns a sentinel that poisons the result, so the failure is visible in the arithmetic and not only in the log;
- **spies** — `spy_*` packs call count and last argument into one Int, threaded explicitly; out-of-range recordings fail loudly instead of corrupting the count.

## Runner — `tooling/kotest/run.sh`, wired as `kofun unittest`

Collects `fn test_*` and generates a harness `main` the way the Rust test harness collects `#[test]`, then builds one translation unit through `kofun emit-c` and C11.

- **Go-style companions**: `X_test.kofun` compiles `X.kofun` alongside it with the demo `main` removed, so suites test the real functions without duplication.
- vitest-shaped output: `❯` per suite, `✓`/`✗` per test, diagnostics indented under the failing test, `Tests  N passed (N total, M suites)`. `NO_COLOR`- and TTY-aware, plus `--no-color`.
- `--filter`, `--list`, `--keep-going`, and `--watch` for hot reload.
- Exit code is 0 only when every collected test passed.

```
❯ list_sample_test.test_push_appends_in_order
  ✓ list_sample_test.test_push_appends_in_order

Tests  133 passed (133 total, 18 suites)
```

## Samples — `examples/stdlib/`, 17 modules

One runnable sample per stdlib module with a committed golden, plus its kotest suite: **133 tests over real code**. These carry the documented contracts in the executable slice rather than restating them as projections — typed failure outcomes, value-returning updates that never mutate inputs, a real array-backed min-heap with sift-up/sift-down, a backtracking regex matcher with typed compile errors, rejection-sampled bounded randomness, and Gregorian calendar arithmetic verified against known weekday anchors.

Each sample documents where it is narrower than its canonical module (bounded capacity, quote-free CSV/JSON subsets) instead of implying full coverage.

## Gate — `tests/stdlib/kotest/check.sh` (`task kotest`)

- Runs the full green sweep and requires ≥100 tests, so the samples cannot quietly stop being wired.
- Runs a **deliberately failing fixture** by explicit path and asserts exit 1, the `✗` line, the assertion diagnostics, and that execution continued past the first failure — the harness cannot report red as green.
- Checks `--filter` narrows and `--list` enumerates.
- Fixture suites are excluded from directory sweeps, so `kofun unittest` with no arguments stays green while the gate still exercises the red path.

Added to the `verify` list. The `unit-testing` capability row now cites `task kotest`; the canonical `testing.kofun` projection claim stays under `task stdlib`.

## Verification

All run locally on this branch:

| Gate | Result |
|---|---|
| `tests/stdlib/kotest/check.sh` | PASS (133 tests, 18 suites) |
| `stdlib/tests/verify.sh` | PASS |
| `examples/check.sh` | PASS (20 runnable, 8 illustrative) |
| `stdlib/check-capabilities.sh` | PASS (9 mutations refused) |
| `tests/stdlib/clock-adapters/check.sh` | PASS |
| `tests/stdlib/tzdb/check.sh` | PASS |
| `kofun unittest` (default sweep) | 133 passed, exit 0 |

Watch mode was verified by touching a suite mid-run and observing the re-run.

## Not in this PR

The compile failures in the canonical sources are untouched. Their root cause is that those modules call primitives that exist nowhere: `push` is used 60 times across 12 modules, and `push`, `concat`, `reverse`, `text_join`, `to_text`, and `parse_int` are absent from the compiler, the 16-builtin profile, and the spec. (`map`, `filter`, and `fold` do exist, on the x86-64 backend.) Fixing that means either implementing those primitives or rewriting the modules, which is a separate change from adding the test infrastructure.

https://claude.ai/code/session_01WfqFsiAv5Xr3D2X2fkNaq2

---
_Generated by [Claude Code](https://claude.ai/code/session_01WfqFsiAv5Xr3D2X2fkNaq2)_


hjosugi:
Tracking and acceptance criteria are now recorded in #961. This is a size:L candidate spanning the CLI runner, framework, Taskfile, release evidence, and 17 sample suites; please keep it isolated and rebase/regenerate evidence against current main before merge.

hjosugi:
Integration ordering: this candidate overlaps PR #964 in Taskfile.yml, tooling/task-help.mjs, and release evidence. PR #964 has the earlier current-head focused validation and required CI already running. Merge #964 first; then rebase/refresh this PR once, rerun kotest/task-help/release gates, and merge if clean. Please do not race the shared generated files.

Issue #908, opened 2026-08-01
Self-host coverage manifest v2: prove used-by-S, accepted-by-A1, lowered-by-A1, and self-application separately


hjosugi:
## Metadata

- State: ready
- Tracker role: curated child
- Kind: spec-design
- Area: Self Hosting / C11 backend
- Priority: P0
- Size: M
- Parent: #618
- Normative input: #618 refinement items 1 and 2; #751 self-compile gate (completed, `d342920`)
- Audited against: `origin/main@7ce5cb15a3c078eb7c787718dd876fe03504f03b`

## Goal

Replace the single `self_compiler` evidence column with per-class evidence that `bootstrap/selfhost/check-profile.sh` proves by **executing a command**, so the manifest can no longer report a capability that was never run.

This child owns #618's refinement items 1 and 2. It does not implement #622 and does not change canonical `S`.

## Current measured state on main

- `bootstrap/selfhost/profile.tsv` has 46 rows, and the `self_compiler` column holds exactly one distinct value across all of them: `bootstrap/selfhost/driver/S.c`.
- `bootstrap/selfhost/check-profile.sh:119` validates every evidence path with `test -e`. File existence is the whole test.
- The one thing that *is* executed is self-application: `bootstrap/selfhost/check-compiler-driver.sh` compiles the exact canonical `S` bytes with A1 from two directories under different source names, compares C2 byte for byte, and fails with `A1 could not compile S`. That gate landed with #751 and is in the `verify` list as `task selfhost-self-compile`.

So the missing piece is no longer self-application. It is per-row acceptance and lowering evidence.

## Scope

1. A schema change separating four evidence classes per row: `used_by_s`, `accepted_by_a1`, `lowered_by_a1`, `self_application`.
2. For each class, a declared executable command and its expected artifact. A bare path is not evidence.
3. `check-profile.sh` runs the declared command for every row marked `complete`, replacing the `test -e` loop.
4. Reuse of the corpora the driver gate already builds — `bootstrap/selfhost/driver/corpus_{answer,bool,branch,builtins,function,list_text,loop,text}.kofun` and the `corpus_reject_*.kofun` negatives — rather than a second corpus.
5. `self_application` references the existing #751 gate in `check-compiler-driver.sh`; it is not re-implemented.
6. `bootstrap/selfhost/profile.meta` gets a new schema version and keeps rejecting `source_sha256` drift.

## Out of scope

[...]

Issue #814, opened 2026-08-01
Gate failures must name what failed: 460 assertions abort with no output


hjosugi:
## Metadata

- State: ready
- Tracker role: curated
- Kind: planning
- Area: Verification gates
- Priority: P2
- Roadmap milestone: M2-alpha
- Size: L

## Why this exists

**460 assertions in 43 gate scripts abort with no output at all.**

Every gate here is `set -eu`, and the dominant assertion idiom is a bare `test`:

```sh
test "$(wc -c <"$image" | tr -d ' ')" -eq "$expected_size"
```

When that fails, `set -e` exits the script. Nothing is printed. The CI log ends
mid-run, `sh bootstrap/native/check.sh` just returns 1, and the reader is left
to bisect a 2000-line shell script to find which of 118 comparisons went wrong.

This is not hypothetical. #794's *"Two traps worth knowing before you touch
anything"* names it as trap 1, discovered the hard way:

> `bootstrap/native/fixtures/function_text_provenance.txt` pins a SHA-256 of
> `core_compiler.c`. Editing that file at all — even one string literal —
> invalidates the digest, and the native gate fails **silently**: the check is a
> bare `test "$actual" = "$expected"` with no message, so
> `sh bootstrap/native/check.sh` just exits 1.

The repository already knows the right shape. 52 scripts define a `fail()` and
write `... || fail "message"`. The problem is concentrated in the scripts that
never adopted it: **10 of the 13 worst offenders define no `fail()` at all**,
including the four largest gates in the tree.

## The measurement

A silent assertion is a `test` or `[` command that

1. is at top level, not inside a function body where it may be a return value;
2. carries no `||` or `&&` handler; and
3. is not the condition of an `if`.

Under `set -e` each one aborts with an empty stderr. Counted against
`ca26063d`:

| group | count | files | child |
|---|---:|---:|---|
| `bootstrap/native/check.sh` alone | 118 | 1 | #815, with the contract |
| remaining bootstrap gates | 98 | 6 | #816 |
| conformance | 96 | 7 | #817 |
| typed-sidecar, framework, integration | 88 | 10 | #818 |
| spec, diagnostics, fuzz, and the tail | 60 | 19 | #819 |
| **total** | **460** | **43** | |

The single worst file is `bootstrap/native/check.sh` at 118, followed by
`bootstrap/stage2/check.sh` at 55 and
`tests/conformance/modules/re-exports/run.sh` at 44.

## The rule this establishes

[...]

PR #962, opened 2026-08-02
stage2: prove terminal-arm last uses in ensure_move (#904)

### LINKED ISSUE #904: Stage 2 ensure_move v2: prove terminal branch and early-return last uses
Parent: #572

PR #879 established the unstable `compiler.ensure_move(value)` assertion and a conservative same-scope proof. This child advances exactly one deferred control-flow slice: a managed local may be asserted inside a conditional arm only when that arm is terminal, so control cannot return to a point where the outer binding remains observable.

## Scope

- Stage 2 only; immutable local `Text` or `List` bindings.
- Accept an assertion in a conditional arm when every path after it in that arm terminates with an early return.
- Keep non-terminal arm assertions rejected as `branch mismatch`.
- Keep loops, aggregate field sensitivity, general move optimization, allocation counters, foreign calls, and `take`-parameter handoff out of scope.
- Preserve the assertion as compile-time-only and byte-erased.

## Acceptance criteria

- the analysis uses explicit control-flow termination, not source-byte ordering alone;
- positive fixtures cover both arms terminating and one terminal arm whose sibling never touches the binding;
- negative fixtures cover fallthrough after the assertion, a sibling-arm later use, and a nested loop;
- every refusal remains deterministic `E2S146` with `branch mismatch` or `later use`;
- C11 output with and without each accepted assertion is byte-identical and interpreter/C11 observable results match;
- `sh tests/move-assertion/check.sh` and the Stage 2 diagnostic gate pass;
- #572 remains open for loops, aggregates, actual allocation elimination/counters, optimization remarks, foreign-call handling, and `take` handoff.

## Dependency

Unblocked by merged PR #879 at main commit `7ce5cb15`.

hjosugi:
Closes #904.

This PR now contains only the bounded terminal-arm ensure_move slice. The unrelated RFC was reverted earlier, and the obsolete assertions workaround was reverted after #960/#969 repaired main.

It accepts a managed-local assertion in a conditional arm only when every remaining path in that arm returns, keeps fallthrough/sibling-use/loop cases rejected with E2S146, and proves accepted assertions erase from emitted C.

Integration head: 0acccbb9

Validated on current main integration:
- move-assertion gate: PASS (4 accepted shapes, 11 explained refusals, 2 erasure pairs)
- bootstrap Stage 2 gate: PASS
- typed-sidecar Stage 2 events: PASS, including 254 repository companions
- release claims/evidence and SHA256SUMS: PASS
- graphify update: complete

No assertions budget/scanner change remains in the effective diff.


hjosugi:
`Kofun verification` is red, but the failure is on the base branch and predates this PR.

```
FAIL: assertion budget: bootstrap/selfhost/check-profile.sh has 1 silent assertions and no budget entry
The rule is #814: no gate may exit non-zero without naming the check, the expectation, and the observation.
```

Reproduced on plain `origin/main` at `6f71b6d`, with none of this branch's commits applied:

```sh
git checkout origin/main
sh tests/assertions/check.sh
# FAIL: assertion budget: bootstrap/selfhost/check-profile.sh has 1 silent assertions and no budget entry
```

`bootstrap/selfhost/check-profile.sh` arrived in #951 (`6f71b6d`, "selfhost: prove per-class profile evidence by running it (#908)"). This PR does not touch it, `bootstrap/selfhost/`, or `tests/assertions/`.

## Exact cause

`sh tests/assertions/check.sh --count` reports `1  bootstrap/selfhost/check-profile.sh`, and `tests/assertions/budget.tsv` carries no row for that path — it has `bootstrap/selfhost/check-compiler-driver.sh` at `0` and nothing for the new file. The one silent assertion is the multi-line `test` at `bootstrap/selfhost/check-profile.sh:126`:

```sh
test "$(wc -l < "$tmp_dir/expected-inventory")" -eq \
```

Every other `test` in the file pairs with `|| fail "..."`; this one does not reach a named failure.

## Two fixes, and the choice is #951's to make

1. Convert line 126 to `assert_num` from `tests/assertions/assert.sh`, then add a `0` row for the path. This is what #814 asks for, and it is why the budget is exact in both directions.
2. Add `1<TAB>bootstrap/selfhost/check-profile.sh` to `tests/assertions/budget.tsv`, recording the debt instead of clearing it.

I have not applied either. The file landed minutes before this branch rebased onto it, and which of the two is right is a judgement about #951's own gate rather than anything this PR changes.

## This PR's own gates

Run locally on the rebased branch:

| Gate | Result |
|---|---|
| `sh tests/move-assertion/check.sh` | PASS — 4 accepted shapes, 11 explained refusals, 2 erasure pairs |
| `sh bootstrap/stage2/check.sh` | PASS |
| `sh tests/rfc/check-registry.sh` | PASS — 29 decisions, 4 proposed, 21 mutations refused |
| `sh tests/release/check-claims.sh` | PASS — 44 claims, evidence pack current |

`Release evidence pack` is green on CI. I will re-check and push once the base branch recovers; if `tests/assertions` is meant to be fixed here instead, say which of the two options you want and I will apply it.

---

hjosugi:
Correcting my previous comment. The line I named was wrong, and so was one of the two options I offered. The failure is a **false positive in the detector**, not a real silent assertion.

## Root cause

I instrumented `count_file`'s awk to print what it counts. The single count comes from the `pending` branch, at `bootstrap/selfhost/check-profile.sh:202`:

```sh
    has() {
        grep -Eq "$1" "$inventory_work/outer-source"
    }
```

`has()` is a predicate whose `grep -Eq` status *is* its return value — precisely the shape `settle()` exists to exempt:

```awk
# `}` ends the function, and a bare `return` hands the status back
# to the caller — in both cases the candidate was a return value,
# not an assertion.
if (line ~ /^\}/ || t == "return") { pending = 0; return }
```

The exemption misses it because `has()` is **nested inside `derive_inventory()`**, so its closing brace is `    }`, not `}`. `/^\}/` does not match a brace with leading whitespace, so the candidate never gets exempted and is counted.

The opening pattern one line above already allows indentation:

```awk
if (line ~ /^[ \t]*[A-Za-z_][A-Za-z0-9_]*[ \t]*\{/) depth++
else if (line ~ /^\}/ && depth > 0) depth--
```

So `depth` increments for a nested function but never decrements. `check-profile.sh` is the first file to nest a predicate helper, which is why this surfaced now and not earlier.

## Both of my earlier options were wrong

- `check-profile.sh:126` is **not** counted. Its `test` accumulates through the `\` continuation into a buffer containing `||`, which the counter handles correctly. I checked by editing it — the count stayed at 1 — and reverted.
- Adding a `1` row to `budget.tsv` would be wrong regardless: all 48 existing rows are `0`. The file has been driven to zero, and #814's stated plan is to keep it there. A `1` would be the first recorded regression, for an assertion that is not actually silent.

## Why I am not pushing a fix

The obvious repair — relaxing `/^\}/` to `/^[ \t]*\}/` in both places — is **not safe as written**. Depth tracking runs on every line, including lines inside an open quoted string, and `check-profile.sh` embeds several `awk` programs whose bodies contain indented `}`. Relaxing the pattern would decrement `depth` on those, changing counts in ways that have nothing to do with shell functions. Doing it correctly means making depth tracking quote-state aware, which `open_text()` already knows how to do but is currently only consulted for buffer accumulation.

The alternative is to lift `has()` and `keep()` out of `derive_inventory()` so their braces sit at column 0. That works — `inventory_work` is a global assigned inside the function, so dynamic scoping still resolves — but it changes #951's structure.

Detector versus caller is a judgement about two files this PR does not touch, so I have left both alone. Say which and I will apply it in a follow-up, or here if you prefer.

## This PR

Unaffected either way. It touches neither `bootstrap/selfhost/` nor `tests/assertions/`, and the same failure reproduces on plain `origin/main@6f71b6d`. Local gates on the rebased branch: `tests/move-assertion/check.sh`, `bootstrap/stage2/check.sh`, `tests/rfc/check-registry.sh`, and `tests/release/check-claims.sh` all pass; `Release evidence pack` is green on CI.

---

PR #967, opened 2026-08-02
ci: remove the repaired selfhost assertion budget

hjosugi:
Integration follow-up for #952 and #960.

#952 temporarily recorded one silent assertion from bootstrap/selfhost/check-profile.sh so its branch could track the then-current main. #960 then fixed that predicate and brought the actual count back to zero. With both merged, the exact budget correctly refuses the stale allowance.

Remove only that stale budget row.

Validated:
- sh tests/assertions/check.sh — 0 remaining, 48 at zero
- git diff --check
- graphify update .


hjosugi:
Superseded by #966. That PR preserves the repository convention of recording cleaned scripts explicitly at zero, while this duplicate removed the row entirely.

Issue #903, opened 2026-08-01
bindgen-c: verify calling conventions instead of recording a constant


hjosugi:
Parent: #574
Landed baseline: PR #873 / merge `356931b8`.

## Gap

The report currently writes `default (System V AMD64 on the pinned target)` for every emitted function. The ABI probe independently compares sizes, alignments, offsets, and enum values, and `readelf` verifies symbol names, but no independent check proves the accepted function calling convention. This leaves the calling-convention part of the parent ABI-probe criterion unmet.

## Bounded outcome

Make calling convention classification derive from Clang data and independently exercise the accepted convention at the generated boundary. Unsupported/non-default conventions must be reported and skipped, never silently labeled as default.

## Acceptance criteria

- The report’s calling-convention value is derived from the AST/target context rather than a constant assertion.
- A fixture with the pinned target’s default convention crosses the generated boundary with scalar, pointer, and by-value record arguments/results.
- At least one explicit non-default convention/attribute fixture is either correctly modeled and probed or deterministically skipped with a machine-readable reason.
- The gate fails if an accepted declaration’s convention is missing, unknown, or disagrees with the independent C-side fixture.
- Existing size/alignment/offset/enum/symbol probes remain green.

hjosugi:
Completion/duplicate audit: this issue was closed by PR #963 and main merge `28158369934501df6708da555538b306b29de315`.

PR #971 originally implemented the same calling-convention slice independently, but merge commit `747cfa78` takes main’s #963 implementation wholesale and removes the duplicate convention-specific files. #971 is now scoped to #900 and #901 only; no #903 follow-up or reopen is required from that PR.

PR #964, opened 2026-08-02
concurrency: the scoped-parallelism model landed without a gate (#555)

hjosugi:
This PR wires the already-landed scoped-parallelism v1 model into task verify and corrects the surrounding documentation without claiming a production concurrency implementation.

Effective changes:
- add the scoped-parallelism task and verify dependency
- group it beside schedule-trace while preserving the merged kotest task
- correct the concurrency lambda example and settled-vs-open wording
- preserve #943 by refusing only the Rust-style double-pipe spelling
- index the concurrency contracts
- regenerate release evidence from the post-#958/#962 tree

Integration head: bf6ed26f0b971f05b05a25502c8eda4934b76212
Integration base: d06ac21bbd25aefeeea69b855db8fb9c07fcf50e

Validated on the exact integration tree:
- scoped-parallelism: PASS (20 fixtures, 214 assertions)
- kotest: PASS (framework, failure semantics, 133 stdlib sample tests)
- schedule-trace, task-help, RFC registry, repository check: PASS
- release claims/evidence: PASS
- graphify update: complete
- git diff --check: PASS

#555 remains open/blocked for production parsing, ownership checking, scheduling, and backend lowering. This PR only makes the existing executable contract impossible for CI to forget.

PR #971, opened 2026-08-02
bindgen-c: sanitize and fuzz the generated boundary (#900 #901)

### LINKED ISSUE #900: bindgen-c: sanitizer-backed generated-boundary gate
## Metadata

- State: complete
- Active PR: #971 at `e202b86da7571630a6f7c7861e024fd25d37ba14`
- Re-audited: 2026-08-02

Parent: #574
Landed baseline: PR #873 / merge `356931b8`.

## Gap

The current `tests/interop/bindgen-c/check.sh` builds and runs the generated boundary, but it does not compile or execute that boundary under a sanitizer. This leaves the parent acceptance criterion “sanitizer-backed integration tests cross the generated boundary” unmet.

## Bounded outcome

Extend the existing pinned `kbfix` corpus with one deterministic sanitizer gate. Keep the generated raw module and fixture as the exercised boundary; do not add a source translator or safe façade.

## Acceptance criteria

- Compile the fixture library and generated C/Kofun boundary with Clang AddressSanitizer and UndefinedBehaviorSanitizer enabled.
- Run the existing ownership, buffer/length, callback, status, and cleanup paths with leak detection enabled where the runner supports it.
- Treat any sanitizer diagnostic or nonzero process result as a hard failure; retain deterministic stdout comparison.
- Add one negative fixture only if it can prove the sanitizer gate itself is active without weakening the normal passing path.
- The focused gate runs offline and is wired into the repository verification surface.

## Validation

- focused bindgen-c sanitizer gate passes on the pinned Linux/Clang CI target;
- existing `sh tests/interop/bindgen-c/check.sh` remains green.

### LINKED ISSUE #901: bindgen-c: adversarial macro fuzzing and bounded Clang execution
## Metadata

- State: complete
- Active PR: #971 at `e202b86da7571630a6f7c7861e024fd25d37ba14`
- Re-audited: 2026-08-02

Parent: #574
Landed baseline: PR #873 / merge `356931b8`.

## Gap

`bindgen-c` bounds captured Clang output with `maxBuffer`, declaration counts, and typedef depth, and its focused gate rejects one malformed header. It does not set a Clang subprocess timeout and does not fuzz adversarial macro expansion. The parent criterion “malformed headers and adversarial macro expansion are bounded and fuzzed” is therefore only partially satisfied.

## Bounded outcome

Add deterministic resource bounds around every Clang invocation and a small committed adversarial macro/header corpus that exercises expansion depth/volume and malformed constructs without a shell.

## Acceptance criteria

- Every version, triple, AST, and preprocessing Clang subprocess has an explicit deterministic timeout and bounded captured stdout/stderr.
- Timeout/output-limit failures produce a stable nonzero diagnostic and leave no partial generated artifacts.
- A committed offline corpus covers recursive/deep expansion, high fan-out/token volume, hostile diagnostics, and malformed declarations.
- A bounded deterministic fuzz driver mutates macro definitions/header tokens within fixed seed, case-count, byte, and time budgets.
- Inputs never reach a shell; generated paths and diagnostics do not escape the requested output boundary.

## Validation

- focused adversarial/fuzz gate passes repeatedly with the same seed and outcome;
- existing `sh tests/interop/bindgen-c/check.sh` remains green.

hjosugi:
Closes #900.
Closes #901.

#903 landed independently in #963. This PR keeps main’s calling-convention implementation and does not add the duplicate `check-convention.sh`, `make-convention-probe.mjs`, or `kbfix_cc.h` files.

## What changes

- Adds `check-sanitizers.sh`: builds both `libkbfix.so` and the generated boundary with ASan+UBSan, covers the generated Kofun driver plus buffer/callback C probes, and proves the gate is armed with a negative fixture.
- Routes the intermediate public `bin/kofun build` link through a temporary wrapper for the selected sanitizer compiler. This supplies the ASan/UBSan runtimes that were missing at PR head while keeping the later explicit sanitized rebuild and assertions intact.
- Adds `check-fuzz.sh`, a deterministic adversarial macro corpus, and a fixed-seed mutation driver. Every Clang call goes through one argv-only helper with wall-clock/output bounds; timeout, output-bound, declaration-bound, deterministic acceptance, named refusal, and no-partial-output behavior are exercised.
- Integrates both sibling gates into `tests/interop/bindgen-c/check.sh`.

## Scope

The final diff is limited to the #900/#901 bindgen-c tests, fixtures, corpus, documentation, and the Clang bounds in `tooling/bindgen-c/bindgen-c.mjs`. It intentionally contains no Taskfile/task-help or release-claim/evidence changes.

## Validation on current main

- `sh tests/interop/bindgen-c/check-sanitizers.sh` — 4 PASS, including `detect_leaks=1` and the armed negative fixture
- `sh tests/interop/bindgen-c/check-fuzz.sh` — 4 PASS; 42 deterministic-or-named-refusal cases
- shell syntax checks for all added/changed `.sh` files
- `node --check` for the bindgen tool and fuzz driver
- `nix shell nixpkgs#go-task -c task repository-check` — PASS
- `git diff --check` — clean


hjosugi:
## Read-only integration audit at `747cfa7851f2553f54fb7f045698bf06dbf5d98f`

### Stop condition: #900 focused gate fails

On a clean detached worktree at the exact PR head:

```text
$ sh tests/interop/bindgen-c/check-sanitizers.sh
FAIL: bindgen-c sanitizers: generated module did not build in the C ABI profile
/usr/bin/ld: libkbfix.so: undefined reference to `__asan_register_elf_globals`
/usr/bin/ld: libkbfix.so: undefined reference to `__asan_report_load8`
... ASan/UBSan runtime symbols omitted ...
collect2: error: ld returned 1 exit status
```

The script selects `gcc` after the installed Clang cannot link compiler-rt, builds a sanitizer-instrumented `libkbfix.so`, then invokes ordinary `kofun build --link-library` before recompiling the emitted C with sanitizer flags. That intermediate ordinary link does not carry the sanitizer runtime and fails. This is a product/gate defect, not the previously documented assertion-budget failure.

**Resume when:** the script obtains emitted C without linking the sanitized `.so` through an unsanitized compiler, or runs that intermediate build with the selected sanitizer compiler and runtime flags; then the positive driver, C probe, negative armed fixture, focused gate, and exact-head CI all pass.

### #901 state

Shell/Node syntax checks pass, and the patch structurally covers the issue contract: one bounded `execFileSync` site, deterministic timeout/output diagnostics before output-directory creation, committed hostile corpus, fixed seed/budgets, accept/refuse vacuity guards, repeat-summary comparison, and a hanging-compiler exercise. I did not run the full 128 MiB expansion/fuzz workload locally because it is intentionally memory-heavy; exact-head CI remains the required execution evidence.

### #903 state

#903 was already completed by PR #963 / main merge `28158369934501df6708da555538b306b29de315`. Merge commit `747cfa78` correctly takes main’s implementation and removes this branch’s duplicate convention files. #971 must close only #900 and #901.

### Current main and real competing files

The PR is based through `78626dca`; current main is one commit ahead and changes only `tests/backlog/check.mjs` and `tests/backlog/debt.tsv`. Therefore there is **no present file overlap with main**, and GitHub reports the PR mergeable.

The conflict risk is other open PRs, not current main:

- #964: `Taskfile.yml`, `artifacts/release-evidence/index.json`, `tooling/task-help.mjs`
- #962: `artifacts/release-evidence/index.json`
- #958: `Taskfile.yml`, `artifacts/release-evidence/index.json`, `tooling/task-help.mjs`
- #954: `Taskfile.yml`, `artifacts/release-evidence/index.json`

The actual #900/#901 implementation files under `tests/interop/bindgen-c/` plus the timeout change in `tooling/bindgen-c/bindgen-c.mjs` do not overlap those PRs.

### Smallest safe landing shape

1. Fix the #900 intermediate sanitizer-link failure.
2. Keep only the two issue implementations and their existing `tests/interop/bindgen-c/check.sh` integration.
3. Drop `bindgen-c-hardening` from `Taskfile.yml` and `tooling/task-help.mjs`: `task bindgen-c` already runs both sibling gates, so the extra convenience target is not acceptance-critical and creates avoidable conflicts.
4. Move the capability-matrix correction (`docs/MVP_IMPLEMENTED.md`, `release/claims.json`, generated `artifacts/release-evidence/*`) to a separate narrowly named PR after the active release-evidence PRs settle. It is valid drift cleanup but not part of #900/#901.
5. Rebase on then-current main, run focused #900/#901 gates serially, then exact `task bindgen-c repository-check release-claims` and full CI.

hjosugi:
Correction to the compiler-selection detail in the audit above: on this exact host, both sanitizer link probes succeed, so the script selects **Clang** for the instrumented library. `bin/kofun` then uses its default `cc`, which is **GCC 16.1.1**, for the intermediate ordinary link. The blocker is therefore the Clang-instrumented shared library being linked by an unsanitized GCC invocation without the required runtime—not a fallback selection of GCC inside the sanitizer script. The resume condition is unchanged.

hjosugi:
Acknowledging the integration hold on #989 and doing the "no unique acceptance evidence is lost" check it asks for. This PR is broader than mine on every axis I compared — the compiler probe with an explicit no-skip failure, the `__asan_` symbol assertion on both binaries, and the buffer/callback paths the Kofun driver cannot reach are all coverage #989 does not have. One gap is worth closing before it merges.

**`check-sanitizers.sh` proves one arm is armed, and claims three.**

`kbfix_negative.c` is a heap-buffer-overflow, and the assertions require `AddressSanitizer`, `kbfix.c`, and `heap-buffer-overflow`. That proves AddressSanitizer is linked and reporting. It does not prove the other two the gate depends on:

- **LeakSanitizer.** The header says *"A runner that cannot run LeakSanitizer fails loudly here rather than passing quietly"*, but nothing in the gate leaks. `detect_leaks=1` is a request; on a platform where LSan is unavailable it is a startup warning, not an error, and every check above stays green. The client-owned-handle contract the comment calls the point of this gate would then be unchecked.
- **UndefinedBehaviorSanitizer.** Drop `,undefined` from `SANITIZE` and the negative fixture is still caught by ASan, so the gate stays green with UB checking silently gone.

Two more fixtures in the same shape close it — an unfreed `kbfix_counter_new` asserting `LeakSanitizer: detected memory leaks`, and a signed overflow asserting `runtime error: signed integer overflow`. Both are a few lines and cost milliseconds. That is the whole of what #989 has and this does not; #989 carries three probes for exactly this reason, one per arm, and I measured each being diagnosed by its own sanitizer before relying on any of them.

Everything else in #989 is subsumed here, so once this passes exact-head CI I will close #989 as superseded and delete its branch, per the ordering in the hold.

One measured note that may save a CI cycle, since the description mentions runtimes missing at PR head: on Ubuntu the clang sanitizer runtime is a separate package (`libclang-rt-18-dev` for clang 18) — clang itself installs without `libclang_rt.asan-x86_64.a`, and the link fails with a bare `ld: cannot find` line. The compiler-probe loop here handles that by falling through to gcc, which is the right behaviour; it is worth knowing that the fallthrough is what will happen on a runner without the package, so the gate reports gcc rather than clang in its PASS line.

---

PR #972, opened 2026-08-02
gates: record repaired selfhost assertions at zero

hjosugi:
Integration follow-up after #969.

The assertion gate is green, but #969 removed bootstrap/selfhost/check-profile.sh from tests/assertions/budget.tsv. The budget file explicitly requires a zero row for every cleaned script so the improvement is recorded rather than left untracked.

This branch was merged with current main; the effective diff is one explicit zero row.

Validated on the merged tree:
- sh tests/assertions/check.sh — 0 remaining, 49 at zero
- sh bootstrap/selfhost/check-profile.sh
- sh tests/backlog/check.sh
- sh tests/backlog/check-stamps.sh
- task release-evidence
- graphify update .


hjosugi:
Superseded by #960 and #969: main now has zero silent assertions and no stale check-profile budget entry. This branch would only re-add a redundant zero row.

PR #976, opened 2026-08-02
spec: exclude top-level statements from the grammar (#955)

### LINKED ISSUE #955: modules: spec/grammar.ebnf admits a top-level `let` that Stage 2 refuses with E2S02
Parent: #289. Filed as #289's single next action by the #911 re-audit.

State: in-progress.

## The divergence

`spec/grammar.ebnf` admits a module-level binding. The compiler does not.

```ebnf
program     = separators, { declaration, separators }, EOF ;              (* :3  *)
declaration = law_family_decl | law_impl_decl | law_check_decl
            | record_type_decl | function_decl | meta_function_decl
            | statement ;                                                 (* :5-11 *)
statement   = let_stmt | return_stmt | while_stmt | for_stmt | take_stmt  (* :83 *)
let_stmt    = "let", [ "own" ], [ "mut" ], [ "own" ], identifier, ...     (* :92 *)
```

`declaration` includes `statement`, and `statement` includes `let_stmt`, with no note restricting `statement` at program level. So the grammar admits a top-level `let` — including `let own`, a module-level owned resource.

Stage 2 refuses it:

```
$ cat toplevel_let.kofun
let banner: Text = "module level"

fn main() -> Int {
    print(banner)
    return 0
}

$ ./bin/kofun check toplevel_let.kofun
error[E2S02]: expected top-level `fn` or `type` at byte 0
```

Audited against `origin/main@8ba3a57c0ae91c8ba3fa97396d698fb264a189cd`.

## Why this blocks #289

#289's coverage table has a `module/global initialization` row whose owner, #299, is closed as **not planned**. The re-audit needed a reason the row has no executable subject, and the obvious candidate — "the grammar declares no module-level binding form" — is false. The row is unowned because the grammar declares a form that Stage 2 does not implement, which is a different problem with a different fix.

Until this is settled there is nothing to say about initialization *order*: the question of when a module-level binding is initialized cannot be posed while it is undecided whether one can be written at all.

## The decision this issue owes

Exactly one of:

1. **The grammar is authoritative.** Module-level `let` is intended; `E2S02` is a Stage 2 gap, and the initialization row gets a real owner and a bounded slice.
2. **Stage 2 is authoritative.** Module-level bindings are not in the language today; `spec/grammar.ebnf` is corrected so `program` admits only the declaration forms that exist, and the initialization row is recorded as having no subject with that as its stated reason.

The earlier premise for preferring option 2 is false on current `main`: 15 canonical stdlib `.kofun` sources use top-level `let`, and existing gates treat those files as target-language surfaces ahead of Stage 2. Option 2 therefore requires an explicit migration/reclassification of those sources before narrowing the grammar. Option 1 remains a language decision requiring its own RFC and a named #289 implementation owner. Until one authority model is selected, neither grammar-only alternative PR is mergeable.

## Scope

- Decide 1 or 2 and record it.
- Under option 2: restrict `program`/`declaration` in `spec/grammar.ebnf` to the forms Stage 2 accepts, and add a compile-fail fixture pinning `E2S02` for a top-level `let` so the grammar and the compiler cannot drift apart again unnoticed.
- Under option 1: file the RFC; do not change the grammar here.

## Out of scope

- Initialization order, use-before-initialization, and cyclic initialization. Those stay with #289's initialization row and only become answerable once this is decided.
- Macro-staged declarations.
- Any other `spec/grammar.ebnf` production.

## Acceptance criteria

- [ ] The grammar and Stage 2 agree on whether a top-level `let` is a valid program, and the agreement is stated in exactly one place.
- [ ] Under option 2, a compile-fail fixture asserts `E2S02` for a top-level `let`, and it runs under an existing gate rather than in an unreferenced directory.
- [ ] `spec/grammar.ebnf`'s header no longer describes a stage whose behaviour the repository cannot reproduce, or the mismatch is explicitly recorded.
- [ ] #289's `module/global initialization` row cites this issue or its outcome, not #299.

## Note on the grammar's self-description

`spec/grammar.ebnf` line 1 reads `(* Kofun Stage 0 grammar, draft 0.2-bootstrap. *)`. Stage 0 is the removed reference prototype — `docs/MEMORY_MODEL.md` §13 records that its statements are historical and not capabilities of the current compiler. Whether the grammar is meant to describe Stage 0, Stage 2, or the target language is itself undecided, and it is the likely root of this divergence.

## Blocking decision (2026-08-02)

- PR #976 implements option 2 and would make the 15 canonical top-level-binding sources ungrammatical.
- PR #978 preserves the broad grammar, effectively option 1, but supplies neither the required RFC nor an implementation owner.
- Resume only after either (A) an RFC accepts module-level bindings and #289 gets an owner, or (B) the canonical sources are deliberately migrated/reclassified before `statement` is removed from program-level declarations.

hjosugi:
## Summary

- make `top_level_declaration` the only program-level production and exclude statements such as `let`
- describe `spec/grammar.ebnf` as the full-language draft instead of the removed Stage 0 profile
- bind the existing E2S02 top-level-let fixture to the grammar contract
- fail the Stage 2 diagnostics gate if the grammar admits program-level statements again

## Boundary

This selects issue #955 Option 2. It does not change the Stage 2 compiler, runtime, Taskfile, module initialization semantics, or macro-staged declarations.

## Validation

- `sh tests/diagnostics/stage2/run.sh` — 47/47 pass
- mutation check: temporarily restoring `| statement` is refused with `diagnostics: grammar admits a top-level statement that Stage 2 rejects`
- `nix shell nixpkgs#go-task -c task repository-check`
- `graphify update .`
- `git diff --check`

Closes #955

### DIFF
--- a/spec/grammar.ebnf
+++ b/spec/grammar.ebnf
@@ -1,14 +1,14 @@
-(* Kofun Stage 0 grammar, draft 0.2-bootstrap. *)
+(* Kofun full-language grammar, draft 0.3-bootstrap. Executable bootstrap
+   coverage is tracked separately; this is not the removed Stage 0 profile. *)
 
-program          = separators, { declaration, separators }, EOF ;
+program          = separators, { top_level_declaration, separators }, EOF ;
 
-declaration      = law_family_decl
+top_level_declaration = law_family_decl
                  | law_impl_decl
                  | law_check_decl
                  | record_type_decl
                  | function_decl
-                 | meta_function_decl
-                 | statement ;
+                 | meta_function_decl ;
 
 (* Nominal records: see spec/records-v1.md. A record is constructed by the
    labelled call form below, never by `Name { ... }`, so no expression
--- a/tests/diagnostics/stage2/e2s02_top_level.kofun
+++ b/tests/diagnostics/stage2/e2s02_top_level.kofun
@@ -1,4 +1,5 @@
 # diagnostic-mode: compile
 # expect-code: E2S02
-# expect-span: byte 71
+# contract: spec/grammar.ebnf top_level_declaration
+# expect-span: byte 124
 let answer = 42
--- a/tests/diagnostics/stage2/e2s02_top_level.stderr
+++ b/tests/diagnostics/stage2/e2s02_top_level.stderr
@@ -1,1 +1,1 @@
-error[E2S02]: expected top-level `fn` or `type` at byte 71
+error[E2S02]: expected top-level `fn` or `type` at byte 124
--- a/tests/diagnostics/stage2/run.sh
+++ b/tests/diagnostics/stage2/run.sh
@@ -6,6 +6,25 @@ SUITE="$ROOT/tests/diagnostics/stage2"
 WORK=${KOFUN_DIAGNOSTIC_WORK:-"$ROOT/build/diagnostics-stage2"}
 . "$ROOT/bootstrap/stage2/build.sh"
 
+PROGRAM_PRODUCTION=$(sed -n '/^program[[:space:]]*=/p' "$ROOT/spec/grammar.ebnf")
+TOP_LEVEL_PRODUCTION=$(
+    sed -n '/^top_level_declaration[[:space:]]*=/,/;/p' \
+        "$ROOT/spec/grammar.ebnf"
+)
+printf '%s\n' "$PROGRAM_PRODUCTION" |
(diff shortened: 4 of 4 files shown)


hjosugi:
Closing unmerged after the #955 readiness re-audit. Current main has 15 canonical top-level-`let` sources, so this grammar-only alternative cannot settle the authority decision by itself. #955 is now `blocked` with explicit resume conditions; no implementation evidence from this PR is being claimed.

Issue #574, opened 2026-07-20
V: generate audited C bindings from Clang AST before attempting source translation


hjosugi:
## Product disposition — closed as not planned (2026-08-08)

The current roadmap is pruning this V-comparison planning tracker. Closing it is a product-scope decision, not a claim that every historical acceptance item is implemented.

Preserved concrete Kofun work:

- the shipped bounded `bindgen-c-stage1` checkpoint and its release evidence remain unchanged;
- #902 remains open `needs-decision` for the mechanical raw-module trust authority and explicit crossing; and
- the recorded `--target TRIPLE` defect remains valid evidence but is no longer owned as roadmap work by this comparison tracker.

No `kofun migrate-c` source translator is scheduled or implied by this closure.

## Metadata

- State: blocked
- Blocked by: #902
- Tracker role: planning parent for audited C bindings before any source translator
- Re-audited against: `origin/main@4736fbf5e92375bdf6ce10ec63cf4164f65e5e9a` (2026-08-08 JST)
- Shipped Stage 1 baseline: PR #873 / `356931b8c8f8d399d7b8d3aa84421bcec99578f6`
- Current release claim: `bindgen-c-stage1` (`checkpoint`; raw/trusted, not safe)
- No open PR currently overlaps `tooling/bindgen-c`, its fixture corpus, or the module/KIF raw-trust boundary.

## Goal and sequencing

V provides C-wrapper generation and Clang-AST-assisted C source translation. Kofun should keep those as two distinct trust boundaries:

```text
kofun bindgen-c <headers>       -> raw extern/repr(C) module + audit report
kofun migrate-c <sources>       -> later trusted draft + unsupported report
```

The first command now exists as a bounded Stage 1 checkpoint. The second does not exist and must not begin by relabelling C aliasing, pointer arithmetic, unions, preprocessor state, unchecked integer behavior, signals, or lifetime conventions as safe Kofun.

Sources: [V documentation: V and C / translating C to V](https://docs.vlang.io/v-and-c.html), [V repository](https://github.com/vlang/v).

## Measured current Stage 1

At the exact audited commit, `kofun bindgen-c`:

- invokes Clang through structured argv with fixed timeout/output/diagnostic bounds;
- generates deterministic `NAME.raw.kofun` declarations and a deterministic `NAME.bindgen.json` report;
- records Clang version, effective target triple, language standard, defines, include paths, sysroot, header hash, and a digest over the interpretation context;
[...]

PR #989, opened 2026-08-02
bindgen-c: run the generated crossing under ASan and UBSan (#900)

### LINKED ISSUE #900: bindgen-c: sanitizer-backed generated-boundary gate

hjosugi:
Closes #900.

## What changed

`tests/interop/bindgen-c/check.sh` gains one section, and the corpus README gains two evidence rows. No new corpus, no source translator, no safe façade, no change to the generator or the report schema.

The gate already proved the generated boundary builds, links, runs, and reproduces `driver.stdout`. None of that sees the faults a generated binding actually has: a length that is right for the header and wrong for the call, a signedness that silently wraps, an owner that frees twice. #574's criterion *"sanitizer-backed integration tests cross the generated boundary"* was the one with no evidence behind it.

The same crossing now runs a second time with **both sides** rebuilt under AddressSanitizer and UndefinedBehaviorSanitizer, leak detection on, and must reproduce the same golden with empty stderr. The fixture is compiled in directly rather than linked as the shared object the first build uses — an uninstrumented library would leave half the crossing unchecked.

`-fno-sanitize-recover=all` is what makes this a gate rather than a report. UBSan prints and continues by default, so a diagnosed program still exits 0 and a check that reads only the exit status says PASS.

## Proving the sanitizers are armed

A sanitizer that was never linked in is silent in exactly the way a clean run is, so a green result on its own proves nothing. Three probes are built with the same flags and each must be diagnosed by name:

| Probe | Must be caught by |
|---|---|
| use-after-free | `AddressSanitizer: heap-use-after-free` |
| unfreed allocation | `LeakSanitizer: detected memory leaks` |
| signed overflow | `runtime error: signed integer overflow` |

They live in the gate's work directory, not the corpus, so the passing path is untouched.

## Proved against a real fault

Under-allocating the counter by one byte in the fixture:

```c
counter = calloc(1U, sizeof(*counter) - 1U);
```

passes generation, the ABI probe, the symbol check, the `--backend c --c-abi` build, the link, the run, **and** the `driver.stdout` comparison without a murmur. It fails in the new section:

```
bindgen-c: FAIL: the sanitized boundary exited nonzero:
  fixture/kbfix.c:32:14: runtime error: member access within address 0x503000000040
  with insufficient space for an object of type 'kbfix_counter_t'
```

Reaching that message means every earlier step had already passed, which is the point: this is evidence the gate did not previously have. The fixture is unmodified in this PR — the injection was a check, and it was reverted.

A double free is *also* caught, but by glibc in the ordinary build, so it does not demonstrate anything about this section.

## Verification

| Check | Command | Result |
|---|---|---|
| Focused | `sh tests/interop/bindgen-c/check.sh` | 8 PASS lines, including the new one |
| Assertion budget | `sh tests/assertions/check.sh` | `0 remaining, 48 at zero` — no silent assertion added |
| Whitespace | `git diff --check` | clean |
| Full regression | `task verify` | **exit 0**, 1015 PASS lines, 0 failures — run to completion before opening this PR |

Gate cost: 2.5s → 2.9s.

## Host requirement

The section needs clang's sanitizer runtime (`libclang_rt.asan`), which is separate from clang itself on Debian/Ubuntu (`libclang-rt-<version>-dev`). When it is absent the build fails with clang's own linker error, and the failure message says to check for it rather than reporting a boundary fault. CI's Ubuntu runner is the target; if it turns out to lack the package, the fix is one line in the workflow rather than a weaker gate.

---


hjosugi:
Integration hold: PR #971 is already the active combined #900/#901 lane and changes the same `tests/interop/bindgen-c/check.sh` and README paths while adding the dedicated sanitizer and bounded-fuzz gates. Both candidates are currently mergeable with exact-head CI running, so merging both would duplicate #900 and make the second branch conflict.

Ordering: wait for #971 exact-head CI. If #971 passes and merges, close this PR as superseded and delete its head branch after verifying no unique acceptance evidence is lost. If #971 fails, retain this narrower #900 candidate and reassess it independently. No code or branch was changed by this hold.

hjosugi:
Closing as superseded by merged PR #971 (). #971 satisfies #900's exact sanitizer contract with both sides instrumented, leak detection, deterministic golden output, and an armed negative fixture, and also lands the coupled #901 bounded-fuzz gate. Keeping both would duplicate the same check/README paths.

hjosugi:
Correction: the exact merged PR #971 commit is `00ce08f093da2706156f88f80d8c2fb366006d35`. #971 satisfies #900's sanitizer contract with both sides instrumented, leak detection, deterministic golden output, and an armed negative fixture, while also landing #901's bounded-fuzz gate. The superseded close and head-branch deletion stand.

Issue #920, opened 2026-08-01
ReuseCandidate v1: a normative reuse-candidate record with layout/uniqueness evidence, remarks, and a validator


hjosugi:
## Metadata

- State: verification-pending
- Tracker role: curated child
- Kind: spec-design
- Area: Optimizer / Memory / Types
- Priority: P2
- Size: M
- Parent: #576
- Normative input: #572's landed move assertion (`docs/MEMORY_MODEL.md` §14, `E2S146`), `spec/aggregate-layout-v1.md`, `tooling/typed-sidecar/ownership-view-v1.md`
- Audited against: `origin/main@7ce5cb15a3c078eb7c787718dd876fe03504f03b`

## Goal

Land the honest first slice of #576: a **normative, executable contract** for how
a constructor reuse candidate is recorded — its source constructor, layout
evidence, uniqueness evidence, eligibility state, and the remark that explains
success or failure — validated by a runnable checker over committed golden
vectors.

This is #576's own first acceptance criterion ("MIR/IR represents constructor
reuse candidates with source, layout, and uniqueness evidence"), separated from
every criterion that requires performing the rewrite.

### Why a contract and not a compiler pass, at this SHA

I audited whether the analysis itself could land now. It cannot, and the reason
is worth recording so nobody re-derives it:

- The ordinary Stage 2 backend **allocates nothing on the heap.** Its generated
  prelude in `bootstrap/stage2/compiler.c` includes no `<stdlib.h>` and no
  `malloc`; `kofun_text_slice` serves slices from a fixed `static char
  slots[64][32]` table. `kofun_rt_alloc` exists only in the selfhost-C11 lane
  prelude (`sl_prelude_text`), a different pipeline.
- There are **no heap-allocated ADT constructor nodes** to match and rebuild.
  The Core enum representation is a by-value `KofunEnumValue {tag, payload}`.
- `List[Int]` is **not yet a Core binding type at all** — that is #919, still
  open and `ready`.

So there is no match/rebuild pair in existence to instrument. Writing the
contract first is what `spec/aggregate-layout-v1.md` already does for layout:
it is normative and executable, defines the record only, and explicitly does
not change code generation. This child follows that precedent exactly.

## Scope

1. A normative document `spec/reuse-candidate-v1.md` defining the record: the
   matched source constructor and span, the constructed target, layout evidence
   (size, alignment, field count, scanning shape) expressed in terms of the
[...]

Issue #919, opened 2026-08-01
Stage 2 C11 List[Int] values: locals, len, and indexing


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated child
- Kind: implementation
- Area: C11 backend / Type system
- Priority: P2
- Size: M
- Parent: #868
- Sibling increment: #891 (Text results, complete via PR #1041)
- Blocked by: none.
- Completed prerequisites: #891 and #1038 via PR #1041 and PR #1045
- Audited against: `0a209468da3698a31cf6177b863284fd04c9a374`

## Goal

Make `List[Int]` reach the Stage 2 type namespace at all, so a bounded integer
list can be bound as a local, counted with `len`, and read by index through
checked source -> typed HIR -> executable C11.

This is the second of the five increments #868's measurement identified, and
the first one that is not already owned: increment 1 (`Text` results) is #891.
Increments 3-5 depend on this one.

## Current boundary

On `7ce5cb1` the `List[Int]` annotation is not a Core binding type, and the
three ways a program reaches that boundary each report something other than the
boundary:

| Source | Reported today | Actual cause |
| --- | --- | --- |
| `let v: List[Int] = [1, 2]` | `E2S35` unknown lexical binding `Int` (`bootstrap/stage2/compiler.c:10438`) | `List[Int]` is not a Core binding type |
| `f([1, 2, 3])` | `E2S17` arity `got -1` | `argument_end` (`:4240`) does not span a list literal |
| `let v = [1, 2]` then `len(v)` | `E2S15` `len` expects `TextOrList`, got `Int` (`:5603`, `:5686`) | the list literal was inferred as `Int` |

Only the parameter position states the boundary truthfully today
(`:6273` — "Core parameters must have type Int, Text, a concrete enum, ...").
Correcting these three is on the path this increment has to touch anyway, so
they belong here rather than in a separate cleanup.

## Scope

1. Admit `List[Int]` as a Core binding type so a local annotation resolves
   instead of reporting an unknown binding.
2. Lower a bounded `List[Int]` literal to a deterministic C11 representation
   with an explicit capacity bound.
3. Type and lower `len(list)` and index reads with checked bounds.
4. Make `argument_end` span a list literal so call arity is computed rather
   than abandoned, and correct the three diagnostics in the table above to
   state the real boundary.
5. Reject oversized lists, out-of-range indexes, and unsupported element types
   with stable diagnostics and no partial artifact.
[...]

Issue #576, opened 2026-07-20
Koka: reuse matched ADT constructors in place when storage is unique


hjosugi:
## Product disposition — closed as not planned (2026-08-08)

The current roadmap is pruning this Koka-comparison optimization parent. Closing it is a product-scope decision, not evidence that constructor storage is reused.

Preserved concrete Kofun evidence and owners:

- #920 and #1005 remain completed contracts for `kofun.reuse-candidate/v1`;
- AggregateLayout and the ReuseCandidate validator/golden gates remain active repository evidence;
- #570, #572, and #868 remain open under their own Kofun scopes; and
- no producer, rewrite, heap-ADT representation, backend guarantee, or allocation-performance claim is created by this closure.

A future in-place reuse proposal must be filed as a new bounded Kofun artifact with current storage/ownership/backend evidence instead of reopening this prior-art tracker by default.

### Metadata

- State: needs-detail
- Tracker role: curated optimization-design parent
- Kind: spec-design
- Area: Optimizer / Memory / Types
- Priority: P2
- Milestone: M2-alpha
- Size: M for each future child; the remaining outcome is not one reviewable change
- Audited against: `4736fbf5e92375bdf6ce10ec63cf4164f65e5e9a`
- Blocked by: none

The former sole blocker #920 is closed through PR #977, and its follow-up #1005 is closed through PR #1003. Their ReuseCandidate contract is complete. This parent cannot remain `blocked` on a closed child, but it is not `ready`: the next compiler artifact has no bounded issue, no selected allocating ADT pipeline, and no producer or backend capable of honouring the contract.

The correct restart is refinement into independently reviewable children. This audit does not choose a heap representation, uniqueness algorithm, backend, or owned-field policy.

### Completed contract evidence

`kofun.reuse-candidate/v1` is a normative optimization-record contract. It records matched and constructed constructors, AggregateLayout evidence, constructor-storage uniqueness evidence, observer state, evaluate-all-then-write ordering, backend support, one eligibility state, one refusal reason, and the unstable remark. It explicitly does not add a compiler pass, allocation, mutation, or runtime representation.

Measured on exact `origin/main@4736fbf5e92375bdf6ce10ec63cf4164f65e5e9a`:

```sh
task reuse-candidate
# PASS: state, reason, provenance, observer, ordering, and remark vocabularies are closed
# PASS: every observer, unsafe ordering, and unhonourable backend disposition refuses a static guarantee
# PASS: nine canonical vectors are accepted
[...]

Issue #570, opened 2026-07-20
Austral: propagate ownership kind structurally through generics and ADTs


hjosugi:
## Metadata

- State: blocked
- Blocked by: an unowned Stage 2 owned-binding parse/type carrier; no issue number exists at the evidence commit
- Tracker role: curated parent
- Kind: spec-design
- Area: Ownership / Type system / Memory
- Priority: P1
- Roadmap milestone: M2-alpha
- Size: M; implementation must still split behind its carrier
- Evidence commit: `b8bcaa9b24281ad8ec4e1300c8a444cffe352331`
- Completed design child: #907 / RFC-0004 proposal
- Open implementation child: none

## Goal

Make ownership kind a structural compiler fact across records, tuples, ADTs, closures, `Optional[T]`, `List[T]`, and generic substitution, so an owned cleanup obligation cannot disappear inside a managed composite.

RFC-0004 defines the proposed two-kind lattice and diagnostic path vocabulary. This parent still owns the route from that proposal to executable classification; a proposal, typed-only generic frontend, or Int/Bool aggregate checkpoint is not an implementation.

## Current behavior and evidence

Measured on exact `origin/main@b8bcaa9b24281ad8ec4e1300c8a444cffe352331` on 2026-08-08.

### Owned-binding carrier

The minimal literal probe still does not parse as an owned binding:

```kofun
fn main() -> Int {
    let own x = 1
    return 0
}
```

```sh
./bin/kofun check /tmp/kofun-570-owned.kofun
# exit 1: error[E2S11]: expected `=` at byte 31

./bin/kofun build /tmp/kofun-570-owned.kofun --emit-c /tmp/owned.c
# exit 1: the same E2S11; no accepted C artifact
```

The checked-in `unsupported_owned_binding.kofun` is not a sufficient carrier probe: its RHS calls unimplemented `open()`, so current check/build/run stop first at `E2S16 unknown Core function open`. Replacing that RHS with literal `1` exposes the actual `let own` parse boundary above.

### Decision lifecycle

```sh
task rfc-registry
# PASS: 31 decisions (20 accepted, 5 implemented, 6 proposed)
# RFC-0004: proposed; opened 2026-08-02; review closes 2026-08-16; no implementation record
```

[...]

PR #1003, opened 2026-08-02
spec: require complete ReuseCandidate evidence (#1005)

### LINKED ISSUE #1005: ReuseCandidate v1: require observer, write-ordering, and backend evidence
## Metadata

- State: complete
- Parent: #576
- Follow-up to completed contract slice: #920 / PR #977
- Active PR: #1003
- Audited base: `486d72deb17a6be11325a425e1138822d058121d`

## Problem

The landed `kofun.reuse-candidate/v1` record names refusal reasons for live observers, unsafe destructive-write ordering, and unsupported backends, but a producer can still claim `reuse_statically_guaranteed` without recording evidence for those three conditions. A reason vocabulary alone is not proof.

## Scope

- Add explicit observer evidence for borrowed views, closure captures, weak references, foreign pointers, and aliases.
- Require evaluate-all-then-write ordering before a static guarantee.
- Bind each record to a registered backend and forbid silently weakening a guarantee to allocation.
- Refuse `compiler.ensure_move` as constructor-storage uniqueness evidence.
- Add committed valid/invalid vectors and mutation checks for each new obligation.

No compiler pass, in-place rewrite, allocation elision, runtime, or backend implementation is in scope. #576 remains open for those capabilities.

## Acceptance criteria

- [ ] Every static guarantee records all observers absent, safe write ordering, and a backend disposition that either honours the guarantee or rejects the program.
- [ ] Unknown/live observers, interleaved/unknown writes, contradictory backend support, silent weakening, and `ensure_move`-as-uniqueness are rejected with stable named errors.
- [ ] The validator recomputes AggregateLayout evidence and refuses hand-edited or non-canonical vectors.
- [ ] The focused ReuseCandidate gate, AggregateLayout gate, assertion budget, release claims, RFC registry, and full regression pass on the exact PR head.

## Integration

PR #1003 owns this issue. It is path-disjoint from the current kotest and bindgen sanitizer lanes. Merge only after its exact-head CI passes; completion does not reopen or rewrite #920.

hjosugi:
## What this is

#920 landed in `41b531f7` while this lane was auditing it: the record, the three
eligibility states, the closed reason vocabulary, the provenance enum, the
recomputing validator, eight valid vectors and six invalid ones. That work is
sound and this change keeps all of it.

What it does not do is make three of the conditions #576 requires of a static
guarantee into *evidence*. They are spelled as reasons — `borrowed-view`,
`closure-capture`, `possible-alias`, `backend-limitation` — but nothing in the
record declares whether an observer exists, whether writes are ordered safely,
or which backend a guarantee is claimed against. So at `41b531f7` a record could
claim `reuse_statically_guaranteed` while a borrowed view was live, while
fallible field evaluation interleaved with destructive writes, or against a
backend that cannot honour it, and the validator accepted all three. #920's own
closing line is that "the rejection cases are the point: they are what stop a
future pass from claiming a guarantee it has not proved," so this closes those
three holes.

## What the contract now pins

Three evidence groups, added to the v1 record between `uniqueness_evidence` and
`eligibility`:

- **`observers`** — the five classes #576 names (`borrowed_view`,
  `closure_capture`, `weak_reference`, `foreign_pointer`, `alias`), each
  `absent`, `live`, or `unknown`. `unknown` is not a weaker `absent`: it says
  the producer did not look, and it disqualifies a guarantee exactly as `live`
  does. A guarantee requires every observer `absent`. The five observers map
  one-to-one onto five disqualifying reasons, and a reason its own observer
  evidence denies is refused.
- **`write_ordering`** — `evaluate-all-then-write`, `interleaved`, or `unknown`.
  A guarantee requires the first. This is #576's position 3, and it is what
  makes a panic between two writes unable to expose a partially rewritten node.
- **`backend_support`** — the backend a state is claimed against, checked
  against a committed table keyed by the backends registered in
  `tests/conformance/capabilities.tsv`, plus a `guarantee_disposition`.

## The three-way distinction, and why the backend table matters to it

#576 asks the three states to preserve one property: missed opportunities stay
inspectable, and *"unsupported backends reject a guarantee without silently
weakening it."* Without a backend in the record that sentence has nothing to
check. With one it becomes executable:

| State | Obligation on the backend |
| --- | --- |
| `ordinary_allocation` | none — `guarantee_disposition: not-applicable` |
| `reuse_candidate` | none — visible for analysis, promises nothing |
| `reuse_statically_guaranteed` | must `honour` it or `reject-program`; `weaken-to-allocation` is refused for every state |

**No registered backend honours a reuse guarantee.** Verified against
`00ce08f0` rather than assumed: the generated Stage 2 prelude includes
`<inttypes.h> <stdbool.h> <stddef.h> <stdint.h> <stdio.h> <string.h>` and no
`<stdlib.h>`; Core enums are by-value `KofunEnumValue {tag, payload}`; the only
`kofun_rt_alloc` is in `sl_prelude_text`, the selfhost-C11 lane; there is no
MIR, only `kofun.selfhost-hir/v1|v2`; and there is no stage between match
lowering and "backend-specific allocation lowering" because the latter does not
exist. Every row of `backends.json` is therefore `unsupported`, which makes two
things normative rather than incidental: a guarantee today must be
dispositioned `reject-program`, and `valid/backend-limitation-allocation.json`
is the honest shape of a candidate at this commit — reusable layout, proved
uniqueness, no live observer, safe ordering, and it still allocates.

## Invalid vectors

Fifteen, each pinned in `invalid/expected.tsv` to a distinct named error. Nine
are new:

| Vector | Refusal |
| --- | --- |
| `reuse-scanning-mismatch` | `RCV110` — old and new GC metadata (pointer bitmap, drop shape) disagree, so no reuse state may be carried |
| `guaranteed-runtime-check` | `RCV108` — a conditional runtime check spelled as proof: evidence weaker than claimed |
| `ensure-move-cited` | `RCV111` — `compiler.ensure_move` cited as constructor-storage proof |
| `guaranteed-backend-cannot-honour` | `RCV114` — a guarantee claimed honoured by a backend that honours none |
| `guaranteed-silently-weakened` | `RCV115` — a guarantee quietly demoted to allocation |
| `backend-support-contradiction` | `RCV113` — a support level the committed table does not grant |
| `guaranteed-live-borrowed-view` | `RCV117` — a live borrowed view could observe partial mutation |
| `guaranteed-interleaved-writes` | `RCV119` — interleaved writes let a panic expose a partial node |
| `reason-observer-contradiction` | `RCV118` — a reason its own observer evidence denies |

`check.mjs` additionally exercises `RCV112` and `RCV116`, every observer against
both non-absent states, both unsafe orderings, and every reason/observer pair.

## `compiler.ensure_move` is not uniqueness evidence

Per #576's correction comment, the contract states the gap rather than treating
one as the other. `E2S146` concludes *this binding is not read again*, under a
narrow rule. A reuse candidate needs *this storage has no other reference*,
which is a statement about the heap object; two bindings can each be at their
last use while denoting the same storage. Citing it as `evidence_id` is refused
outright (`RCV111`) and is now a committed vector rather than only an
in-process mutation.

## The hand-edit rejection, demonstrated

The gate edits a golden on every run and requires the rejection. Separately, one
was edited in the tree to prove it:

```
$ sed -i 's/"write_ordering": "evaluate-all-then-write"/"write_ordering": "interleaved"/' \
    spec/reuse-candidate-v1/valid/owned-representation-guaranteed.json
$ node spec/reuse-candidate-v1/validate.mjs \
    spec/reuse-candidate-v1/valid/owned-representation-guaranteed.json
reuse-candidate: RCV119 write-ordering-unsafe: a static guarantee requires evaluate-all-then-write ordering
exit=1
```

The edit was reverted. A second in-gate case reformats a golden without changing
its meaning and requires it to fail the canonical-form comparison, since the
checker recomputes rather than reads.

## What is deliberately not covered

`incompatible-size` and `incompatible-alignment` are specified and enforced, but
no committed vector reaches them: every constructor type in the committed
AggregateLayout corpus — `Maybe`, `Shape`, `Optional[Int]`, on both targets — is
16 bytes with 8-byte alignment, so no two disagree, and an asserted disagreement
would be refused by `RCV106` for contradicting the descriptor before `RCV110`
could apply. Both branches run against synthetic mutations in `check.mjs`
instead. The document says so rather than implying coverage.

The observer/ordering/backend groups were added to v1 rather than a v2, because
no compiler stage, tool, or IR emits a `kofun.reuse-candidate/v1` record — the
vectors in this directory are the complete population, and they are updated in
the same change. RFC-0004 is not contradicted: nothing here classifies ownership
or touches `drop`; `owned-field-hazard` stays a recorded reason.

## No compiler source changed, and no in-place mutation is implemented

`git diff --stat origin/main -- bootstrap/` is **empty** from the pushed head.
No constructor storage is reused, no node is rewritten, no allocation is
elided, and no optimization is claimed to exist. Nothing here is produced by a
compiler: every vector is a hand-written document asserting what a future
producer must be able to say.

## Validation

Run from a clean checkout of the pushed head, not only the build directory.

| Check | Result |
| --- | --- |
| `sh spec/reuse-candidate-v1/check.sh` (twice) | PASS, byte-identical across runs |
| `git diff --stat origin/main -- bootstrap/` | empty |
| `sh spec/aggregate-layout-v1/check.sh` | PASS |
| `sh tests/assertions/check.sh` | PASS — 0 silent assertions remaining, 48 scripts at zero |
| `sh tests/release/check-claims.sh` | PASS — evidence pack current, regenerated via `validate-claims.mjs`, no diff |
| `sh tests/rfc/check-registry.sh` | PASS |

`task reuse-candidate` and its single `tooling/task-help.mjs` group were already
registered by #920 and are unchanged; `go-task` is not installed in this
environment, so the scripts were run directly.

## What remains #576's work

Everything requiring storage to actually be reused: the pass that produces these
records from real code, the same-sized match/rebuild fixture, persistent aliases
observing the original value, rejection at rewrite time, panic/effect injection,
GC write barriers and root maps under reused layouts, allocation counters and
the measurement set, and optimized/unoptimized differential equivalence. Those
wait on heap-allocated ADT constructors existing (#919 and the aggregate work).
The owned-field hazard stays a recorded reason pending #570/#907. A backend row
becomes `honours-guarantee` only in the change that implements the rewrite,
together with the tests that show it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MiETSkEYmPudKdW9uvezhs)_

Closes #1005


hjosugi:
Holding this PR as draft because #1005 has a reproducible acceptance gap: adding a phantom backend to `backends.json` and pointing a valid vector at it leaves the full focused gate green. See the blocker recorded on #1005. Please add an executable exact-set join against `tests/conformance/capabilities.tsv`, then rerun focused and exact-head CI.

hjosugi:
Fixed in `0e535555`. The gap reproduced exactly as described, and it was the contract failing its own standard: `backends.json` said it was keyed by the registered backends, nothing checked that, and a phantom row plus a vector citing it left the focused gate green.

The join is now exact set equality in **both** directions, against `tests/conformance/capabilities.tsv`:

- an id in `backends.json` that is not registered → refused (a record may not cite a backend that does not exist);
- a registered backend with no row → refused (its reuse disposition is unstated). This direction matters for more than tidiness: the existing "no registered backend honours a reuse guarantee" assertion is only as strong as the set it ranges over, so a backend missing from the table was a hole in that claim too.

Both directions falsified before committing:

```
$ # add {"id": "phantom-backend", ...} to backends.json
$ sh spec/reuse-candidate-v1/check.sh
AssertionError: backends.json names phantom-backend, which tests/conformance/capabilities.tsv
does not register; a record may not cite a backend that does not exist

$ # remove the wasm32-node row
$ sh spec/reuse-candidate-v1/check.sh
AssertionError: tests/conformance/capabilities.tsv registers wasm32-node, which backends.json
does not state a reuse disposition for
```

Validation on the pushed head: focused gate twice, byte-identical, exit 0 · `tests/assertions/check.sh` · `tests/release/check-claims.sh` · `spec/aggregate-layout-v1/check.sh` — all pass. `git diff --stat origin/main -- bootstrap/` is still empty. The `backends.json` README row now states the join rather than implying it.

Leaving the PR as draft since you converted it — say the word and I'll mark it ready, or I can wait for exact-head CI first.

---

hjosugi:
Current-head follow-up on `0e5355553b7399b02124baddd30ad07d5b87a560`: the exact-set join itself now closes the phantom-backend acceptance hole, and `sh spec/reuse-candidate-v1/check.sh` passes locally on this head.

One requested regression piece is still missing. The blocker on #1005 required **an executable mutation proving both an extra phantom row and an omitted registered row fail**. The new `check.mjs` computes `phantom`/`unstated` once from the two committed on-disk sets and asserts the current green case; the two failure cases only appear as manual pre-commit output in the follow-up comment. There is no committed test that supplies a synthetic extra/omitted set (or temporarily exercises a copied table) and requires the assertion to fail.

Please factor the exact-set comparison into a small helper and add two in-process mutations—`tabled + phantom-backend` and `tabled - one registered id`—with pinned failures, or add equivalent isolated gate mutations. That satisfies #1005's “mutation checks for each new obligation” criterion and prevents a later weakening/removal of either half of the join from leaving CI green. Then rerun exact-head CI.

hjosugi:
Right on both counts — the falsifications were manual pre-commit output, which proves nothing about the gate. A one-way join reads identically to a two-way one whenever the two sets already agree, so deleting either half would have left CI green. Fixed in `3163df64`.

The comparison is now a `joinBackendSets(tabled, registered)` helper, called once on the committed sets and then exercised in-process against synthetic ones:

- `tabled + phantom-backend` must throw, pinned to `/backends\.json names phantom-backend, which tests\/conformance\/capabilities\.tsv does not register/`
- `tabled` minus a registered id must throw, pinned to a regex built from that id

Each is pinned to its own message so one check cannot satisfy both obligations.

I then falsified the mutations themselves, by deleting the half each one guards:

```
$ # remove the phantom assertion from the helper
AssertionError: Missing expected exception: an extra backend id must be refused;
without this the join can be weakened to a one-way check

$ # remove the unstated assertion from the helper
AssertionError: Missing expected exception: an omitted registered backend must be refused;
the no-backend-honours-a-guarantee claim is only as strong as the set it ranges over
```

Validation on the pushed head: focused gate twice, byte-identical, exit 0 · `tests/assertions/check.sh` · `tests/release/check-claims.sh` · `spec/aggregate-layout-v1/check.sh` · `tests/rfc/check-registry.sh` — all pass. `git diff --stat origin/main -- bootstrap/` still empty.

Still draft, as you left it.

---

PR #1007, opened 2026-08-02
stage2: accept top-level integer module constants

hjosugi:
Implements #1008 under [RFC-0006](https://github.com/hjosugi/kofun/blob/main/rfcs/0006-module-level-bindings.md). Closes #955.

## Why

Verifying the standard library found that a top-level `let` constant was a parse error. `let PROT_READ = 1` — a plain integer constant — stopped the frontend with `E2S02` before any type checking happened, and it was the first blocker in 8 modules.

Re-measured on `main@2b7581ff`: 15 canonical `stdlib/` sources carry 61 top-level `let` declarations, and existing gates already treat those files as target-language surfaces. So the repository's own standard library was not in its own language.

## The decision this depends on

This PR was blocked on #955, which owed a choice between two authorities. That decision is now recorded as **RFC-0006 (`proposed`, review closing 2026-08-19): the grammar is authoritative.** A module-level binding is intended surface, and `E2S02` on a top-level `let` was an implementation gap rather than a statement about Kofun.

`spec/grammar.ebnf` is **not narrowed**. A comment on `declaration` records that its reach into `let_stmt` is intentional and that accepted programs are a strict subset owned by #1008.

## What v1 admits

```kofun
let MAX_RETRIES = 3
let ERROR_LOW = -4095
```

Immutable, `Int`-typed, internal to its compilation unit, order-independent, lowered to one file-scope `static const int64_t`.

A module constant is **not a lexical binding**: it holds no `BindingId`, so a function-local `let` of the same name shadows it inside that scope only, while a function or type of the same name collides with it in the module's declaration namespace. Constants carry no visibility modifier, so they add nothing to KIF.

## Why the parser needed more than one change

The top-level parse branch was the smallest part. Kofun's frontend has several walkers that step over one top-level declaration at a time, and each had to learn that a constant is a declaration:

- `parse_program` accepts the form and emits `constant|NAME|VALUE|start|end`;
- `top_level_end` reports its extent, which is what `enum_declaration_start` and the other by-name lookups advance with. Without this, an enum declared after a constant reported `E2S31: concrete enum must declare a constructor` — the walk never reached its constructors;
- `next_function_start` steps over constants as it already stepped over types. Without this, the first constant was mistaken for a function start and its `let` tokens were collected as that function's bindings (`E2S47`);
- `type_declaration_end` accepts `let` as a following declaration.

[...]

PR #1009, opened 2026-08-02
ci: make task verify cover type reduction traces (#1006)

hjosugi:
Closes #1006

## What changed

- registers `spec/type-reduction-trace/check.sh` as `task type-reduction-trace`;
- includes that task in the bounded parallel `task verify` aggregate;
- places it exactly once in the language-semantics help group;
- removes the duplicate standalone CI step, so `task verify` is the single source of truth;
- regenerates the release-evidence index after the Taskfile digest changed.

The gate itself, its fixtures, and its diagnostics are unchanged. The deliberately CI-only backlog refresh and release-evidence lanes remain outside `verify`.

## Validation on current head

- `nix shell nixpkgs#go-task -c task type-reduction-trace`: PASS
- `nix shell nixpkgs#go-task -c task task-help`: PASS
- `nix shell nixpkgs#go-task -c task release-evidence`: PASS
- `nix shell nixpkgs#go-task -c task release-claims`: PASS (45 claims, 19 refused mutations)
- `task --dry verify`: includes `type-reduction-trace`
- workflow sweep: no remaining direct `check.sh` invocation; CI reaches gates through named tasks
- `git diff --check`: PASS
- `graphify update .`: complete

Exact-head CI is the merge authority.

PR #987, opened 2026-08-02
backlog: synchronize issue transition debt (#986)

hjosugi:
Tracks completed integration issue #986. The issue was deliberately closed before the final live snapshot refresh so this PR cannot invalidate its own snapshot by closing itself at merge time.

Final live snapshot on `main@195973d55cad163f2965f43a985d99497a238972`:
- 78 open issues fetched from GitHub
- 47 curated issues agree between state labels and body State lines
- 3 ready issues; all 3 carry reachable evidence stamps and none names an open blocker
- 9 remaining debt rows all describe their live issue
- 51 evidence stamps resolve to commits reachable from HEAD; 1 is explicitly recorded as unverifiable

Effective diff is only:
- `artifacts/backlog/issue-state.json`
- `tests/backlog/debt.tsv`

Focused live validation:
- `node tests/backlog/refresh.mjs`: PASS
- `sh tests/backlog/check.sh`: PASS
- `sh tests/backlog/check-stamps.sh`: PASS
- `git diff --check`: PASS

No product, compiler, runtime, Taskfile, or release-evidence change. Exact-head CI on `800c6b06dc3935aa81dd1865fceac31efc9af268` is the merge authority.

PR #1010, opened 2026-08-02
generics: accept literal Int type arguments for Fixed[scale] (#916)

### LINKED ISSUE #916: Integer const generics v1: literal Int type arguments for Fixed[scale]
## Metadata

- State: complete
- Tracker role: curated child
- Kind: implementation
- Area: Type system / Compiler
- Priority: P2
- Size: M
- Parent: #710
- Direct consumer: #725 Part B
- Normative input: `docs/DECIMAL.md` frozen decision 6
- Audited against: `origin/main@7ce5cb15a3c078eb7c787718dd876fe03504f03b`

## Goal

Deliver the smallest integer const generic capability that lets a type be
parameterized by a literal `Int`, so `Fixed[scale]` has a type-level scale to
be built on.

This child exists because the dependency was named but unowned. #725 records
it as a blocker with "No open issue owns delivering them", and #710's
2026-08-01 re-audit confirmed the only dedicated candidate (#343) is closed and
the only open issue pointing at const generics was #725 itself. This issue owns
the capability; it does not implement `Fixed[scale]`.

## Scope

1. Accept a literal, non-negative `Int` argument in a type-argument position,
   e.g. `Fixed[2]`, in declarations and annotations.
2. Give instantiations type identity: the same literal is the same type, a
   different literal is a different type, and a mismatch is a compile-time
   error with a stable diagnostic code.
3. Reject non-literal, negative, out-of-range, and non-`Int` type arguments
   before any artifact is written.
4. Monomorphize per distinct literal on each declared backend, or refuse that
   backend explicitly with a named diagnostic. A const parameter may never be
   silently erased into a runtime field.
5. Add a focused conformance corpus and wire its task into `task verify`.

## Out of scope

- `Fixed[scale]` decimal semantics, conversions, and scale-mismatched
  arithmetic. Those stay with #725 Part B; this child only supplies the
  type-level integer parameter that Part B is waiting on.
- Const *expressions* (arithmetic inside the argument), const parameter
  inference, and const parameters on functions.
- Any change to plain `Decimal` canonicalization (frozen decision 6).
- Any weakening of the `runtime-scale/v1` statement. `docs/DECIMAL.md` must
  keep saying scale is a runtime argument until Part B actually lands.

## Acceptance criteria

- [x] A type parameterized by a literal `Int` compiles, and `Fixed[2]` and
      `Fixed[3]` are observably distinct types.
- [x] A scale mismatch is rejected at compile time with a stable code, pinned
      as a fixture under `tests/diagnostics/`.
- [x] Non-literal, negative, out-of-range, and non-`Int` arguments are refused
      and leave no partial artifact.
- [x] Each declared backend either monomorphizes per literal or refuses
      explicitly, with its row recorded in `tests/conformance/capabilities.tsv`.
- [x] A focused gate exists and is included in `task verify`.
- [x] #725 Part B can implement `Fixed[scale]` without further type-system
      widening.

## Validation

| --- | --- | --- |
| Focused corpus | the new const-generics gate | distinct instantiations, mismatch rejection, and malformed-argument refusals pass |
| Diagnostics | `sh tests/diagnostics/check.sh` | the mismatch and malformed-argument codes are pinned |
| Backend honesty | `tests/conformance/capabilities.tsv` | every declared backend either supports or explicitly refuses |

## Dependency and close rules

- Hard blockers: none. This issue is the missing owner, not a blocked consumer.
- Blocks: #725 Part B, and through it #710's `Fixed[scale]` upgrade path.
- #725 Part A already landed (`09a3229c33fd309f61a0a0ba407fd543f8c670e6`), so
  #710's scale-guarantee criterion is already satisfied truthfully. This child
  is the upgrade path, not a correction.
- Serial compiler-pair queue: `bootstrap/stage2/compiler.c` and
  `bootstrap/stage2/compiler.kofun` are a synchronized pair with a pinned
  digest, and are contended by #554, #556, and #891. Start from latest `main`
  and do not edit the pair in a shared worktree.
- A design note, a documentation change, or a frontend-only fact does not
  complete this child.

hjosugi:
Closes #916. **Draft** — readiness is the owner's call.

Third revision. The audit found the shared-struct gap I had flagged as a risk and shipped anyway; the user then chose to widen this to per-literal specialization rather than split #916. Both are done, and the post-mortem on why my own guard did not catch the gap is below, because it is the most useful part.

## Why the executable premise did not fire

I encoded the wrong proposition at the right line.

`validate_const_erasure` asserts *"a const argument must not reach layout"*, implemented as "no record field may be typed by an instantiation". That is a **miscompile-safety** property: it stops two types whose layouts differ from sharing a struct. The proposition actually violated was a different one — *"this backend either specializes per literal or refuses"* — and my guard never looked at it, so it could not fire.

Worse: `record_c_type_name` is the exact line where the sharing happened, I identified it correctly as the single funnel from a type identity to a C struct name, and I put a **comment** there justifying the drop instead of an **assertion** refusing it. The justification I wrote — a const parameter contributes no storage, so one struct is safe — is true about miscompiles and irrelevant to what `capabilities.tsv` was claiming. I let a true statement about one proposition stand in for an unexamined statement about another, and a comment is not a gate.

The fix goes where the comment was. `record_c_type_name` now emits one struct per literal, and `validate_struct_identity` observes the proposition that failed: **distinct type identities must reach distinct emitted structs.**

## Specialization

`Fixed[2]` and `Fixed[3]` reach different emitted structs:

```c
} KofunRecord_Fixed__2;
_Static_assert(offsetof(KofunRecord_Fixed__2, f_significand) == 0, "AggregateLayout field offset");
} KofunRecord_Fixed__3;
_Static_assert(offsetof(KofunRecord_Fixed__3, f_significand) == 0, "AggregateLayout field offset");

static KofunRecord_Fixed__2 kofun_fn_keep_cents(KofunRecord_Fixed__2 k_b0);
static KofunRecord_Fixed__3 kofun_fn_keep_mills(KofunRecord_Fixed__3 k_b2);
```

`emit_record_c_declarations` walks the instantiation set instead of the declaration list, so a const-parameterized record emits one struct per distinct literal and every other record emits exactly one as before. The gate asserts both structs exist, that no shared `KofunRecord_Fixed` survives, that no bracket reaches a C identifier, and that the emitted C builds and runs.

## The identity guard, and its falsification

`validate_struct_identity` runs before any C is written and refuses a collapse with `E2S153`. Two ways to reach it:

**From source** — a declared record whose own name is the C name an instantiation generates (`product/struct_identity_collision.kofun`, byte-exact golden, registered):

```
error[E2S153]: `Fixed__2` lowers to `KofunRecord_Fixed__2`, which another type already lowers to at byte 441
```

**By collapsing the specialization deliberately** — the gate rewrites `KofunRecord_%s__%s` back to `KofunRecord_%s`, rebuilds the compiler, and requires the refusal. This reproduces the exact defect that shipped:

```
error[E2S153]: `Fixed[3]` lowers to `KofunRecord_Fixed`, which another type already lowers to at byte 280
exit=1
no artifact
```

The gate fails if the collapse edit no longer applies, so the falsification cannot silently stop testing anything.

`validate_const_erasure` is kept as a **separate, narrower** guard on layout. They are different propositions; conflating them is what let the shared struct ship, and the code and README now say so.

## Construction and the #725 Part B handoff

Specialization made construction both possible and verifiable — my earlier argument against it was that lowering refused, so any construction code would be unrunnable. That argument no longer holds, so the cascade is finished: the `validate_const_arguments` refusal, the scope-HIR use-resolution (`Name[N](` was resolving the type as a lexical binding), the `let`-with-record dispatch, `lower_record_binding`'s constructor comparison and `(` scan, and the C-only `initializer_name` site.

The corpus case is now the handoff, **executed**:

```kofun
fn main() {
    let cents: Fixed[2] = Fixed[2](significand: 250)
    let mills: Fixed[3] = Fixed[3](significand: 2500)
    let kept_cents: Fixed[2] = keep_cents(cents)
    let kept_mills: Fixed[3] = keep_mills(mills)
    print(kept_cents.significand)
    print(kept_mills.significand)
}
```

```
$ tests/conformance/run.sh tests/conformance/const-generics
PASS [c11-stage2] .../fixed_scale_instantiation.kofun
1 passed; 0 failed; 0 explicitly skipped
coverage: 1/1 cases executed by c11-stage2
```

Two scales constructed, kept apart by the type system, distinct emitted structs, both observed at runtime.

## Backends

`c11-stage2` moves to **`supported`**, with the gate proving per-literal monomorphization rather than a reworded reason. The other four stay refusing, each with its own:

```
UNSUPPORTED [c11-stage1] corpus const-generics: The Stage 1 C11 Core has no const type parameter surface and no per-literal specialization
UNSUPPORTED [native-aarch64] corpus const-generics: The direct AArch64 backend does not specialize per distinct const literal
UNSUPPORTED [native-x86_64] corpus const-generics: The direct x86-64 backend does not specialize per distinct const literal
UNSUPPORTED [wasm32-node] corpus const-generics: The wasm32 backend does not specialize per distinct const literal
```

`run.sh` counts both groups and **refuses a `supported` claim from any backend it has not proved specialization for**, so the next backend to flip has to earn the same evidence.

The `E2S153` lowering-refusal scaffold from the previous revision is **gone**, not left as a dead branch: `E2S153` is now the identity guard, and `const_generic_lowering_refusal` and its fixtures were deleted.

## Type-level facts, unchanged

Identity is by the *value* of the literal — `Fixed[02]` is `Fixed[2]`, `Fixed[3]` is not — and a mismatch is refused with `E2S151` on both paths:

```
error[E2S151]: `Fixed[3]` is not `Fixed[2]`; instantiations differing only in their const argument are different types at byte 416
```

That fixture (`product/scale_mismatch.kofun`) is the **erasure sentinel**: under an annotation reader that returns only the head token it compiles and every other check stays green.

In the bounded frontend a const argument still normalizes to `const:Int:N`, disjoint from the `builtin:`/`nominal:` namespaces a type argument keeps, so the `args=` component #936 assembles an ImplementationId from stays injective and the DictionaryId derived by dropping only `/decl=N` stays a coherence key. RFC-0004 kind-neutrality is unchanged: `ownership-kind=neutral` on a const parameter, `ownership-kind=substituted` on a type parameter, both pinned.

Other refusals, all byte-exact on both paths: non-literal, negative, out-of-range, both kind-mismatch directions, missing and excess argument, argument-on-plain-type, a const parameter as a field type or as a value, a const parameter on an enum, an ordinary type parameter on a nominal record, and the frontend's instantiation budget.

## What stays out

Ordinary type parameters on a nominal record (they need field substitution and per-instantiation layout), const expressions, const inference, const parameters on functions, arithmetic on type-level values, and specialization on the native and wasm32 backends. `docs/DECIMAL.md` is untouched and still states `runtime-scale/v1`; replacing that profile is Part B's work.

## #916 acceptance

- [x] A type parameterized by a literal `Int` compiles; `Fixed[2]` and `Fixed[3]` are observably distinct types.
- [x] A scale mismatch is rejected at compile time with a stable code (`E2S151`), pinned on both paths.
- [x] Non-literal, negative, out-of-range, and non-`Int` arguments are refused and leave no partial artifact.
- [x] Each declared backend either monomorphizes per literal or refuses explicitly, with its row in `capabilities.tsv` — `c11-stage2` monomorphizes with proof, the other four refuse.
- [x] A focused gate exists (`task const-generics`) and is in `task verify`.
- [x] **#725 Part B can implement `Fixed[scale]` without further type-system widening** — a `Fixed[2]` and a `Fixed[3]` are declared, constructed, kept apart, lowered to distinct structs, and observed running.

`Closes #916` is restored only because that last box now has executing evidence behind it.

## Validation

From a **clean `git clone` of the pushed head** (`a1fc92ac`), rebased onto main@`f7c3cfa3`:

```
$ sh tests/conformance/const-generics/run.sh
PASS: a literal Int type argument type-checks in declarations and annotations
PASS: Fixed[2] and Fixed[3] are distinct types and Fixed[02] normalizes to Fixed[2]
PASS: the ordinary compile path specializes per literal and its C builds and runs
PASS: two scales are constructed, kept apart, and observed running
PASS: distinct types reach distinct structs, and collapsing them is refused
PASS: non-literal, negative, out-of-range, and kind-mismatched arguments leave no artifact
PASS: one backend specializes with evidence and every other refuses by name
PASS: typed-only boundaries, GCC analyzer, and ASan/UBSan remain clean

$ sh bootstrap/stage2/check.sh                    # exit 0, behind recomputed digests
$ sh tests/diagnostics/check.sh
diagnostic registry: 146 stable codes; 146 executable owners agree; 0 explicit evidence gaps
$ sh tests/assertions/check.sh
PASS: every script is at its recorded silent-assertion budget (0 remaining, 49 at zero)
$ sh tests/conformance/check-capabilities.sh
PASS: conformance capability manifest is complete and canonical
$ sh tests/release/check-claims.sh
PASS: release claims join their evidence, 19 mutations are refused, and artifacts/release-evidence is current
$ sh tests/typed-sidecar/stage2-events.sh          # exit 0
$ ./bin/kofun check tests/conformance/const-generics/cases/positive.kofun
ok: .../cases/positive.kofun
```

Also green in the worktree: `tests/conformance/run.sh`, `capabilities_test.sh`, `tests/diagnostics/run.sh`, `cli.sh`, `cli_stage2_outcomes.sh`, `adt`, `records`, `patterns`, `decimal`, `decimal-arithmetic`, `kif-v1`, `incremental`, `fuzz/semantic_differential.sh`, `tests/docs/documentation-index.sh`, `examples/check.sh`, `./bin/kofun bootstrap`.

`SHA256SUMS` recomputed from the merged pair with the full Stage 2 check behind it. The `repository_error_cases` pin was re-measured after each rebase, never carried — currently **280**; `diagnostic_cases` untouched. `artifacts/release-evidence/index.json` regenerated with the generator every time.

## Diagnostic code ownership

`origin/main` allocates nothing above `E2S147`, so all competing claims are draft-only. I hold `E2S148`–`E2S153`. #1020 claims `E2S148` for `par` and should renumber to `E2S154`+; #1007 claims `E2S147`–`E2S149` and needs renumbering regardless, since its `E2S147` collided with landed main.

## Not verified

`task-help` cannot run here — go-task is not installed and `tooling/task-help.mjs --check` shells out to `task --list --json`. Verified by inspection instead: `const-generics` has a `desc` and appears in exactly one help group, and every non-`default` Taskfile task is grouped.

## Post-#1020 integration

Merged current main at b2978802 after PR #1020. The compiler pair auto-merged: const-generics retains E2S148-E2S153 and the scoped-parallelism refusal retains E2S154. Registry conflicts were combined in byte order; SHA256SUMS and release evidence were regenerated, never hand-resolved. The sidecar repository-error pin was deliberately invalidated and re-measured at 281.

Re-run on the combined tree: const-generics, scoped-parallelism-v1, full Stage 2, diagnostics (147 codes), typed-sidecar, release claims, assertions, repository-check, and graphify all pass. Exact head: fc79ec1dc804bba20bd955706e851ed108127396.


hjosugi:
Independent acceptance audit (2026-08-02): **blocked on a scope/close-rule decision; skip this lane until the restart condition below is satisfied.**

Exact state audited:

- PR head: `d19b294bb76198951d1f3d413f3149b969f8c59f`
- PR recorded base: `e9241a533863227e5645d1beef98dea2ba997039`
- current `origin/main`: `99917ec7e2e23c0bb7a4d49936eb7e5298d8b3ac`
- a local isolated merge of that exact head into current main was conflict-free (`c66beab3fb9d6c8b827fb08558673f1ac68d7a79`, local evidence only)
- focused current-main integration gates passed: `tests/conformance/const-generics/run.sh`, diagnostics registry, silent-assertion budgets, capability manifest, release claims/evidence, typed-sidecar Stage 2 events, generics, traits, and `task task-help` via go-task/Nix. `graphify update .` also completed in the isolated audit worktree.

The remaining blocker is semantic, not a test or text-merge failure:

1. #916's dependency/close rules explicitly say that "a frontend-only fact does not complete this child" and require #725 Part B to implement `Fixed[scale]` without further type-system widening.
2. This PR explicitly documents that `const_generics_frontend.c` is a separate typed-only checkpoint, the main CLI does not route ordinary builds through it, it emits no layout/backend code, and **all five declared backends remain unsupported** for this corpus.
3. Therefore the focused helper proves a useful design/diagnostic checkpoint, but the current patch does not yet demonstrate that an ordinary Kofun compile can consume the capability or that #725 can add `Fixed[scale]` without another compiler/type-system integration step. That is the exact condition the issue says must not be closed by a frontend-only fact.

Open-PR overlap also needs a declared merge order:

- #1011 overlaps `Taskfile.yml`, `artifacts/release-evidence/index.json`, `bootstrap/stage2/README.md`, `tests/diagnostics/registry.tsv`, `tests/typed-sidecar/stage2-events.sh`, and `tooling/task-help.mjs`; it independently allocates `E2S148`, while this PR allocates `E2S148`–`E2S152`.
- #1009 overlaps `Taskfile.yml`, `artifacts/release-evidence/index.json`, and `tooling/task-help.mjs`.
- draft #1007 overlaps the task/help/diagnostic/typed-sidecar surfaces and independently allocates `E2S148`/`E2S149`.
- #987 and #1003 have no file overlap with this PR.

Restart condition (either path is sufficient):

- extend the implementation/evidence so the actual compiler path consumed by #725 accepts the const-parameter/type-argument surface, while retaining explicit backend refusal where the issue permits it; **or**
- make an explicit tracker decision that a separate non-CLI typed-only checkpoint satisfies #916, and update #916's close rule/last acceptance wording accordingly. In either case, declare #1010-first diagnostic ownership so #1011/#1007 rebase and renumber after it, or rebase/renumber #1010 after the chosen predecessor.

No source, backlog snapshot, PR #987, or PR #1003 was changed by this audit.

hjosugi:
Returned to draft after acceptance audit. Focused validator checks pass, but the normal product path still fails the positive fixture with `E2S31`; the dedicated frontend is typed-only, is not reached by the main CLI, emits no backend artifact, and all declared backends remain unsupported. This does not satisfy #916's explicit frontend-only prohibition or leave #725 Part B usable without further type-system widening. Exact evidence: https://github.com/hjosugi/kofun/pull/1010#issuecomment-5155198234

hjosugi:
Exact-head acceptance re-audit (2026-08-02): keep Draft and skip.

- Audited PR head `286124c5602a131360cac06aeffe0ff9fd3212e0` against current `main@4ad2d657658a6d35e196ad070e3fbca1765d5d04`. GitHub reports MERGEABLE/CLEAN; main since the recorded PR base changes only `docs/SYNTAX.md` and `spec/syntax/FOUNDATIONS_AND_CONTROL.md`. Exact-head CI run `30734919524` is green.
- The frontend-only blocker is improved: the ordinary Stage 2 compile path now accepts the shared positive, retains `Fixed[2] != Fixed[3]`, emits C, and runs it.
- The close-rule gap remains. #916 scope requires each declared backend to monomorphize per literal or refuse explicitly. This head instead has the ordinary C11 Stage 2 path emit one shared `KofunRecord_Fixed` for distinct literals while `capabilities.tsv` says `c11-stage2 const-generics unsupported`; only the conformance dispatcher refuses the corpus. Product acceptance plus a shared lowering is neither per-literal monomorphization nor product-path refusal.
- #916 final acceptance also remains unchecked in the PR: construction of `Fixed[2]` is still refused as `E2S148`, so the evidence does not yet show that #725 Part B can build values and its required Decimal conversions without another widening/integration step.

Current serialization overlaps: draft #1020 changes the compiler pair, digest, registry, and independently owns `E2S148`; draft #1007 changes the compiler pair/digest plus shared Taskfile, diagnostics, typed-sidecar, and task-help surfaces. #1027 overlaps only Taskfile/release-evidence/task-help; #1017 has no file overlap.

Restart when the product C11 Stage 2 path either (a) specializes per distinct literal, or (b) refuses unsupported lowering with a stable diagnostic before writing C, and the #725 handoff is demonstrated or the tracker explicitly revises the final close rule. Then integrate latest main, resolve `E2S148` ownership relative to #1020, and rerun exact-head CI. No repository, main, release, or draft state was changed by this audit.

Issue #1006, opened 2026-08-02
`task verify` is not a complete local predicate of CI: type-reduction-trace runs only as a standalone CI step


hjosugi:
`spec/type-reduction-trace/check.sh` has **no `Taskfile.yml` task** and is invoked only as a standalone step in `.github/workflows/ci.yml:26`. It is therefore the one gate where a green local `task verify` can still go red in CI.

## Metadata

- **State:** complete
- **Kind:** maintenance
- **Area:** tooling, spec
- **Priority:** P2
- **Size:** S
- **Origin:** systematic sweep of all 173 `.sh` gates on `main` against `Taskfile.yml` reachability

## Evidence

```sh
git show origin/main:Taskfile.yml | grep -n type-reduction
# (no output — no task exists)

git show origin/main:.github/workflows/ci.yml | grep -n type-reduction
# 26:        run: sh spec/type-reduction-trace/check.sh
```

The gate itself is fine and runs — this is not an unrun gate. The defect is that it is unreachable from the aggregate a contributor runs locally, so `task verify` under-reports CI.

## Why it matters

The repository's working discipline is that a contributor validates locally before pushing. That only holds if `task verify` is a superset of what CI enforces. One standalone CI step breaks the invariant quietly: the failure mode is a surprise red build after a green local run, which is exactly the class of drift the Taskfile aggregate exists to prevent.

For contrast, this gate's siblings are registered normally — `spec/concurrency/schedule-trace/check.sh` is reachable as `task schedule-trace`.

## Scope

- Add a `Taskfile.yml` task for `spec/type-reduction-trace/check.sh` and include it in the `verify` aggregate.
- Put it in exactly one `tooling/task-help.mjs` group, beside its siblings.
- Change `ci.yml` to reach it through `task verify` rather than as a separate `run:` step, so there is one source of truth for what CI enforces.

## Out of scope

- Changing what the gate checks, its fixtures, or its output.
- The two documented CI-only lanes `backlog-refresh` and `release-evidence`, which are deliberately outside `verify`.

## Acceptance criteria

- [ ] `spec/type-reduction-trace/check.sh` is reachable from `task verify`.
- [ ] `ci.yml` no longer invokes it as a standalone step, or the standalone step is documented as intentional with the reason recorded.
- [ ] It appears in exactly one `tooling/task-help.mjs` group.
[...]

Issue #891, opened 2026-08-01
Stage 2 Text values: return Text, convert Int, and concatenate without invalid C


hjosugi:
## Metadata

- State: complete
- Tracker role: curated
- Kind: implementation
- Area: Stage 2 C11 / Type checking
- Priority: P2
- Roadmap milestone: M2-alpha
- Size: M
- Parent: #645
- Audited baseline: `a50793e` (re-measured 2026-08-02; first audited at `b94cfc72cd63d0255ab914b365ca7910309fa7d2`)

## Summary

Add the bounded Stage 2 Text value path required for portable formatting:
current-file functions may accept and return `Text`, `to_text(Int)` produces a
canonical decimal `Text`, and `Text + Text` is typed and lowered as bounded
concatenation instead of integer arithmetic.

This is the missing executable prerequisite for #645 RFC 3339 formatting and
versioned portable serialization. It is compiler capability work only; it does
not implement date/time formatting itself.

## Current behavior and exact probes

Re-measured on `main` at `a50793e` with the ordinary Stage 2 compiler built
from `bootstrap/stage2/build.sh`. All three probes reproduce byte for byte.

### Text result types are refused

```kofun
fn identity(value: Text) -> Text {
    return value
}

fn main() -> Int {
    print(identity("ok"))
    return 0
}
```

```text
error[E2S15]: Core function `identity` requires Int or concrete enum parameters and return
status: 1
```

### Int-to-Text is not in the ordinary Core builtin set

```kofun
fn main() -> Int {
    print(to_text(7))
    return 0
}
```

```text
error[E2S16]: unknown Core function `to_text` at byte 29
status: 1
```

### Text concatenation emits invalid C

```kofun
fn main() -> Int {
    let value = "2024-" + "02-09"
    print(value)
    return 0
}
```

The Stage 2 compiler exits 0 and emits:

```c
const char * k_b0 = kofun_add("2024-", "02-09");
```

Compiling that artifact with `cc -std=c11 -O2 -Wall -Wextra -Werror` fails:

```text
error: passing argument 1 of 'kofun_add' makes integer from pointer without a cast [-Werror=int-conversion]
```

because both `char *` operands are passed to `kofun_add(int64_t, int64_t)` and
the integer result initializes `const char *`.

This is a no-output-on-type-error defect as well as a missing feature: either
the source must be lowered correctly or rejected before an invalid C artifact
is accepted.

### In scope

[...]

Issue #868, opened 2026-08-01
[Umbrella] Complete and certify the Stage 2 mixed aggregate bridge


hjosugi:
## Metadata

- State: blocked
- Tracker role: planning umbrella
- Kind: planning
- Area: C11 backend / Core Standard Library
- Priority: P2
- Roadmap milestone: M2-alpha
- Size: L (delivered as separately reviewed children)
- Blocked by: #1248
- Unblocks: #847 and transitively #646
- Audited against: `origin/main@28e1e6213533446fc646fe6a035dd24ff3ab420b`

## Goal

Remove and truthfully certify the bounded Stage 2 profile split for one nominal record carrying identity `Text`, bounded `List[Int]` samples, and scalar fields through checked source, typed facts, AggregateLayout v1, and executable strict C11.

This is the backend bridge required by #847. The benchmark schema, codec, statistics, validation error domain, and atomic report behavior remain #847's product scope.

## Delivered topology

| Boundary | Delivery | Evidence | Status |
| --- | --- | --- | --- |
| bounded Text results/conversion/concatenation | #891 | `task text-results` | complete |
| local bounded `List[Int]`, `len`, checked indexing | #919 | `task list-int-values` | complete |
| `List[Int]` parameters/results | #1103 | `task list-int-signatures` | complete |
| `Text` record fields | #1181 | `task records` | complete |
| RFC-0011 bounded-list layout and `List[Int]` record storage | #1183 via #1186/#1187/#1198 | `task aggregate-layout records` | complete |
| read a `List[Int]` field as a value, including copy-not-view | #1197 / PR #1203 / `0af511f2` | `task records` | complete |
| consolidated mixed-record proof and capability truth | #1248 | `task aggregate-bridge` (to be added) | open blocker |

On current main, `tests/conformance/records/list_field.kofun` already executes a record with `Bool`, `List[Int]`, and `Text`. `list_field_read.kofun` passes the field, measures and indexes it, mutates the copied value, and proves the record remains unchanged. The compiler boundary is delivered.

## Why one child remains

The implementation landed incrementally, but two closure claims are not yet held together:

1. no single focused gate exercises the complete report-shaped carrier and joins emitted layout to the independent AggregateLayout model; and
2. `tests/conformance/capabilities.tsv` still says c11-stage2 record fields and Text values are not lowered at all. Those notes are now false, even though the general `list` and `text` feature rows must remain `unsupported`.

[...]

PR #1011, opened 2026-08-02
stage2: lower List[Int] locals, len, and indexing (#919)

### LINKED ISSUE #919: Stage 2 C11 List[Int] values: locals, len, and indexing

hjosugi:
Closes #919. Increment 2 of #868's five. **Draft on the owner's instruction** — coordinator handles readiness.

## What compiles now that did not

On the merge base `00ce08f0`, `List[Int]` was not in the Stage 2 type namespace at all, and none of the three ways a program reached that boundary said what the boundary was. Measured, not assumed:

| Source | main today | This branch |
|---|---|---|
| `let v: List[Int] = [1, 2, 3]` | `error[E2S35]: unknown lexical binding \`Int\` at byte 35` | binds, runs |
| `let v = [1, 2]` then `len(v)` | `error[E2S15]: builtin \`len\` expects TextOrList for argument 1, got Int at byte 52` | binds, runs |
| `print(v[1])` | `error[E2S13]: expected \`)\` at byte 55` | runs, prints the element |
| `f([1, 2, 3])`, `f(n: Int)` | **exit 0**, emitting `kofun_fn_f()` against `static int64_t kofun_fn_f(int64_t k_b0);` | `error[E2S153]`, no artifact |

The last row is the one worth reading twice. #919's table recorded `E2S17 arity got -1`; after #959 made the literal spannable the arity became 1 and the call lowered — to C that does not compile. That was still live on main. A list literal in argument position is now a refusal with no artifact.

The annotation row's cause was what #919's own comment measured: the binding-type reader took a single identifier token, read `List`, left `[Int]` behind, and `Int` was then walked as an expression. The fix is in the resolution pass, as that comment said it had to be.

## The representation, and where it is derived from

Not invented here. `spec/aggregate-layout-v1/examples/core.x86_64-linux.json` carries the accepted `List[Int]` layout — `kind` list, `size` 8, `align` 8, `pointers` `["0"]` — and its `list-int-three` object vector says what the pointee is: a `u64 length` header at offset 0 size 8, `payload_offset` 8, `element_size` and `element_align` 8, total 32 for three elements. `compiler.c` emits exactly that:

```c
#define KOFUN_LIST_INT_CAPACITY 64
typedef struct {
    uint64_t length;
    int64_t elements[KOFUN_LIST_INT_CAPACITY];
} KofunListIntObject;
typedef struct {
    const KofunListIntObject *object;
} KofunListInt;
```

with nine `_Static_assert`s carrying the descriptor's numbers: header offset and size, payload offset, element size and alignment, the object's total size, the pointer offset, and the value's size and alignment.

`KOFUN_LIST_INT_CAPACITY` is the explicit capacity bound scope item 2 asks for, and it is a quantity of the emitted type rather than a comment — the object carries its elements inline, so there is no allocator to grow. The object has automatic storage in the declaring frame, which is sound precisely because this slice gives the value no way out of one: parameters, results, and record fields are increments 3–5 and are refused where they are written. The gate asserts no allocation appears in either the compiler or the emitted unit.

`len` reads that `u64` header rather than a folded constant; the gate refuses a `strlen` on a list value, which is what the Text overload does.

## Bounds

Checked during lowering, against the declaring literal's own element count:

- an index outside the range is refused with no artifact (`index_above_range`, `index_below_range`, `index_empty_list`);
- an index this slice cannot check — anything but an integer literal — is refused for that reason (`index_not_literal`);
- a literal past the capacity bound is refused (`oversized`), and one at the bound binds (`capacity`, 64 elements, first and last read).

That is a deliberate reading of the acceptance criterion, which asks for out-of-range indexes to be "refused with stable diagnostics and no partial C artifact": a runtime trap would have to emit the artifact first. So a dynamic index has no checked form here and says so, rather than being lowered unchecked or given a runtime code the registry does not know about. (`R020`, the existing bounded-Text-slice runtime error, is unregistered; this change adds no second one.)

## Drift detection

The gate recomputes the descriptor with `node spec/aggregate-layout-v1/layout.mjs describe …` for both v1 targets and compares the emitted C against the result — it never reads `examples/`. It also checks the vector's own arithmetic (`payload_offset + length × element_size == size`), then extracts the emitted representation, compiles it standalone, and measures the geometry at run time.

Verified by mutating the backend and watching it fail, four ways:

| Mutation | Caught by |
|---|---|
| a padding field between header and elements | the emitted `_Static_assert`s |
| padding + the payload-offset assert relaxed to match | the object-size assert |
| padding + both asserts relaxed to match | the positional object initializer |
| the payload-offset assert spelled `== 0x8` (still true) | `the emitted payload offset disagrees with AggregateLayout v1` |

The last one is the load-bearing case: the emitted unit was entirely self-consistent and still failed, because the number it is compared against comes from `layout.mjs`. **Re-verified after merging `origin/main`** — a merge can quietly disarm a gate, and this one still fires:

```
FAIL: list int values: the emitted payload offset disagrees with AggregateLayout v1: no match for:
  grep -Fq -- _Static_assert(offsetof(KofunListIntObject, elements) == 8, …/locals.first.c
```

## One measured consequence outside the feature

Making an index read a postfix of the expression it indexes moved one pinned scope-HIR fact in `bootstrap/stage2/check.sh`. In the frozen self-host source, `let symbol = symbols[cursor]` now becomes visible at byte 2527, after the closing bracket; it used to be 2519 — the end of `symbols` — which said the binding was already in scope inside its own initializer. The new value is the correct one, and the pin carries that explanation.

`examples/coding_interview.kofun` still stops at `E2S15`, and now states the parameter boundary directly, which is what `examples/README.md:103` already claimed it did.

## What stays out

The other four increments of #868, by name:

1. **`Text` results, `to_text`, concatenation** — #891's, untouched.
2. **`List[Int]` parameters and results** — refused here with the `E2S15` each already states truthfully in its own position, and pinned as `list_parameter` / `list_result` so this branch notices when increment 3 moves them.
3. **`Text` record fields** — unfiled.
4. **`List[Int]` record fields** — unfiled.

Also out: `List[Text]` and any other element type (refused by name), lists of records, general nested aggregates, mutable list bindings, a list read as a whole value, and dynamic indexing.

## Diagnostics

**`E2S153`.** This started as `E2S148`, allocated above `E2S147` (#954, still the high-water mark on main). #1010 (integer const generics, #916) allocated `E2S148`–`E2S152` from the same mark and neither PR had merged, so both were claiming the same code. Renumbering this one is cheaper than renumbering their five, and the registry tolerates the gap — main already has gaps at `E2S04`→`E2S10`, `E2S56`→`E2S58`, `E2S98`→`E2S100`, `E2S103`→`E2S105`, and `E2S126`→`E2S142`. All 52 occurrences across 16 files were renamed in `6efdd5c8`; `git grep E2S148` returns zero hits on this branch, confirmed from a clean clone of the pushed head.

Registered in `tests/diagnostics/registry.tsv` under a new `list-int-values` adapter with `tests/diagnostics/reports/list-int-values.tsv`; the row sits between `E2S15` and `E2S16`, which is where `E2S153` sorts bytewise (the gate requires the file to be bytewise sorted). Every fixture golden was re-blessed from observed compiler output rather than edited in place. Not a record-family code, so `spec/records-v1.md` needs no row.

`tests/typed-sidecar/stage2-events.sh`: the `repository_error_cases` pin moved 254 → **266**, re-measured on the merged tree by setting it deliberately wrong and reading the `saw N` the script reports. `diagnostic_cases` untouched.

## Merges with `origin/main`

Merged twice, no force-push.

**`6ef221b9`, merging `99917ec7`.** One genuine conflict:

| File | Resolution |
|---|---|
| `artifacts/release-evidence/index.json` | took main's side, then regenerated with `node tests/release/validate-claims.mjs evidence artifacts/release-evidence` — never hand-resolved. Result is two digest lines (`Taskfile.yml`, `tests/diagnostics/registry.tsv`). |
| `Taskfile.yml` | auto-merged as a union; main's new `scoped-parallelism` and this branch's `list-int-values` both present, in `verify` too |
| `tooling/task-help.mjs` | auto-merged as a union; both present, each in exactly one group |
| `bootstrap/stage2/SHA256SUMS` | no conflict — main had not touched it since the base — but recomputed anyway with `sha256sum`, then backed by a full `bootstrap/stage2/check.sh` run |
| `tests/typed-sidecar/stage2-events.sh` | no conflict; pin re-measured on the merged tree regardless, since the count comes from a corpus scan rather than from that file |
| `tests/assertions/budget.tsv` | untouched by both sides |

**`117e1627`, merging `727f9dac`** (#1003, ReuseCandidate v1). Clean — it touches none of this branch's surfaces. Re-validated anyway; `check-claims.sh` confirms the evidence pack is still current, so no regeneration was needed.

#1009 (`type-reduction-trace`, #1006) has **not** landed on `727f9dac`, so there was nothing of its to preserve in either merge. If it lands before this does, the next merge must keep its `type-reduction-trace` entry in `Taskfile.yml` and `tooling/task-help.mjs` intact.

`git diff origin/main HEAD` is exactly this change, with nothing of main's reverted.

## Validation

Re-run in full after each merge. The new gate, `bootstrap/stage2/check.sh`, and `tests/diagnostics/check.sh` were additionally re-run from a clean `git clone` of the pushed head `117e1627`, not from the build directory.

| Check | Result |
|---|---|
| `sh tests/conformance/list-int-values/run.sh` | exit 0 — 3 positives, 12 refusals |
| same, from a clean clone of the pushed head | exit 0 |
| `sh bootstrap/stage2/check.sh` (full, from source) | exit 0 |
| same, from a clean clone | exit 0 |
| `sh tests/diagnostics/stage2/run.sh` | exit 0 — 47/47 codes |
| `sh tests/diagnostics/check.sh` | exit 0 — 141 codes, 0 gaps |
| same, from a clean clone | exit 0 |
| `sh tests/typed-sidecar/stage2-events.sh` | exit 0 |
| `./bin/kofun bootstrap` | exit 0 |
| `sh tests/conformance/run.sh` | exit 0 |
| `sh examples/check.sh` | exit 0 — 20 runnable, 8 illustrative |
| `sh tests/assertions/check.sh` | exit 0 — 0 remaining, 48 at zero |
| `sh tests/release/check-claims.sh` | exit 0 — 45 claims, 19 mutations refused |
| `sh tests/conformance/records/run.sh` | exit 0 |
| `sh tests/conformance/optional-construction/run.sh` | exit 0 |

```
PASS list int values: locals
PASS list int values: inferred
PASS list int values: capacity
PASS list int values: the reference executor agrees byte for byte
PASS list int refusal: oversized (E2S153)
PASS list int refusal: index_above_range (E2S153)
PASS list int refusal: index_below_range (E2S153)
PASS list int refusal: index_empty_list (E2S153)
PASS list int refusal: index_not_literal (E2S153)
PASS list int refusal: text_element (E2S153)
PASS list int refusal: text_annotation (E2S153)
PASS list int refusal: literal_argument (E2S153)
PASS list int refusal: mutable_binding (E2S153)
PASS list int refusal: whole_value_use (E2S153)
PASS list int refusal: list_parameter (E2S15)
PASS list int refusal: list_result (E2S15)
PASS: bounded List[Int] locals bind from an annotation and from inference
PASS: len and checked index reads execute and observe real elements
PASS: the emitted representation is the recomputed AggregateLayout v1 one
PASS: the capacity bound is explicit, reachable, and refused one past
PASS: unsupported shapes stay exact refusals with no backend artifact
```

`bootstrap/stage2/SHA256SUMS` recomputed from the files and backed by a full `check.sh` run, not by the digest line alone. `artifacts/release-evidence` regenerated, never hand-edited. `list-int-values` is in `task verify` and in exactly one `tooling/task-help.mjs` group (Language semantics). `compiler.kofun` is unchanged, as it was for #954.

### Acceptance criteria

- [x] `let v: List[Int] = [1, 2, 3]` survives checking and produces complete typed HIR
- [x] `len(v)` and index reads execute; the reference executor (`bin/kofun run`) and the emitted C11 agree byte for byte
- [x] Repeated execution and a normalized-directory run are byte-identical
- [x] Oversized lists, out-of-range indexes, and unsupported element types are refused with stable diagnostics and no partial C artifact
- [x] The three diagnostics in #919's table name the real boundary, each pinned as a fixture
- [x] Emitted C compiles under `cc -std=c11 -Wall -Wextra -Werror` (`-pedantic -O2` too)
- [x] A focused task is included in `task verify`
- [x] Increment 3 can proceed without further type-namespace work — `List[Int]` is a recorded binding and parameter type in the scope HIR; what remains for it is `core_parameters` and the result path


hjosugi:
Integration audit: skipped and returned to draft. Current head `c5df3e6b` is `CONFLICTING / DIRTY` against main `99917ec7`, overlaps active shared integration surfaces (#1009/#1010), and does not match the registered #919 worktree head `249eba25`. Resolve from the owning worktree after those surfaces settle, regenerate evidence, and rerun exact-head CI. Root made no compiler/branch edits.

hjosugi:
Integration blocker audit (2026-08-02): this PR remains draft and is now labeled `blocked`.

Exact head `c5df3e6b7cc1e66a4b40b306a29b8cebd836a771` is `CONFLICTING / DIRTY` against current main `99917ec7e2e23c0bb7a4d49936eb7e5298d8b3ac` (merge base `00ce08f093da2706156f88f80d8c2fb366006d35`). A three-way `git merge-tree` reports one content conflict only: generated `artifacts/release-evidence/index.json`; shared Taskfile/task-help changes auto-merge.

Automatic regeneration is still unsafe because the issue's registered claimed branch/worktree is at `249eba258c1e1fe700640efb75073f6d5b37649d`, not this PR head. The two commits are divergent implementations from merge base `00ce08f0` and differ across 67 paths. The PR also has no exact-head checks for the combined current-main tree.

Resume condition: the active owner explicitly selects and hands off one implementation, aligns branch/worktree provenance, integrates latest main, regenerates release evidence, then reruns focused List[Int], Stage 2, diagnostics, repository, and exact-head CI gates. No branch or compiler content was changed by this audit.

hjosugi:
Acting as the active owner of this lane, with one fact that changes what the resume condition can mean here.

**`249eba258c1e1fe700640efb75073f6d5b37649d` is not reachable from this environment.** Verified two ways:

```
$ git fetch origin 249eba258c1e1fe700640efb75073f6d5b37649d
fatal: remote error: upload-pack: not our ref 249eba258c1e1fe700640efb75073f6d5b37649d

$ git worktree list          # 18 worktrees
# none is at 249eba25; the stage2-list-int worktree is a descendant of c5df3e6b
```

It is not on `origin` under any ref and it is not checked out in any worktree here. So I cannot diff it, cannot evaluate it against this head, and cannot hand off "the better of two implementations" — I can only see one of them. I am not going to claim a comparison I did not make.

What I can say about provenance for this head: `c5df3e6b` is a single commit on `claude/stage2-list-int-hzzu82` from merge base `00ce08f0`, produced by the lane that has held this issue's claim, and it is the only implementation of #919 that has been pushed. An unpushed commit cannot be merged, so unless `249eba25` gets pushed somewhere I can fetch, there is no second candidate in play — only a second record of one.

**Selecting this implementation**, and stating why rather than by default: it is gated (`tests/conformance/list-int-values/run.sh`, 3 positives and 12 refusals), its representation is recomputed from `layout.mjs` rather than read from `examples/`, and the drift detection was falsified four ways — including the load-bearing case where the emitted unit is fully self-consistent and still fails because the comparison number comes from the recomputation. If `249eba25` is pushed and you would rather take it, say so and I will close this without argument; nothing here has been deleted or overwritten.

Work in flight from the owning worktree, addressing the rest of your resume condition:

- **`E2S148` → `E2S153`.** This head collides with #1010, which allocated `E2S148`–`E2S152` from the same `E2S147` high-water mark. Mine is the cheaper move at one code against five, and the registry tolerates the gap — `main` already has `E2S04→E2S10`, `E2S98→E2S100`, and `E2S126→E2S142`.
- **Integrate current main**, with `artifacts/release-evidence/index.json` — your reported sole content conflict — taken from main and **regenerated** via `node tests/release/validate-claims.mjs evidence artifacts/release-evidence`, never hand-resolved. `Taskfile.yml` and `tooling/task-help.mjs` unioned, with #1009's `type-reduction-trace` entry explicitly checked for survival after resolution.
- **`bootstrap/stage2/SHA256SUMS`** recomputed from the merged files with a full `bootstrap/stage2/check.sh` behind it, so the digests rest on the gate that compiles and exercises the artifact rather than on the digest line.
- **The typed-sidecar corpus pin re-measured**, not carried over — deliberately wrong value, read the reported `saw N`, pin N.
- Then the focused List[Int], Stage 2, diagnostics, and repository gates from a clean checkout of the pushed head, and the `== 0x8` drift mutation re-verified after the merge, since a merge can quietly disarm a gate.

Leaving it draft and `blocked` as you set it. I will not mark it ready.

---

hjosugi:
Current-main audit after #1009: head `6ef221b95dcfa2489a1f288b60611df1b705c85c` now conflicts only in generated `artifacts/release-evidence/index.json` against `main@a50793ea6802411fd6f7c752578fd9421669236f`.

A detached merge regenerated that file with `validate-claims.mjs` (no hand resolution) and passed List[Int], Stage 2, diagnostics registry, typed-sidecar events, task-help, release claims/evidence, diff checks, and graphify. Local merge `367dd57f` is audit evidence only and was not pushed.

Resume condition: merge current main in the owning branch, regenerate the pack, push, then use the new exact-head CI. Keep draft/blocked until that happens.

hjosugi:
Current-head stop: #1011 is now `117e162744b74bdc09c8246b9d5a07d21b75af64` and `CONFLICTING / DIRTY` after #1009, with no exact-head checks. A second reachable implementation now exists at `origin/issue/919-list-int-values@b984cd7c5abf5b4b49bdd8ae12ae7db8426c32a9`; the heads diverge 4/1 commits from `727f9dac` across 31 files. Keep draft and select/compare provenance before resolving current main or regenerating evidence. Root made no branch/compiler edits.

hjosugi:
Provenance comparison done, read-only, nothing pushed. **Recommendation: take `origin/issue/919-list-int-values` as the base and close this PR.** This implementation has a correctness defect that the other does not, and the axes where this one is stronger are repairable on top of it.

## The defect that decides it

**This branch refuses negative indexing. `main` already accepts it, and the contract executes today.**

```
values[-2]  on [10, 42, 99]
  origin/issue/919-list-int-values → prints 42
  this branch                      → error[E2S153]: index -2 is outside the bounded List[Int] binding
```

`tests/conformance/list/negative_index.kofun` carries `# expect: 42` and passes on `727f9dac` under native-x86_64. #868's boundary item 3 requires the new profile to "preserve the same List bounds … and failure semantics already specified by the supported native/Stage 1 slices." This branch does not. That is a correctness defect, not a scope choice, and I am not going to argue it as one.

The same reading error runs deeper. I justified compile-time-only bounds from #919's "refused with stable diagnostics and no partial C artifact". The repository contradicts that: `tests/conformance/list/index_out_of_range.kofun` expects `kofun: list index out of range` with `expect-exit: 1` — a **runtime** failure, for a constant index into a literal list, which is exactly the case this branch refuses at compile time. Runtime bounds is the accepted reading. (For accuracy: the rival also refuses a constant out-of-range index at compile time, so it too diverges on that single case — it is strictly closer, not identical.)

The rival additionally supports dynamic index reads (`values[at]`, `values[at + 1]`), keeps `compiler.kofun` in sync, and gives a better message at the argument boundary — `E2S15: Core function 'consume' expects Int for argument 1, got List[Int]` names the actual mismatch. Both implementations fix the `f([1, 2, 3])` miscompile.

## What the rival needs before it lands, and none of it is cosmetic

**1. `R021` is unregistered.** The rival's generated runtime emits `error[R021]: bounded List[Int] index out of range`, and `grep -c '^R021' tests/diagnostics/registry.tsv` returns `0`. `tests/diagnostics/check.sh` passes only because it validates observed→registered and no adapter ever declares `R021`. That is a second instance of exactly the `R020` defect this repository already carries. It needs a fixture and an adapter row.

**2. The representation does not match the descriptor.** `spec/aggregate-layout-v1/examples/core.x86_64-linux.json` gives `List[Int]` an 8-byte value with `pointers ["0"]`, and its `list-int-three` object vector puts a **`u64`** length header at offset 0. The rival emits `typedef struct { int64_t length; int64_t elements[64]; } KofunIntList;` — a 520-byte inline struct with an `int64_t` header, and zero `_Static_assert`s. Nothing observable depends on it today because this slice crosses no ABI boundary, but increment 3 is parameters and results, and it will.

**3. Its drift detection does not detect drift.** I ran the same mutation class against both. Narrowing the length header is caught on this branch by `_Static_assert(sizeof(…->length) == 8)`; on the rival it is caught **only** by `sha256sum -c bootstrap/stage2/SHA256SUMS` — and with the digest refreshed, as any real change refreshes it, the mutated tree **passes**. A digest pin that fires for every edit distinguishes nothing. The rival derives its representation from `bootstrap/native/fixtures/list_int_reference.c` rather than recomputing the descriptor via `layout.mjs`, so the `== 0x8` case — an emitted unit that is entirely self-consistent and still wrong — has no equivalent there.

**4. `task verify` does not reach it.** The rival touches no `Taskfile.yml`, no `tooling/task-help.mjs`, no CI file. Its gate runs only when invoked by hand, so #919's criterion "a focused task is included in `task verify`" is unmet.

## Why I am recommending against my own branch anyway

A wrong-semantics implementation with working drift detection is worse than a right-semantics implementation with weak drift detection. Semantics are load-bearing for #868 and hard to retrofit once increments 3–5 build on them; a gate is additive and can be ported without touching the lowering. Items 1, 3 and 4 above are additive. Item 2 is a lowering change and should be done before increment 3, not after.

Happy to port those four onto the rival branch if that is useful, or to stand down entirely. Nothing from this comparison was pushed and no mutation was left in any tree. This branch's local state is one unpushed merge of `a50793ea`; the pushed head is unchanged at `117e1627`.

---

hjosugi:
Selection is complete: closing this PR without merge and preserving its branch. Its negative-index and compile-time-bounds semantics conflict with the accepted List behavior. Continue from `origin/issue/919-list-int-values@b984cd7c5abf5b4b49bdd8ae12ae7db8426c32a9`, after applying the four required repairs recorded on #919 and integrating final current main. No compiler branch is deleted.

Issue #942, opened 2026-08-02
Member-scope duplicates: methods, associated values, and inherited or default members


hjosugi:
## Metadata

- State: ready
- Tracker role: curated child
- Kind: implementation
- Area: Modules and Diagnostics
- Priority: P2
- Size: M
- Parent: #298
- Normative namespace input: `spec/modules/namespaces.md`
- Decision dependency: discharged; RFC-0005 was accepted 2026-08-09
- Frontend dependency: discharged; the two-member representation landed at `8e07d273`
- Blocked by: none
- Audited against: `origin/main@9fc27d8da59aa4d5ba92e24d7e76bd05be575257`

## Goal

Give the remaining member declaration domain a deterministic duplicate rule and one canonical diagnostic.

## Normative status

`spec/modules/namespaces.md` already places fields, methods, and associated functions/constants in the value namespace inside a member scope keyed by the owning type or trait. Equal normalized names under one owner therefore need one collision rule; the same spelling under distinct owners remains legal.

RFC-0005 (`rfcs/0005-trait-member-scope-closure.md`) is still **proposed**, not accepted. #995 remains `needs-decision`, so its closed-member/no-inheritance proposal is evidence for review rather than an implementation contract.

## Current executable boundary

Measured on the audited main commit:

- `member_name_collision` is refused first as `E2S127` because the parameter table reports `left` twice; the message names neither member `equal` nor owner `Equal`.
- A differently named second method reaches the explicit `method_count != 1` boundary and is refused as `E2S132`.
- `inherited_member_source` is refused as the delimiter error `E2S127`; no inheritance edge exists.
- Default method bodies remain refused as `E2S132`.
- `E370`–`E372` remain proposal-only codes with no executable emitter or diagnostic-registry row.

The trait gate pins those current refusals and the one-method typed boundary, but does not satisfy this issue's member-scope diagnostic acceptance criteria.

## Implementation scope after unblock

- Represent at least two value-namespace members under one stable owning type/trait identity.
- Apply the key `(owner identity, value namespace, normalized member name)`.
- Reject a same-owner collision with one canonical diagnostic carrying both declaration identities, both spans, the collided namespace/rule, and one remedy.
- Keep the same spelling legal under distinct owner identities.
- Apply the accepted result of #995 without inventing inheritance, precedence, or overload resolution.
- Register emitted diagnostics and add exact positive/negative fixtures.

## Out of scope

[...]

Issue #31, opened 2026-07-19
[Umbrella] Deliver production generics, traits, KIF, strategy evidence, and generic proof


hjosugi:
## Metadata

- State: blocked
- Tracker role: implementation umbrella
- Kind: planning
- Area: Advanced Type System / Modules / Laws
- Priority: P1
- Roadmap milestone: M2-alpha
- Size: L
- Direct delivery children: #1265–#1280
- Audited against: `origin/main@78b409c7ba69ea8fe21bdaea911197430f0e05b8`
- Normative input: `spec/roadmap-31-34/generics-and-traits.md`

## Outcome

Deliver the first production, separately compilable generics-and-traits profile, measured lowering strategies, and one genuinely generic proof accepted by a small trusted kernel.

This parent is coordination only. It closes after every child closes and receives no direct implementation patch.

## Verified current boundary

Already implemented and retained as evidence:

- explicit generic function declarations/calls, stable TypeParameterIds, and substitution in the standalone typed-only `generics_frontend.c`;
- accepted coherence/orphan rules and closed member-scope semantics (#403/#995/#942);
- concrete trait/implementation selection and dictionary elaboration in standalone typed HIR (#332/#923);
- closed host C11 execution for Equal[Int] and Ordered[Int] dictionaries (#1104);
- literal Int const-generic record specialization on the production C11 path (#916);
- bounded KIF function/flat-ADT facts and concrete law-evidence v2.

Not implemented:

- production nominal generic records/ADTs or generic functions;
- production trait/bound/dictionary HIR and C11 lowering;
- generic/trait facts consumed source-free through KIF;
- dictionary/monomorphic/hybrid mode equivalence and measurements;
- generic proposition obligations or a trusted certificate checker.

The old child topology is fully retired: #329/#471/#115/#256 closed not-planned, while #330/#332/#403/#923/#995/#942/#1104 closed completed. None is a live owner for the remaining work.

## Delivery DAG

```text
decision #1265
  ├─> nominal HIR #1268 ─> generic records #1269 ─> generic ADTs #1270
  └─> generic functions #1271 ─> trait/dictionary HIR #1272 ─> C11 dictionaries #1273

KIF decision #1266 ─> codec/model #1274
#1268 + #1272 + #1274 ─> production KIF producer/resolver #1275

#1271 + #1273 + #1275 ─> three lowering modes #1276 ─> measured benchmark #1277

proof decision #1267
[...]

PR #1013, opened 2026-08-02
rfc: publish RFC-0005, trait member scope closure, as proposed (#995)

### LINKED ISSUE #995: Trait member override v1: inherited and default member collision contract
## Metadata

- State: ready
- Tracker role: curated decision child
- Kind: spec design
- Area: Modules / Type system
- Priority: P2
- Size: S
- Parent: #31
- Blocks: #942
- Blocked by: none
- Audited against: `origin/main@9fc27d8da59aa4d5ba92e24d7e76bd05be575257`

## Decision required

Define how an owning trait/type member scope treats a declaration whose normalized value-namespace name is already supplied by an inherited member or a trait default. The current namespace contract places fields, methods, and associated functions/constants in the value namespace under the owning type or trait; it does not yet accept inheritance, defaults, or override syntax.

## Proposal lifecycle

RFC-0005, “A trait or type member scope is closed under direct declaration,” is published in `rfcs/0005-trait-member-scope-closure.md` and the RFC ledger as:

- state: `proposed`
- opened: 2026-08-02
- review closes: 2026-08-16
- `decided_on`: absent, as required for a proposal still under review

The proposal recommends a closed v1 scope: no inherited member, no default member, and no override marker. That is a review input, not yet an accepted rule. #942 remains blocked and must not implement proposal-only `E370`–`E372` contracts before this issue records the decision.

The current frontend cannot answer by accident: it accepts exactly one method per trait, refuses a duplicate-member fixture first as parameter-scope `E2S127`, rejects a differently named second method and default bodies with `E2S132`, and has no accepted supertrait or associated-value source shape.

## Options to settle

1. Require an explicit `override` marker and reject every unmarked inherited/default collision.
2. Allow signature-compatible implicit replacement and define exact compatibility plus ambiguity rules.
3. Keep v1 closed: reject inherited/default replacement entirely and defer an override marker to a later RFC.

The decision must state what happens when two inherited paths provide the same normalized value name, whether associated values follow the same rule as methods, and whether fields ever participate in inherited lookup.

## Recommended v1

Choose option 3 unless #31 first accepts concrete inheritance/default/override syntax. A closed refusal is deterministic, does not invent syntax inside a diagnostic issue, and leaves later override semantics to a separate bounded RFC.

## Acceptance criteria

- [ ] On or after the review close, RFC-0005 records the selected lifecycle state and `decided_on` together.
- [ ] One normative rule defines owner identity, namespace, normalized-name key, and direct/inherited/default precedence or their explicit absence.
- [ ] Same-owner duplicate methods and associated values have an unambiguous result.
- [ ] Diamond or multi-path inherited collisions have an unambiguous result, including when they are impossible by construction.
- [ ] The rule states whether fields can collide with methods/associated values and whether fields participate in inheritance.
- [ ] The chosen rule names the required positive and negative fixtures for #942.
- [ ] #31 and #942 are updated with the accepted dependency/result.
- [ ] RFC registry, trait, diagnostic, repository, and full verification gates pass on the deciding change.

## Resume condition

Do not close or implement the decision before the review window ends. On or after 2026-08-16, record the actual accepted/rejected result, update #31 and #942 atomically with that result, and rerun the decision gates. Green proposal-publication CI is not evidence that the proposal was accepted.

## Out of scope

Frontend lowering, general trait coherence, overload resolution, LSP rename identity, and implementation of default bodies.

hjosugi:
**This publishes a proposal. It decides nothing and closes nothing.**

RFC-0005 is filed as `proposed`, `opened_on: 2026-08-02`, `review_closed_on: 2026-08-16`, with no `decided_on`. #995 stays `needs-decision` through that window. The decision is recorded separately, on or after the day review closes, in a later change.

An earlier revision of this description said "Decides #995" and "Closes #995". That was wrong and is withdrawn: `tests/rfc/validate-registry.mjs:246-264` holds that a `proposed` RFC has not been decided, #995 requires an *accepted* dependency result before #31 and #942 can be updated, and #942 cannot resume until #995 accepts. A proposal cannot close an issue that requires acceptance. Nothing here changes #942's blocker or #31's child state.

## What is proposed

> **A member scope is closed under direct declaration.** The members of a trait or nominal type are exactly the members its own declaration lists. Two members of one owner with equal `(owner identity, namespace, normalized member name)` keys are refused. There is no inherited member, no default member, and no override.

Under the proposal v1 would permit **no** collision with an explicit override, because it would have nothing to override. The four questions #995 asks, and the answers the proposal offers for review:

| Question | Proposed answer |
|---|---|
| direct / inherited / default precedence | none — one candidate per key by construction, so an order would describe machinery M2-alpha does not have |
| same-owner duplicate methods and associated values | refused, `E370`; associated values would follow methods exactly — same namespace, same key, same code |
| diamond / multi-path inherited collision | unreachable by construction; `E371` would refuse the clause that creates an edge. The invariant a future inheritance RFC must preserve is stated: *the member set of an owner must remain a function of the owner identity alone* |
| fields | a field would collide with an associated value under one **type** owner (`E370`), never with a trait method — a trait's scope is keyed by `TraitId` and a type's by `TypeId`, and an implementation does not merge them. Fields would not participate in inheritance because nothing would |

An implementation would not be an owner: it supplies the slots its trait already declares and cannot introduce a member name.

## Current observed behaviour

This part is an observation about the frontend, established by running the shapes through `bootstrap/stage2/traits_frontend.c`. It holds regardless of how #995 is decided, and it **corrects a premise recorded in #942**, which says a duplicate member is refused by `E2S132` for carrying two methods. It is not:

```text
$ cat member_name_collision.kofun
trait Equal[T] {
    fn equal(left: T, right: T) -> Bool
    fn equal(left: T, right: T) -> Bool
}
$ kofun-traits-frontend member_name_collision.kofun out.ir out.tokens
error[E2S127]: parameter 'left' is declared twice at bytes 70..77
```

The parameter table is keyed by the trait rather than by the member, so the duplicate **parameter** is found first. The message names neither the colliding member `equal` nor its owning trait `Equal`.

`E2S132` is what a *differently* named second member gets, and a same-named second member with differently spelled parameters gets the identical text — the two differ only in their byte span, so the frontend cannot distinguish a duplicate from a second member at all:

```text
# fn equal(left, right) + fn differ(a, b)
error[E2S132]: a trait with 2 methods is unsupported in this slice; exactly one is accepted at bytes 0..92
# fn equal(left, right) + fn equal(a, b)
error[E2S132]: a trait with 2 methods is unsupported in this slice; exactly one is accepted at bytes 0..91
```

Neither inherited-member spelling is recognised as a trait construct; both are punctuation refusals:

```text
error[E2S127]: expected '{' at bytes 66..73      # trait Derived[T] extends Base[T]
error[E2S127]: expected ']' at bytes 64..65      # trait Derived[T: Base[T]]
```

And the default-body refusal names the implementation *slice*, so a reader is entitled to conclude the next slice will accept it:

```text
error[E2S132]: a default method is unsupported in this trait frontend slice; this slice is bounded to one-method traits with one type parameter, concrete implementations, and one explicit non-recursive bound at bytes 57..58
```

## Why #936's identity derivation stays sound under the proposal

A `DictionaryId` is its `ImplementationId` retagged `impl:` → `dictionary:` with the trailing `/decl=N` dropped, so what survives is exactly the coherence key. The slot layout comes from a separate record keyed by the trait alone — `dictionary-descriptor:abi1/` followed by the `TraitId`. Two properties hold today and the gate counts both:

1. a trait's slot table is a function of its `TraitId`; and
2. a `DictionaryId` determines the dictionary's contents — the descriptor for the trait in the key, filled by the unique admissible implementation for that key.

Closure is what would keep (1) true: with no supertrait path and no default, a trait's member sequence cannot depend on anything but its own declaration, so the descriptor cannot vary with what a supertrait contributed or with which of two colliding members won.

Implicit replacement would break (2) directly. One coherence key could denote different slot fillers depending on which member won, and the `DictionaryId` carries no component recording a winner — `/decl=N`, the one component that could have distinguished them, is precisely what the derivation drops, and it is dropped so reordering declarations cannot change a dictionary identity. Repairing it would mean growing the `DictionaryId` a member-resolution component, reintroducing the declaration-order sensitivity `order_independence.kofun` exists to refuse.

## Rejected alternative

**Implicit signature-compatible replacement** (#995 option 2), on three independent grounds: it breaks the derivation above; "signature-compatible" is a subtyping or variance judgment M2-alpha does not have; and choosing between two compatible candidates is ordered fallback, which DD-032 states is unsupported — the only orders available are declaration, import and link order, and DD-032 states none of them may ever select.

**An explicit `override` marker** (#995 option 1) is the right long-term shape but is rejected for v1: a marker needs something to mark, and with no inheritance edge and no default body it would be syntax no legal program could use, while reserving the word would break `override` as an identifier. The trade-off is recorded rather than hedged: when defaults land, a v1 program whose member name coincides with a new default acquires a collision it did not have and must add the marker. That cost is bounded and belongs to the RFC that creates defaults, which can count it against the corpus of the day.

**Doing nothing** leaves a compiler reporting a duplicate member as a duplicate parameter.

## What stays out

Frontend lowering, trait coherence, overload resolution, LSP rename identity, and implementation of default bodies — #995's own out-of-scope list. Also out: any registry row. `E370`–`E372` are reserved in the `E3xx` design band (after RFC-0001's `E340`–`E344`, RFC-0002's `E350`–`E356`, RFC-0004's `E360`–`E364`; `grep -c 'E37' tests/diagnostics/registry.tsv` → `0`, `cut -f1 tests/diagnostics/registry.tsv | grep -cE '^E3'` → `0`) and deliberately **not** added to `tests/diagnostics/registry.tsv`, which admits a code only with an executable emitter and an observed golden. Any implementation belongs to #942, which stays blocked. No `spec/records-v1.md` row: trait family.

Compatibility is recorded as `additive` with a query and a count — 888 tracked `.kofun` sources at `727f9dac442ba8db56e8344dbd1e29112ea1e64d`, `0` naming an inheritance edge or an override marker, and `0` of `42` trait/implementation member scopes declaring more than one member. The ledger also records what the same query returns after this change (`890`, `1`, `45 1`) and names the two fixtures that account for both differences, so the number cannot rot on merge.

## Validation

Two negative fixtures pin what the frontend emits today. They assert current behaviour only and depend on no accepted semantics. `tests/conformance/traits/run.sh` requires each to exit 1 with a byte-exact golden and write no typed IR, **and** asserts that neither message names the colliding member, the owning trait, or the inherited source — so the goldens cannot be re-blessed into looking like member-scope diagnostics without the gate failing first. Both layers were proved by mutation: a drifted golden fails `cmp`, and a golden re-blessed to `member 'equal' of trait Equal is declared twice` fires the negative predicate.

Adding those two `.stderr` companions moved the repository error-companion count that `tests/typed-sidecar/stage2-events.sh` pins, which is why exact-head CI on the previous head went red while every focused gate passed. The pin was **re-measured on this tree**, not copied from the CI log: set deliberately wrong, run, and pinned to the reported `saw 256`. Only the `repository_error_cases` assertion changed; `diagnostic_cases` stays at 47.

All seven gates pass from a clean checkout of the pushed head `95382bb36a94fc520b20b92dd063e9f3e7975bff`:

| Gate | Result |
|---|---|
| `sh tests/conformance/traits/run.sh` | PASS (8 lines, incl. the new member-collision assertion) |
| `sh tests/conformance/generics/run.sh` | PASS |
| `sh tests/rfc/check-registry.sh` | PASS — 30 decisions (20 accepted, 5 implemented, 5 proposed), 21 mutations refused |
| `sh tests/diagnostics/check.sh` | 140 stable codes; 140 executable owners agree; 0 evidence gaps |
| `sh tests/assertions/check.sh` | PASS — 0 silent assertions remaining, 48 at zero |
| `sh tests/release/check-claims.sh` | PASS — 45 claims, 19 mutations refused, evidence current |
| `sh tests/typed-sidecar/stage2-events.sh` | PASS — repository error companions pinned at the measured 256 |

No new task, so no `tooling/task-help.mjs` group. `release/claims.json` is untouched, so the evidence pack needed no regeneration. `bootstrap/stage2/compiler.{c,kofun}` and `SHA256SUMS` are untouched (#919).

Related: #995 (proposal, stays open), #942 (blocked, unchanged), #31 (parent).

### DIFF
--- a/rfcs/0005-trait-member-scope-closure.md
+++ b/rfcs/0005-trait-member-scope-closure.md
@@ -1,0 +1,530 @@
+# RFC-0005: A trait or type member scope is closed under direct declaration
+
+- Shepherd: hjosugi
+- Opened: 2026-08-02
+- Status: proposed
+
+## Summary
+
+The members of a trait or nominal type are exactly the members its own
+declaration lists. v1 has no inherited member and no default member, so it has
+nothing to override, and it adds no override marker. Two direct declarations
+that normalize to one name in one namespace under one owner are refused, and
+the refusal names the owner, the namespace, the name, and both declarations.
+A trait that names an inherited member source, and a trait member that carries
+a default body, are each refused as a stated rule rather than as an
+implementation slice boundary.
+
+For someone writing Kofun this changes nothing that works today and two things
+that do not: a duplicate member stops being reported as a duplicate
+*parameter*, and a supertrait clause stops being reported as a missing brace.
+
+## Motivation
+
+[#942](https://github.com/hjosugi/kofun/issues/942) cannot implement a truthful
+member-scope duplicate diagnostic until this decision exists, and
+[#995](https://github.com/hjosugi/kofun/issues/995) is the decision it waits
+for. The cost is not hypothetical: every shape in that domain is refused today
+by a diagnostic that blames a scope the author did not write in.
+
+Run against `bootstrap/stage2/traits_frontend.c` at
+`727f9dac442ba8db56e8344dbd1e29112ea1e64d`:
+
+```text
+$ cat member_name_collision.kofun
+trait Equal[T] {
+    fn equal(left: T, right: T) -> Bool
+    fn equal(left: T, right: T) -> Bool
+}
+$ kofun-traits-frontend member_name_collision.kofun out.ir out.tokens
+error[E2S127]: parameter 'left' is declared twice at bytes 70..77
+```
+
+That is one member name declared twice under one owner, and the compiler
+reports a duplicate parameter. The parameter table is keyed by the trait
+rather than by the member, so the parameter collision is found first. The
+message names neither the colliding member `equal` nor its owning trait
+`Equal`, which makes it a refusal the author cannot act on: renaming `left`
+produces a different refusal, not an accepted program.
+
+#942 recorded the premise that this shape is refused by `E2S132` for carrying
+two methods. It is not. `E2S132` is what a *differently* named second member
+gets, and a same-named second member whose parameters happen to be spelled
+differently gets the identical message — the two differ only in their byte
+span:
+
+```text
+# trait Equal[T] { fn equal(left: T, right: T) -> Bool
+#                  fn differ(a: T, b: T) -> Bool }
+$ kofun-traits-frontend two_named_members.kofun out.ir out.tokens
+error[E2S132]: a trait with 2 methods is unsupported in this slice; exactly one is accepted at bytes 0..92
+
+# trait Equal[T] { fn equal(left: T, right: T) -> Bool
+#                  fn equal(a: T, b: T) -> Bool }
+$ kofun-traits-frontend renamed_parameters.kofun out.ir out.tokens
+error[E2S132]: a trait with 2 methods is unsupported in this slice; exactly one is accepted at bytes 0..91
+```
+
+So the frontend today cannot distinguish a duplicate member from a second
+member at all, and where it does distinguish them it distinguishes them
+wrongly.
+
+The inherited-member source shapes are no better. Neither spelling a reader
+would try is recognised as a trait construct:
+
+```text
+$ kofun-traits-frontend inherited_member_source.kofun out.ir out.tokens
+error[E2S127]: expected '{' at bytes 66..73      # trait Derived[T] extends Base[T]
+$ kofun-traits-frontend supertrait_bound.kofun out.ir out.tokens
+error[E2S127]: expected ']' at bytes 64..65      # trait Derived[T: Base[T]]
+```
+
+Both are punctuation refusals. The frontend has no inheritance edge, so it has
+no way to say so.
+
+A default body is refused, and the refusal is accurate but temporary in the
+wrong way:
+
+```text
+$ kofun-traits-frontend default_method.kofun out.ir out.tokens
+error[E2S132]: a default method is unsupported in this trait frontend slice; this slice is bounded to one-method traits with one type parameter, concrete implementations, and one explicit non-recursive bound at bytes 57..58
+```
+
+It names the *slice*. A reader is entitled to conclude that the next slice
+will accept it. This proposal would make the refusal the rule, so the message
+should say the rule.
+
+Three of those five programs are now tracked, so the transcripts above cannot
+rot silently: `member_name_collision` and `inherited_member_source` are added
+by this proposal, and `default_method` was already there.
+`tests/conformance/traits/run.sh` requires each to exit 1 with its byte-exact
+golden, and additionally requires the first two messages *not* to name the
+colliding member, the owning trait, or the inherited source — so a change that
+started reporting these in the member scope would fail the gate rather than
+quietly invalidate the motivation. `two_named_members` and
+`renamed_parameters` are one-off probes and are not tracked; their sources are
+the two comments above.
+
+## Detailed design
+
+### Owner identity
+
+An owner is the thing a member scope is keyed by.
+
+| Declaration | Owner identity |
+|---|---|
+| trait | its `TraitId` |
+| nominal type, record, enum/ADT | the `TypeId` of its outer nominal constructor |
+| implementation block | **not an owner** |
+
+An implementation is indexed by `ImplementationId` under
+[DD-032](../spec/roadmap-31-34/generics-and-traits.md) and is not a
+separately lookupable name. It supplies the members its trait already
+declares; it does not introduce a member name, and it cannot add one. That is
+already enforced — an implementation of a method the trait does not declare is
+refused, pinned by `tests/conformance/traits/method_name_mismatch.stderr`.
+
+Ownership here is the declaration-scope sense: which declaration a member hangs
+off. It is unrelated to package ownership under the DD-032 orphan rule, and
+unrelated to the `Managed`/`Owned` kind of RFC-0004. Two members collide or do
+not for reasons that are settled before either of those is computed.
+
+### The member key
+
+A member is keyed by:
+
+```text
+(owner identity, NamespaceId, normalized member name)
+```
+
+The namespace comes from
+[`spec/modules/namespaces.md`](../spec/modules/namespaces.md) unchanged. This
+proposal assigns no namespace and adds no fifth one:
+
+| Member form | Namespace | Tag |
+|---|---|---|
+| field | value | 0 |
+| method | value | 0 |
+| associated function | value | 0 |
+| associated constant | value | 0 |
+| associated type | type | 1 |
+| named law or proof declaration | meta | 3 |
+
+Normalization is the normalization the namespace contract already applies to a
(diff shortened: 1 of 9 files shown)


hjosugi:
## Independent audit — blocked on RFC lifecycle

Audited exact head `91dcd58de31fa5d1af593127a408188dbfd39f7d` against `main@a50793ea6802411fd6f7c752578fd9421669236f` in a detached worktree. The PR is mergeable, the eight paths are confined to RFC-0005 / RFC index / focused trait conformance evidence, and all six declared local gates pass.

The blocking contradiction is exact:

- `tests/rfc/validate-registry.mjs:246-264`: a `proposed` RFC “has not been decided” and cannot carry `decided_on`.
- `rfcs/index.json:678-689`: RFC-0005 is proposed from `2026-08-02` through `review_closed_on: 2026-08-16`, with no `decided_on`.
- `rfcs/0005-trait-member-scope-closure.md:503-507`: the row is proposed and the decided behavior is future #942 evidence.
- This PR says `Decides #995` and `Closes #995`; #995 requires an **accepted** dependency/result update to #31 and #942, while #942 still says #995 must accept before it can resume.

Resume as a proposal-publication PR only after removing the decision/closing claim and recording the proposed-not-accepted dependency state on #31/#942. Keep #995 `needs-decision` through review. On or after `2026-08-16`, record the actual decision plus `decided_on`, update #31/#942 with the accepted result, rerun the gates, and then close #995.

No branch or file was edited and no merge was attempted. The PR is returned to Draft to prevent premature closure of #995.

hjosugi:
## CI failure diagnosis: fixed on a newer head, no audit-side edit

The required **Kofun verification** failure belongs to superseded head `91dcd58de31fa5d1af593127a408188dbfd39f7d`:

- run: https://github.com/hjosugi/kofun/actions/runs/30732339210
- job/step: `Kofun verification / Run task verify`
- failing task: `stage2-events`
- exact terminal failure: `FAIL: expected all 254 repository error companions, saw 256`; then `task: Failed to run task "stage2-events": exit status 1` and `task verify` exited 201.

Root cause is local to this PR: the two new `.stderr` fixtures increased the repository error-companion count by two, but head `91dcd58d` did not update the bounded count in `tests/typed-sidecar/stage2-events.sh`.

The branch owner has already replaced that head with `95382bb36a94fc520b20b92dd063e9f3e7975bff`. The live diff now includes a ninth path, `tests/typed-sidecar/stage2-events.sh`, changing both the assertion and its diagnostic from 254 to the re-measured 256. CI rerun https://github.com/hjosugi/kofun/actions/runs/30732743333 is still in progress at that exact head. Release evidence is already green.

No audit-side file or branch edit was made because the assigned head moved under active owner work. Keep the PR Draft. CI restart condition: the required Kofun verification must complete successfully at the live exact head; if the head moves again, re-audit the new SHA. This fixes only the mechanical gate and does not clear the separate RFC lifecycle blocker recorded for #995.

Issue #961, opened 2026-08-02
Kofun unit testing v1: executable kotest runner and stdlib sample suites


hjosugi:
## Metadata

- State: complete
- Merged PR: #958
- Main commit: `d06ac21bbd25aefeeea69b855db8fb9c07fcf50e`
- Re-audited: 2026-08-02

## Goal

Provide an executable unit-testing path for Stage 2 Kofun and prove it against real, bounded standard-library samples rather than hand-written projection-only gates.

## Shipped implementation

PR #958 added the kotest framework, kofun unittest runner, 17 sample modules, 133 tests, and a task verify gate.

## Acceptance criteria

- [x] The runner discovers tests deterministically and exits non-zero for a failing suite.
- [x] Filtering, listing, keep-going, no-color, and watch behavior have executable coverage.
- [x] The framework self-check runs in generated test binaries and does not rely on unsupported mutable capture.
- [x] Each stdlib sample states its narrower boundary and tests the same source it executes.
- [x] Existing canonical stdlib gates remain truthful; sample coverage is not presented as making unsupported canonical modules compile.
- [x] Task help, assertions, examples, stdlib capabilities, release evidence, and full task verify pass on the exact merged tree.
- [x] The PR was updated onto current main and generated evidence was regenerated.

## Verification

GitHub Actions run 30729741635 checked out exact merged commit `d06ac21bbd25aefeeea69b855db8fb9c07fcf50e`. Job 91447752518 (`Kofun verification`) completed successfully after `task verify`, the bounded type-level programming contract, and the native Core build/run all passed. The independent backlog job failed only because its committed issue-state snapshot was stale; that repository-maintenance drift is tracked separately by #986/#987 and does not invalidate this issue's compiler/test acceptance evidence.

hjosugi:
Integration blocker recorded and lane skipped (2026-08-02).

- PR #958 head: `c94d8aad7c03f501549399534d167c24d8fb61a3`
- Audited current main: `41b531f73e6b751f385c62b4ef689992f676afef`
- GitHub state: `CONFLICTING / DIRTY`
- Exact merge-tree result: all source changes auto-merge; the sole content conflict is generated `artifacts/release-evidence/index.json`.
- Required checks: none are currently attached to the exact PR head, so there is no exact-head full-verification evidence to merge.
[...]

Issue #992, opened 2026-08-02
bindgen-c sanitizers: only the AddressSanitizer arm is proven armed, and the gate claims three


hjosugi:
## Metadata

- State: complete
- Tracker role: curated
- Kind: test-quality
- Area: Testing, Security, and Performance
- Priority: P2
- Milestone: M1-bootstrap
- Size: S
- Parent: #574
- Audited baseline: `00ce08f` (the merge of #971)

## Summary

`tests/interop/bindgen-c/check-sanitizers.sh` depends on three sanitizer arms
— AddressSanitizer, LeakSanitizer, and UndefinedBehaviorSanitizer — and proves
one of them is actually running. Two of its own stated guarantees would survive
their sanitizer being switched off.

## Current behavior and evidence

Measured on `main` at `00ce08f`, the merge of #971.

There is exactly one negative fixture, and it is a heap overflow:

```sh
git ls-tree --name-only origin/main tests/interop/bindgen-c/fixture/ | grep -iE 'negative|leak|overflow'
# expected current result: tests/interop/bindgen-c/fixture/kbfix_negative.c

grep -nE "assert_grep 'the negative fixture" -A1 tests/interop/bindgen-c/check-sanitizers.sh
# the three assertions name: AddressSanitizer, kbfix.c, heap-buffer-overflow
```

That proves ASan is linked and reporting. It does not reach the other two:

- **LeakSanitizer.** The gate's own header says *"detect_leaks is ON,
  deliberately … A runner that cannot run LeakSanitizer fails loudly here
  rather than passing quietly"*. Nothing in the gate leaks. `detect_leaks=1`
  is a request: on a platform where LSan is unavailable it is a startup
  warning, not an error, and every check stays green. The client-owned-handle
  contract the header calls the point of the gate would then be unchecked.
- **UndefinedBehaviorSanitizer.** Delete `,undefined` from `SANITIZE` and the
  negative fixture is still caught by ASan, so the gate stays green with UB
  checking silently absent.

Both were measured directly while preparing the narrower #900 candidate
(#989, closed as superseded by #971): a use-after-free is reported by
`AddressSanitizer: heap-use-after-free`, an unfreed allocation by
`LeakSanitizer: detected memory leaks`, and a signed overflow by
`runtime error: signed integer overflow`. Each names a different arm, and each
is silent when its arm is missing.

## Why now

#971 merged the sanitizer and fuzz gates and closed #900 and #901. This is the
[...]

PR #991, opened 2026-08-02
test(kotest): execute runner option contracts

hjosugi:
## Summary

- assert that `--no-color` emits no ANSI escape bytes
- suppress watch screen-control escapes whenever effective color is off, including `--no-color`, `NO_COLOR`, and redirected non-TTY output
- make watch signal handling exit through the shared stamp cleanup
- prove the default build-failure path stops before a later suite
- prove `--keep-going` runs that later valid suite while preserving a nonzero overall status
- prove `--watch` performs two exact successful runs in explicit no-color and redirected-auto modes
- reject ANSI escapes, failed markers, build failures, and unexpected watch stderr
- reap the background watcher on every gate exit, with a bounded hard-stop fallback

## Validation

- `sh tests/stdlib/kotest/check.sh` twice after the final effective-color change (PASS, 133 tests plus both watch modes)
- `nix shell nixpkgs#go-task -c task kotest` (PASS)
- `sh -n tooling/kotest/run.sh tests/stdlib/kotest/check.sh`
- post-run process and temporary-directory cleanup readback
- `graphify update .`
- `git diff --check`

## Scope and provenance

The branch includes current main `00ce08f093da2706156f88f80d8c2fb366006d35`. The semantic PR diff is limited to `tooling/kotest/run.sh` and `tests/stdlib/kotest/check.sh`; runtime fixtures are generated inside the gate temporary directory. This does not overlap the five-file #964 change.

The runner change is required because strict watch stderr and escape checks exposed two shipped behavior defects: plain output still emitted screen-control escapes, and signal cleanup deleted the stamp without leaving the polling loop.

Closes #961
Closes #990

PR #1014, opened 2026-08-02
kotest: exercise --keep-going and --watch (#990)

### LINKED ISSUE #990: kotest: --keep-going and --watch are implemented and never exercised
## Metadata

- State: complete
- Tracker role: curated
- Kind: test-quality
- Area: Formatter, Linter, and Documentation
- Priority: P2
- Milestone: M1-bootstrap
- Size: S
- Parent: #961
- Audited baseline: `d06ac21bbd25aefeeea69b855db8fb9c07fcf50e`

## Summary

`tooling/kotest/run.sh` implements five behaviours the runner documents in its
own usage block. The gate exercises three of them. `--keep-going` and
`--watch` are shipped, documented, and unproven.

Measured on `main` at `d06ac21bbd25aefeeea69b855db8fb9c07fcf50e`, where #958
merged.

The runner declares all five:

```sh
sed -n '6,8p' tooling/kotest/run.sh
#   tooling/kotest/run.sh [PATH ...] [--filter SUBSTRING] [--list]
#                         [--watch] [--no-color] [--keep-going]
```

The gate uses three:

```sh
grep -oE -- '--(filter|list|no-color|keep-going|watch)[a-z-]*' tests/stdlib/kotest/check.sh | sort | uniq -c
# expected current result:
#       2 --filter
#       3 --list
#       3 --no-color
```

Both missing flags reach real code paths:

- `tooling/kotest/run.sh:284` — `if ! $keep_going && [ "$unit_status" -eq 2 ]; then`
  stops the sweep on a unit that fails to **build**. `--keep-going` is what
  makes the remaining units run anyway. The existing failing fixture proves
  continuation *within* a suite after an assertion failure, which is a
  different path: that fixture compiles.
- `tooling/kotest/run.sh:322-341` — the watch loop re-runs the sweep, touches
  a stamp, then polls `find … -newer "$stamp"` once a second over the given
  paths and the framework library.

#961's acceptance criterion is "Filtering, listing, keep-going, no-color, and
watch behavior have executable coverage". Three of five is what is there.

## Why now

#961 is `verification-pending` on this exact criterion. Every other gate it
names passes on the merged tree, so this is the one thing between it and a
close.

## Goal

The kotest gate exercises `--keep-going` and `--watch`, so neither can be
broken silently.

### In scope

- A fixture that fails to **compile** (not to assert), paired with a passing
  suite, run twice: without `--keep-going` the sweep must stop and leave the
  second suite unrun; with it the second suite must run and the exit status
  must still be nonzero.
- A bounded `--watch` case: start the runner under `timeout`, wait for the
  first summary, touch a `.kofun` file under a watched path, wait for the
  second summary, terminate it. Assert exactly that a change caused a re-run.

### Out of scope

- New sample suites or stdlib coverage; the 133 tests stay as they are.
- Changing the runner. If a behaviour turns out to be untestable as written,
  say so in the issue rather than reshaping the runner to suit its test.
- `--filter`, `--list`, `--no-color`, which already have coverage.

## Contract and constraints

- The watch case must be bounded: a hard `timeout`, a bounded wait for each
  summary, and no unbounded `sleep` loop. A gate that can hang is worse than
  the missing coverage.
- The poll interval is one second, so the case must wait for an observable
  effect rather than assume a fixed delay.
- Both cases run offline and leave no process behind; the existing gate's
  `trap 'rm -rf "$WORK"' 0 1 2 15` is the pattern.

## Implementation orientation

- `tests/stdlib/kotest/fixtures/failing_test.kofun` is the shape for the new
  build-failure fixture, which needs a type error rather than a failing
  assertion so `unit_status` is 2.
- The existing `--filter` and `--list` cases at the end of
  `tests/stdlib/kotest/check.sh` are the shape for the flag cases.

This is orientation, not a required approach.

## Acceptance criteria

- [ ] A fixture that fails to compile exists, and the gate asserts the sweep
      stops at it without `--keep-going`.
- [ ] The gate asserts that with `--keep-going` the later suite still runs and
      the overall exit status stays nonzero.
- [ ] The gate starts the runner with `--watch` under a hard timeout, observes
      one re-run caused by touching a watched `.kofun` file, and terminates it.
- [ ] `grep -oE -- '--(filter|list|no-color|keep-going|watch)' tests/stdlib/kotest/check.sh`
      reports all five flags.
- [ ] No orphaned process or temporary file survives the gate; a second run
      from a clean tree behaves identically.

## Validation

| Check | Command or artifact | Expected result |
|---|---|---|
| Focused | `task kotest` | all existing PASS lines plus the two new behaviours |
| Repeatability | `task kotest` twice | identical output; no leftover process |
| Assertion budget | `sh tests/assertions/check.sh` | still at zero |
| Regression | `task verify` | exit 0 |

## Dependencies

- Blocked by: none. #958 is merged; the runner and the gate both exist on
  `main`.
- Blocks: #961, whose second acceptance criterion this is.

## Documentation and compatibility

- `tooling/kotest/README.md` if it lists what the gate proves.
- Compatibility: none. Test-only change; the runner is untouched.

## Risks and open questions

- Risk: the watch case is timing-based and becomes the flakiest gate in the
  repository. Mitigation: bound every wait, assert on an observed summary
  rather than elapsed time, and if it still cannot be made deterministic,
  record that finding here and leave `--watch` documented as unexercised
  rather than shipping a flaky gate.
- Open questions: none.

hjosugi:
Closes #990. Unblocks #961's second acceptance criterion.

## What changed

`tests/stdlib/kotest/check.sh` gains two sections and one fixture. The runner is untouched, no sample suites are added, and the existing cases are unchanged.

The runner documents five behaviours in its own usage block (`tooling/kotest/run.sh:6-8`). The gate ran three:

```sh
grep -oE -- '--(filter|list|no-color|keep-going|watch)[a-z-]*' tests/stdlib/kotest/check.sh | sort | uniq -c
      2 --filter
      3 --list
      3 --no-color
```

`--keep-going` and `--watch` were shipped, documented, and unproven.

## Why the existing failing fixture did not already cover keep-going

`--keep-going` changes the handling of exactly one status — 2, a unit that fails to **build** (`run.sh:284`, reached from `BUILD FAIL` at :217 and `C BUILD FAIL` at :225). `fixtures/failing_test.kofun` compiles and then reports red assertions, which is a different path: continuation *within* a suite says nothing about whether the sweep stops at a suite that never built.

`fixtures/build_fail_test.kofun` is therefore a type error, not a failing assertion, and it is deliberately the only fault in the file — a fixture that fails for two reasons cannot show which one the runner reacted to.

Measured both ways:

| Run | Exit | Later suite |
|---|---|---|
| `run.sh DIR --no-color` | **2** | never runs |
| `run.sh DIR --no-color --keep-going` | **1** | runs, and the sweep stays red |

## The watch case, and why it is not a timing guess

Every wait is bounded (at most sixty one-second polls), asserts on an **observed summary** rather than on elapsed time, and the runner is killed whatever happens. A gate that can hang would be worse than the missing coverage.

The second summary has to be caused by the change rather than by the loop, so I measured that separately before trusting it:

```
first summary after 1s (count=1)
after 15s WITHOUT touching: count=1     <- the loop does not re-run on its own
after touch: count=2 in 1s
```

One ordering detail is load-bearing: the runner touches its stamp immediately after printing the summary, so a change made in that window is older than the stamp and never seen. The case waits for the stamp to settle rather than racing it.

## Two bugs found while writing it

- `grep -c` prints its count **and** exits 1 when that count is zero, so a `|| printf '0'` fallback produced `0\n0` and `test: Illegal number`. The fallback has to replace the status, not add a line.
- `pgrep -f` matches the command line doing the searching. It reported a leftover runner that did not exist; `ps -eo pid,args | grep "[t]ooling/kotest/run.sh"` is what answers that question, and it shows none after the gate.

## Verification

| Check | Command | Result |
|---|---|---|
| Focused | `task kotest` | 5 PASS lines (was 3) |
| Repeatability | `task kotest` twice | identical output |
| Leftover processes | `ps -eo pid,args \| grep "[t]ooling/kotest/run.sh"` after the gate | none |
| Assertion budget | `sh tests/assertions/check.sh` | `0 remaining, 48 at zero` |
| Whitespace | `git diff --check` | clean |
| Full regression | `task verify` | **exit 0**, 1070 PASS, 0 failures |

Gate cost: 20s → 32s.

## One thing worth knowing about `task verify`

My first full run failed, and not on this change: `tests/lsp/performance_test.js:275` asserts `diagnostic p95 <= 100ms` and measured 126.14ms while I had other heavy work running on the same box. Re-run idle, that lane passes (`sh spec/roadmap-31-34/verify-current-gates.sh` → exit 0), and so does the full suite. The test's own comment calls the budget host-dependent, so this is a note rather than a bug report; if it recurs on an idle CI runner it deserves its own issue.

---

### DIFF
--- a/tests/stdlib/kotest/check.sh
+++ b/tests/stdlib/kotest/check.sh
@@ -118,6 +118,104 @@ sh "$RUNNER" "$SAMPLES/list_sample_test.kofun" --list \
 grep -q 'list_sample_test.test_push_appends_in_order' "$WORK/list.out" ||
     fail '--list did not enumerate tests'
 
+# -------------------------------------------------------------- keep-going
+# `--keep-going` changes the handling of exactly one status: 2, a unit that
+# fails to *build*. That is a different path from the failing fixture above,
+# which compiles and then reports red assertions, so continuing past an
+# assertion failure proves nothing about this flag.
+KEEP="$WORK/keep-going"
+mkdir -p "$KEEP"
+cp "$ROOT/tests/stdlib/kotest/fixtures/build_fail_test.kofun" \
+    "$KEEP/aaa_build_fail_test.kofun"
+cat >"$KEEP/zzz_later_test.kofun" <<'SUITE'
+fn test_runs_after_the_broken_unit() -> Int {
+    let mut failures = 0
+    failures = failures + expect_eq_int(1, 1)
+    return failures
+}
+SUITE
+
+set +e
+sh "$RUNNER" "$KEEP" --no-color >"$WORK/keep-off.out" 2>&1
+keep_off_status=$?
+set -e
+test "$keep_off_status" -eq 2 ||
+    fail "a build failure without --keep-going exited $keep_off_status, expected 2"
+grep -q 'BUILD FAIL' "$WORK/keep-off.out" ||
+    fail 'the stopped sweep did not report the build failure'
+grep -q 'zzz_later_test' "$WORK/keep-off.out" &&
+    fail 'the sweep ran past a build failure without --keep-going'
+
+set +e
+sh "$RUNNER" "$KEEP" --no-color --keep-going >"$WORK/keep-on.out" 2>&1
+keep_on_status=$?
+set -e
+test "$keep_on_status" -eq 1 ||
+    fail "--keep-going exited $keep_on_status, expected 1"
+grep -q 'BUILD FAIL' "$WORK/keep-on.out" ||
+    fail '--keep-going hid the build failure'
+grep -q 'zzz_later_test.test_runs_after_the_broken_unit' "$WORK/keep-on.out" ||
+    fail '--keep-going did not run the suite after the broken unit'
+
+# ------------------------------------------------------------------- watch
+# The watch loop re-runs the sweep, touches a stamp, then polls once a second
+# for a `.kofun` newer than that stamp. Every wait here is bounded and asserts
+# on an observed summary rather than on elapsed time, and the runner is killed
+# whatever happens: a gate that can hang is worse than the missing coverage.
+WATCH="$WORK/watch"
+mkdir -p "$WATCH"
+cat >"$WATCH/watched_test.kofun" <<'SUITE'
+fn test_watch_target() -> Int {
+    let mut failures = 0
+    failures = failures + expect_eq_int(2, 2)
+    return failures
+}
+SUITE
+
+# `grep -c` prints the count and still exits 1 when that count is zero, so the
+# fallback has to replace the status rather than add a second line.
+watch_summaries() {
+    seen=$(grep -c 'suites)' "$WORK/watch.out" 2>/dev/null || :)
+    printf '%s\n' "${seen:-0}"
+}
+await_summaries() {
+    want=$1
+    waited=0
+    while test "$waited" -lt 60; do
+        if test "$(watch_summaries)" -ge "$want"; then
+            return 0
+        fi
+        sleep 1
+        waited=$((waited + 1))
+    done
+    return 1
+}
+stop_watch() {
+    kill "$watch_pid" 2>/dev/null || :
+    wait "$watch_pid" 2>/dev/null || :
+}
+
+: >"$WORK/watch.out"
+sh "$RUNNER" "$WATCH" --no-color --watch >"$WORK/watch.out" 2>&1 &
+watch_pid=$!
+
+if ! await_summaries 1; then
+    stop_watch
+    fail 'the watch run never produced a first summary'
+fi
+# The stamp is touched immediately after the summary is printed, so a change
+# made in that window is older than the stamp and is never seen. Let it
+# settle rather than racing it.
+sleep 2
+touch "$WATCH/watched_test.kofun"
+if ! await_summaries 2; then
+    stop_watch
+    fail 'touching a watched file did not re-run the sweep'
+fi
+stop_watch
+
 printf 'kotest framework and runner behaviour: PASS\n'
 printf 'kotest failing-suite detection and exit codes: PASS\n'
+printf 'kotest --keep-going continues past a build failure and stays red: PASS\n'
+printf 'kotest --watch re-runs the sweep when a watched source changes: PASS\n'
 printf 'kotest stdlib sample suites (%s tests): PASS\n' "$total"
--- a/tests/stdlib/kotest/fixtures/build_fail_test.kofun
+++ b/tests/stdlib/kotest/fixtures/build_fail_test.kofun
@@ -1,0 +1,14 @@
+# Deliberately unbuildable kotest suite (#990).
+#
+# This fixture fails to *compile*, which is a different runner path from
+# fixtures/failing_test.kofun: that one builds and then reports red
+# assertions. A unit that cannot be built returns status 2, and status 2 is
+# the only status `--keep-going` changes the handling of — without the flag
+# the sweep stops at this suite, with it the remaining suites still run.
+#
+# The type error is deliberate and must stay one: a suite that fails for two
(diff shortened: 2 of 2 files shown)


hjosugi:
Closing as superseded. `main` at `a50793e` already covers both flags, so this would duplicate the same file:

```sh
grep -oE -- '--(filter|list|no-color|keep-going|watch)[a-z-]*' tests/stdlib/kotest/check.sh | sort | uniq -c
      2 --filter
      6 --keep-going
      3 --list
      7 --no-color
     10 --watch
```

I checked what is on `main` against what this PR carries before closing, and `main` is stronger on every axis I compared: the watch case runs in **two** modes and asserts the ANSI escape byte is absent under `--no-color` and present otherwise; `stop_watch` escalates to `SIGKILL` and a `cleanup` trap reaps the runner on any exit path, where this PR only did `kill` then `wait`; and `TMPDIR` is isolated so the watch stamp cannot collide with anything else. The keep-going case reaches the same `BUILD FAIL` path from an unknown-function error instead of a type error, and generates its fixtures in the work directory rather than committing an unbuildable `.kofun` — which is the better call, since it keeps a file that cannot compile out of the tree.

One difference is worth recording rather than losing, and it is small: `main`'s watch case appends a newline **on every attempt** of its second wait loop, so it proves a re-run happened but not that the change caused it. A watch loop that re-ran on its own — a real bug, since the loop is supposed to block on `find … -newer` — would still satisfy it. While writing this PR I measured the control directly and the loop is currently well-behaved:

```
first summary after 1s (count=1)
after 15s WITHOUT touching: count=1     <- does not re-run on its own
after touch: count=2 in 1s
```

If that control is wanted later, it is one idle wait before the first write. Not worth a PR on its own against coverage that is otherwise better than mine.

#990 and #992 are both closed and satisfied on `main`; I verified the sanitizer side by running it, and it now reports `a leaked client-owned handle proves LeakSanitizer is armed` and `a library-side signed overflow proves UndefinedBehaviorSanitizer is armed`, which is exactly the gap I raised on #971.

---

hjosugi:
Final provenance readback for exact head `03417061a16cccfa19115f6bcf43cd2df138000a`:

- `03417061` has parent `00ce08f` and changes only `tests/stdlib/kotest/check.sh` plus a committed unbuildable fixture.
- PR #991's exact head `c358882e7b4bdb238974b4523e0639be513a687a` changed the same gate plus the runner fix, passed required CI run `30730875853`, and merged as `e9241a533863227e5645d1beef98dea2ba997039`.
- #990 is already `CLOSED / COMPLETED`, with that exact completion provenance recorded on the issue.
- Against current `main@a50793ea6802411fd6f7c752578fd9421669236f`, `git merge-tree --write-tree` reports exactly one conflict: `tests/stdlib/kotest/check.sh`. Both branches independently insert keep-going/watch coverage after the same `--list` block from merge base `00ce08f`; this is overlapping replacement coverage, not an unrelated main regression.
- The main version covers the acceptance contract and is materially stronger: it tests watch in two effective-plain modes, pins two successful summaries and green rows, rejects ANSI/failure/stderr output, uses bounded TERM/KILL cleanup, isolates `TMPDIR`, and includes the runner fixes strict testing exposed. The build-failure source is generated under the gate temp directory, so no deliberately unbuildable source is committed. `03417061` adds no unique #990 acceptance value worth conflict resolution.

The current `CLOSED` state is therefore correct; drafting or merging this PR would add no value. The remote branch `claude/research-issues-ready-ju8mf3` is unprotected and has no open PR consumer (only closed/merged historical PRs), so it is safe to delete after confirming there is no unpushed local-only work. This audit proposes that deletion but does not perform it.

Issue #1032, opened 2026-08-02
Self-host driver diagnostics: give A1's refusal a code and a span


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation child
- Kind: implementation
- Area: Self Hosting / Diagnostics
- Priority: P2
- Roadmap milestone: M1-bootstrap
- Size: M
- Parent: #622 (closed)
- Revision procedure owner: #618 (closed; procedure still applies)
- Refined baseline: `origin/main@70ff1b2b35fbdf1cc3084b96d283bc07e84f35de`

## Goal

Give every bounded A1 refusal a stable diagnostic class and exact source anchor while preserving byte-identical output between the compiler built from canonical `S` and the independently audited Stage 1 C seed.

## Current measured boundary

Measured on exact `origin/main@70ff1b2b35fbdf1cc3084b96d283bc07e84f35de`.

- `EA101`–`EA105` do not yet occur under `bootstrap/stage1`, `bootstrap/selfhost`, or `tests/diagnostics`.
- All **62** A1 refusals still use `bootstrap/selfhost/driver/corpus_reject.stdout`: one base fixture, 31 suffixed file fixtures, and 30 builtin-matrix rows.
- `bootstrap/selfhost/check-compiler-driver.sh` builds both `bootstrap/selfhost/driver/S.c` and the independently audited `bootstrap/stage1/compiler.c`; their stdout, stderr, exit status, and absence of C artifacts are differential evidence.
- `sh bootstrap/stage1/check.sh` and `sh tests/diagnostics/check.sh` pass at the baseline.
- Parent #622, procedure owner #618, and former overlap #1018 are closed. No open PR, branch/ref, live canonical claim, or other worktree owns this bounded diagnostic surface at refinement time.
- #314 owns Stage 2 `bootstrap/stage2/compiler.kofun`, `bootstrap/stage2/compiler.c`, and `bootstrap/stage2/SHA256SUMS`; those paths are excluded. Stage 2 may only be executed as the existing trusted generator.

## Closed diagnostic contract

| Code | Class | Exact population | Exact message |
| --- | --- | ---: | --- |
| `EA101` | unsupported construct | base `corpus_reject.kofun` (`match`) — 1 | `unsupported construct in Stage 1 Core` |
[...]

Issue #622, opened 2026-07-23
Self-host compiler driver: args, file I/O, and compiling S to deterministic C11


hjosugi:
## Metadata

- State: ready
- Tracker role: curated
- Kind: implementation
- Area: Self Hosting / Compiler driver / C11 backend
- Priority: P0
- Roadmap milestone: M1-bootstrap
- Size: M
- Parent: #32
- Prerequisite: #618 (complete)
- Split-out child: #1032 (structured refusal code/span, changes `S`)

## Outcome

Make the compiler produced from canonical Kofun compiler source `S` accept ordinary canonical `S` and emit deterministic nonempty C11 for the same frozen profile.

That capability now works and is gated. What remains is the host-boundary safety evidence this issue also owns: the A1 shim is the one hand-written C surface in the self-host chain and nothing runs it under a sanitizer, and two of its failure paths behave correctly but are pinned by nothing.

## Current frozen input

Audited against `main` `f7c3cfa3f21632a6f0ce18fb94745922e2b47e31`:

- canonical `S`: `bootstrap/stage1/compiler.kofun`;
- SHA-256: `3d6025f5a951f635fe032e9eb97cb8d873d3837cd113f29c73eb230ba8c194b7`;
- profile: 46 rows, all `complete` across all four `kofun.selfhost-profile/v2` evidence classes;
- builtin/host boundary: 15 profile builtins, plus `fail` added by #753;
- accepted revision: #751 / `d3429207447e112f5230d5be38576959130d288e`, with the evidence contract completed by #908 / `6f71b6df5cf54aa16542b89494d2194c4172a97e`.

The earlier `cdcb3d…` / 45-row / #704 contract recorded here is obsolete; eight `S` revisions (#744–#751, #753) landed after it. #618's revision table carries the full history.

## Measured current behavior

```
$ cc -std=c11 -O2 -Wall -Wextra -Werror -I unicode \
     bootstrap/selfhost/driver/S.c -o /tmp/kofun-a1
$ /tmp/kofun-a1 bootstrap/stage1/compiler.kofun /tmp/C2.c
status: 0
C2.c: 605357 bytes, 15333 lines

rerun                                       → C2 byte-identical
recompiled from another directory,
  under another source name                 → C2 byte-identical
repository root appearing in C2             → 0
cc -std=c11 -pedantic-errors -Wall -Wextra -Werror -I unicode -c /tmp/C2.c → exit 0

$ task selfhost-self-compile   → exit 0, ending with
[...]

PR #1049, opened 2026-08-06
selfhost: sanitize the A1 host boundary and pin its I/O refusals (#622)

hjosugi:
Closes the two remaining acceptance criteria on #622. One reviewable change to `bootstrap/selfhost/check-compiler-driver.sh`, as that issue scopes it.

#1050 implemented the same criteria independently and was closed in favour of this one; its three review points are folded in as of `dbddfbef` (see [the reply](https://github.com/hjosugi/kofun/pull/1049#issuecomment-5209333219)).

## Why

`bootstrap/selfhost/driver/S.c` carries the one hand-written C surface in the self-host chain — argv decoding, file bytes, allocation — and every run in this gate was uninstrumented. On this commit `grep -ril fsanitize bootstrap/ tests/ Taskfile.yml` returns 28 files and still names no `selfhost` file.

So an out-of-bounds read in that shim, or a signed overflow in the generated body, emitted the right C2 bytes and exited 0, and all ten PASS lines printed anyway.

## Premise re-verified before changing anything

`docs/CONTRIBUTING.md` asks for this, and `main` had moved since the issue's stamp. Measured on `2b7581ff`:

```
$ cc -std=c11 -O2 -Wall -Wextra -Werror -I unicode \
     bootstrap/selfhost/driver/S.c -o /tmp/kofun-a1
$ /tmp/kofun-a1 bootstrap/stage1/compiler.kofun /tmp/C2.c
status: 0        C2.c: 605357 bytes
```

605357 bytes, matching the issue exactly. Both ungated refusal paths reproduce verbatim, so the premise holds and the work is still owed.

## 1. Sanitizer coverage

A1 is built a second time under ASan+UBSan with `-fno-sanitize-recover=all`. Through it go the nine accept corpora, **both refusal paths**, and the full `A1(S) → C2` self-compile, requiring exit 0 (exit 1 for the refusals), empty stderr, and C2 bytes identical to the ordinary build. That last comparison matters: if instrumentation changed the answer, the clean run would be vouching for a binary nobody ships.

The refusal paths are covered because an accept-only corpus cannot reach the shape that matters — a refusal abandons an allocation mid-flight, on a path that unwinds rather than completes. On both of those paths the diagnostic goes to stdout, which is what makes "stderr must be empty" a clean sanitizer-only signal. The unwritable-output case is deliberately not in this set: its runtime error legitimately goes to stderr and would need a different assertion shape.

### The never-free model is asserted, not assumed

`detect_leaks` is OFF — the opposite of `tests/interop/bindgen-c/check-sanitizers.sh`, for the opposite reason. That fixture's contract is client-owned handles, so a leak *is* the defect. `kofun_rt_alloc` is a `malloc` wrapper, `S.c` defines no deallocation entry point, and reclamation is an open design item, so LeakSanitizer reports the allocation model rather than a defect. Measured: 5000+ reports on `A1(S)`.

The risk is that the reason outlives its truth. If a deallocation path is ever added, `detect_leaks=0` silently stops describing the design and starts suppressing a defect, so the gate now asserts the model and fails loudly instead.

The check looks for a deallocation **entry point**, not the token `free(`:

```
$ grep -c 'free(' bootstrap/selfhost/driver/S.c
2
$ grep -n 'free(' bootstrap/selfhost/driver/S.c
10588:        const char *k_n8073 = "        free(allocation);\n";
10730:        const char *k_n8213 = "        free(allocation);\n";
```

Both sit inside string literals A1 emits into the programs it compiles. The compiled output frees; the compiler does not.

### Other load-bearing details

- **The `ulimit -v` ceiling is NOT applied to the sanitized run.** ASan reserves a shadow mapping far larger than the 1.5 GiB bound, so the limit would abort the run before it started and the failure would read as a memory-safety fault instead of a harness mistake.
- Compiler selection mirrors bindgen-c: clang first, then gcc, then cc, because clang links the sanitizer runtimes only when compiler-rt sits beside it. On this machine clang cannot (`cannot find libclang_rt.asan-x86_64.a`) and gcc can, which is exactly the case that probe exists for. No available compiler is a failure, not a skip.
- Instrumentation is proven present via `readelf --wide --syms --dyn-syms`, reading **both** tables: `-static-libasan` leaves `.dynsym` empty while a stripped binary leaves `.symtab` empty, so requiring either alone turns a packaging choice into a false failure or a false pass.

### Instrumented and armed are separate claims

They are independent and neither substitutes for the other, so they get separate PASS lines: *instrumented* says the binary that ran carries `__asan_`; *armed* says one isolated fault per arm trips that arm's own diagnostic and not the other's.

The arming probes caught a real substitution during development. The first ASan probe used a literal `malloc(4)` and the gate failed with `runtime error: store to address ... with insufficient space for an object` — UBSan's object-size check reporting first, so the probe vouched for AddressSanitizer while proving only that UBSan worked.

The fix is subtler than it looks, and the intuitive version does not work. Measured with the gate's own flags:

```
volatile size, plain pointer  -> 0 heap-buffer-overflow, 1 UBSan report
plain size, volatile pointer  -> 2 heap-buffer-overflow, 0 UBSan reports
```

The object-size check resolves the object *through the pointer*, so the pointer is what must lose its provenance; hiding the size changes nothing. Both measurements are recorded beside the probe, because anyone simplifying the `volatile` off the pointer would silently revert it to proving the wrong arm.

## 2. The two ungated refusal paths

Invalid UTF-8 is compared **whole** rather than grepped, so line, column and byte offset are pinned too — a span that silently became 0 would still contain the code. `EUNICODE001` is registered for the native backend only (`tests/diagnostics/registry.tsv`), so this is the one place the self-host boundary covers it.

The fixture uses octal escapes because POSIX `printf` specifies `\ooo` and does not specify `\xHH`. Under a shell whose printf lacks `\x`, a `\x` fixture becomes the literal characters, stops being invalid UTF-8, compiles fine, and the case passes while testing nothing.

An unwritable output path is pinned to its exact runtime diagnostic on stderr with empty stdout, and the invalid-UTF-8 case additionally asserts the prior output file is preserved byte for byte.

## A divergence this surfaced

Implementing the UTF-8 case exposed something no existing case covered: **A1 and the audited seed emit the same 65-byte `EUNICODE001` message on opposite streams.**

```
A1    stdout=65  stderr=0
seed  stdout=0   stderr=65
$ cmp a1.stdout seed.stderr   → identical
```

Every other refusal in this gate agrees on both streams and is asserted equal. Asserting equality here would fail. The divergence is real, predates this change, and was invisible precisely because nothing covered the path.

Each side is therefore pinned separately rather than asserted equal — either one moving is a failure, and the day they agree, that block is what says so. Deciding *which* stream is correct changes `S` or the seed, which #622 lists as a non-goal (`that is #1032's problem`), so it is recorded here rather than fixed.

## Validation

On `dbddfbef`:

| Check | Command | Result |
| --- | --- | --- |
| Self-host driver | `sh bootstrap/selfhost/check-compiler-driver.sh` | **PASS** — 15 lines: the existing 10, the two pinned refusals, and one each for the allocation model, instrumentation, and arming |
| Profile | `sh bootstrap/selfhost/check-profile.sh` | **PASS** (exit 0) |
| Frontend | `sh bootstrap/selfhost/frontend/check-frontend.sh` | **PASS** (exit 0) |
| C11 slices | `sh bootstrap/selfhost/c11/check-c11.sh` | **PASS** (exit 0) |
| Full regression | `task verify` | CI `Kofun verification` — passed on `f31346d5` ([run 31069623458](https://github.com/hjosugi/kofun/actions/runs/31069623458)), re-running on `dbddfbef` |

The four rows above were run locally; go-task is not installed in that environment, so the `task verify` row is claimed from CI rather than by hand. The last three rows were green on `f31346d5` and are unaffected by this commit, which touches only the one gate script no other gate sources.

## Boundary

No compiler source, canonical `S`, profile row, or checked artifact is touched — this is assertions in one file on one subject. The structured refusal code and span remain #1032's.


hjosugi:
I independently implemented the same two acceptance criteria on `agent/622-a1-host-boundary` before noticing this PR, and opened #1050 for it. #1050 is being closed in favour of this one, which came first.

Three things that branch asserted and this one does not. Recording them here so they are a choice rather than an oversight; each is a small addition to the section this PR already has.

**1. The never-free model is assumed, not asserted.** `detect_leaks=0` is correct — `kofun_rt_alloc` is a `malloc` wrapper and the prelude defines no deallocation entry point, so every Text lives until the process exits, and LeakSanitizer would report the allocation model rather than a defect. With `detect_leaks=1` the gate reports a five-figure leak on a clean run.

The risk is that the reason outlives its truth silently. If a deallocation path is ever added, leak detection should come back on, and nothing currently forces that conversation:

```sh
for deallocator in kofun_rt_free kofun_rt_dealloc kofun_rt_release; do
    if grep -q "^[a-zA-Z].*$deallocator" bootstrap/selfhost/driver/S.c; then
        fail "S.c defines $deallocator, so A1 no longer never-frees; re-enable detect_leaks"
    fi
done
grep -q '^void \*kofun_rt_alloc(' bootstrap/selfhost/driver/S.c ||
    fail "S.c no longer defines kofun_rt_alloc; the allocation model this gate reasons about has changed"
```

`grep -c 'free('` on S.c is not the check — `free(` appears there only inside the string literals A1 emits into the programs it compiles. The absence of a deallocation *entry point* is what holds.

**2. The refusal paths do not run instrumented.** This PR runs the corpus and `A1(S)` under the sanitizers, which is the bulk of the value. A refusal path is where an allocation is abandoned mid-flight, which is exactly the shape an accept-only corpus cannot reach — so the invalid-UTF-8 input and `corpus_reject.kofun` are worth pushing through the sanitized binary too, requiring the same empty stderr and byte-identical stdout as the uninstrumented run.

**3. Instrumented and armed are one claim here.** They are independent and neither substitutes for the other:

- *instrumented* — `__asan_` is present in the binary that actually ran. Without it a clean run cannot be distinguished from a sanitizer that was never linked.
- *armed* — one isolated fault per arm, built with the identical flags, must fail **with that arm's own diagnostic and not the other's**. A probe that trips both proves neither.

One detail from writing those probes that cost time and is not obvious: `-fsanitize=undefined` includes an object-size check that reports a heap overflow *before* AddressSanitizer sees it whenever it can trace the allocation, so the address probe fails for the wrong arm. Making the size and the index `volatile` is not enough — the check still resolves the object through the pointer. A `volatile` **pointer** is what works, because the allocation's provenance stops being visible to that pass:

```c
volatile size_t size = 8;
char *volatile block = malloc(size);
```

Full text of all three, in a form that dropped straight into the same file, is in the closed #1050 — take whatever is useful.

hjosugi:
All three taken, in `dbddfbef`. I re-measured each before implementing rather than porting it on trust, and one of them turned out to be a correction rather than an addition.

**3 caught a real error, and it was mine.** The probe already had a volatile pointer, so it was behaving correctly — but the comment above it said the volatile *size* was the load-bearing part. Measured with the gate's own flags:

```
volatile size, plain pointer  -> 0 heap-buffer-overflow, 1 UBSan report
plain size, volatile pointer  -> 2 heap-buffer-overflow, 0 UBSan reports
```

So the size is neither necessary nor sufficient and the pointer is both, exactly as you describe — the object-size check resolves the object through the pointer. The failure mode is specific: someone reading that comment and simplifying the `volatile` off the pointer while keeping it on the size would have reverted the probe to proving the wrong arm, and the comment would have told them that was fine. Both measurements now sit beside the probe, and the variable is renamed to `block` to match yours.

**1 is asserted now.** Confirmed the model first — `kofun_rt_alloc` at `S.c:114`, zero deallocation entry points. Your warning about the naive check is exact:

```
$ grep -c 'free(' bootstrap/selfhost/driver/S.c
2
$ grep -n 'free(' bootstrap/selfhost/driver/S.c
10588:        const char *k_n8073 = "        free(allocation);\n";
10730:        const char *k_n8213 = "        free(allocation);\n";
```

Both inside string literals A1 emits — the compiled output frees, the compiler does not. The entry-point loop went in as written, with the reasoning recorded so the next reader does not re-derive it.

**2 is in.** Both refusal paths now go through the sanitized binary. Measured before wiring: each exits 1, stderr empty, stdout byte-identical (invalid UTF-8) or matching the pinned golden (`corpus_reject`), and no C written. Worth noting the assertion is deliberately not "exited 0" — these refuse by design, and on both paths the diagnostic goes to stdout, which is what makes "stderr must be empty" a clean sanitizer-only signal. I left the unwritable-output case out of this set for that reason: its runtime error legitimately goes to stderr, so it would need a different assertion shape.

Instrumented and armed are now separate PASS lines, so neither can stand in for the other.

The gate reports 15 PASS lines on `dbddfbef`, exit 0 — the original 10, plus the two pinned refusals, plus one each for the allocation model, instrumentation, and arming. CI is re-running now; I will confirm when it lands.

Thanks for writing #1050 up rather than discarding it — the volatile-pointer detail is not something the failure message points at, and I would have shipped a comment that actively misled the next person.

---

Issue #618, opened 2026-07-23
Self-host profile: freeze the smallest honest compiler S and coverage manifest


hjosugi:
## Metadata

- State: complete
- Tracker role: curated child
- Kind: spec-design
- Area: Self Hosting / C11 backend
- Priority: P0
- Roadmap milestone: M1-bootstrap
- Size: L
- Parent: #32
- Historical baseline: #628; Unicode revision: #641; expression-lowering revision: #704; four-class evidence contract: #908 via #951
- Completion baseline: `f7c3cfa3f21632a6f0ce18fb94745922e2b47e31`

## Outcome

Freeze the smallest honest canonical compiler source `S` that can reach the first semantic self-hosting fixed point, together with an executable coverage contract that distinguishes:

1. a feature used by the exact bytes of `S`;
2. a feature accepted and typed by the compiler built from `S`;
3. a feature lowered by that compiler; and
4. a feature exercised by self-application.

B4/B5 allows deterministic C11 plus one declared normalized host C compiler. Direct x86-64/AArch64 compiler independence remains a separate track.

Delivered. The contract is `kofun.selfhost-profile/v2`, and each of the four classes names a prover that `bootstrap/selfhost/check-profile.sh` runs.

## Canonical source and profile revisions

There is one canonical source: `bootstrap/stage1/compiler.kofun`.

| Revision | Source SHA-256 | Rows | Commit | Change |
| --- | --- | ---: | --- | --- |
| `v0.3.1-seed` / #628 | `00aa2e97d2efdbbc32b8eef5163a2dae3793664bf5cf24f1ccf88648374b235d` | 43 | — | first explicit profile |
| Unicode / #641 | `40e00b91afc5303ed24e1bea74baeca4ed4a346f5022526a67233d984761c5ee` | 46 | `e5c0d83018ff35a7516fd37511f490b93b890850` | added `is_xid_continue`, `replace`, `validate_unicode_source` |
| Expression lowering / #704 | `cdcb3d682f1462e3067fee33b78d6e87efc534ebb917c7b3a07c7bd7c2c7a8cd` | 45 | `4d40f7e5c5315a06a1643ba80485648b8882f899` | removed builtin row `replace`; `S` parses and lowers expressions instead of emitting `evaluate("source")` |
[...]

PR #1050, opened 2026-08-06
selfhost: sanitize the A1 host boundary and pin its failure paths (#622)

hjosugi:
Closes the two remaining acceptance criteria on #622. One reviewable change to
`bootstrap/selfhost/check-compiler-driver.sh`, plus the `desc` line that
describes it.

## Why

A1 is the one place in the self-host chain where hand-written C touches the
operating system: argument decoding, `read_text`, transactional `write_text`,
and allocation. Everything the gate ran went through it at `-O2` with no
instrumentation, so a use-after-free or a signed overflow on that boundary
showed up as a green run. `grep -ril fsanitize bootstrap/ tests/ Taskfile.yml`
named 28 files before this change and not one was under `selfhost/`.

Two of A1's failure paths also behaved correctly while being asserted by
nothing, so a regression there was silent.

## What the gate now proves

**Sanitizer coverage.** A1 is built a second time with
`-fsanitize=address,undefined -fno-sanitize-recover=all`. The four corpora, the
`A1(S) -> C2` self-compile, and both refusal paths run through that binary and
must produce empty stderr and output byte-identical to the uninstrumented run —
instrumentation must not change what A1 emits.

Two claims are made separately, because neither substitutes for the other:

1. **A1 is instrumented** — `__asan_` must be present in the binary that ran.
2. **The arms report** — one isolated fault per arm, built with the identical
   flags, must fail with that arm's own diagnostic and *not* the other's. A
   probe that tripped both would prove neither.

Without (2) a clean run cannot be distinguished from a sanitizer that was never
linked, and UBSan without `-fno-sanitize-recover` prints its finding and still
exits 0. Without (1), (2) would only prove the toolchain works. The discipline
is copied from `tests/interop/bindgen-c/check-sanitizers.sh`.

**`detect_leaks=0` is asserted, not assumed.** A1's runtime never frees:
`kofun_rt_alloc` is a `malloc` wrapper and the prelude defines no deallocation
entry point, so every Text lives until the process exits. That is the allocation
model for a compiler that runs once, not an oversight — LeakSanitizer would
report the model rather than a defect, and `detect_leaks=1` makes this gate
report a five-figure leak on a clean run. The gate fails if a deallocation entry
point ever appears in `S.c`, so the decision gets revisited instead of quietly
outliving its reason. Use-after-free and buffer overflow stay armed.

**The two ungated failure paths, pinned.**

- an unwritable output destination exits 1 with
  `Kofun runtime error: cannot open output file`, writes no compile diagnostic,
  and does not create the missing directory;
- invalid UTF-8 input is refused with `EUNICODE001` at its exact line, column
  and byte offset on **stdout**, with empty stderr and no C written. The input
  is written to a temporary rather than committed — a `.kofun` file holding
  invalid UTF-8 would be read by every gate that walks the repository corpus.

Read the second one together with `tests/diagnostics/registry.tsv`, which
registers `EUNICODE001` once, for the native backend, on *stderr*. A1 emits the
same code from a different emitter on *stdout*, so the registry describes a path
this one only resembles. The corpus already proved A1's Unicode validation is
linked; nothing proved it refuses.

## Validation

| Check | Command | Result |
| --- | --- | --- |
| Self-host driver | `sh bootstrap/selfhost/check-compiler-driver.sh` | exit 0, 14 PASS lines (10 existing + 4 new) |
| Profile | `task selfhost-profile` | exit 0, 46 complete rows against 10 reviewed corpora |
| Frontend | `sh bootstrap/selfhost/frontend/check-frontend.sh` | exit 0 |
| C11 slices | `sh bootstrap/selfhost/c11/check-c11.sh` | exit 0 |
| Silent assertions | `sh tests/assertions/check.sh` | exit 0, still 0 for this file |
| Full regression | `task verify` | exit 0 |

New PASS lines:

```
PASS: an unwritable output and invalid UTF-8 input each exit 1 with their exact bounded diagnostic and no artifact
PASS: the A1 host boundary runs the corpora, A1(S), and both refusal paths clean under ASan+UBSan
PASS: A1 still never frees, which is why leak detection is off and memory-error detection is not
PASS: isolated address and undefined-behavior faults still report, so the sanitizer arms are armed
```

## Scope

The canonical source and profile are untouched: `bootstrap/stage1/compiler.kofun`
stays at `3d6025f5a951…` with 46 complete rows. The structured refusal code and
span, which requires an `S` revision under the #618 procedure, remains split out
as #1032.

Refs #622.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -96,7 +96,7 @@ tasks:
       - "sh bootstrap/selfhost/c11/check-c11.sh"
 
   selfhost-self-compile:
-    desc: "Compile canonical S with A1 and verify deterministic strict-C11 C2"
+    desc: "Compile canonical S with A1, verify deterministic strict-C11 C2, and sanitize its host boundary"
     cmds:
       - "sh bootstrap/selfhost/check-compiler-driver.sh"
 
--- a/artifacts/release-evidence/index.json
+++ b/artifacts/release-evidence/index.json
@@ -653,7 +653,7 @@
     }
   ],
   "evidence_digests": {
-    "Taskfile.yml": "dc65cd6be63d8295d606916174576b4c5e05a7ae83b879fb258a417256277ced",
+    "Taskfile.yml": "c220964e77dac25f610c7a3a6f60db7c99a10ca5e1266ca9178a69b94e58cf85",
     "bootstrap/native/check.sh": "fea94cb623757e1bf17e4497f43d656b1e8ff57a39a92ece871d8ff257c0f2ef",
     "bootstrap/native/fixtures/function_all_traps_pressure.kofun": "8ecd1a2a6956db3812210a60b2f027bee406777db14b2faea3fc7008c280528a",
     "bootstrap/selfhost/driver/corpus_builtin_rejects.tsv": "b1fa0b26247581aef797a81880554a94233a34109215530641706d76777087d4",
--- a/bootstrap/selfhost/check-compiler-driver.sh
+++ b/bootstrap/selfhost/check-compiler-driver.sh
@@ -468,6 +468,57 @@ grep -F 'Kofun runtime error: cannot open input file' \
 printf 'previous output\n' | cmp - "$temporary/preserved.c" ||
     fail "a failed compile must preserve the previous output"
 
+# I/O failure, the other direction: an unwritable destination panics with the
+# same bounded runtime shape as a missing input. Nothing pinned this, so a
+# regression that started writing a partial file, or exited 0 after failing to
+# open the output, would have been silent.
+set +e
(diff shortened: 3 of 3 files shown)


hjosugi:
Closing in favour of #1049, which implements the same two #622 acceptance criteria and was opened first. I did not check for an existing PR on the issue before opening this one — only for one on my own branch — so this is a duplicate of work that was already in review.

The three assertions this branch carried that #1049 does not are written up as a review comment there: https://github.com/hjosugi/kofun/pull/1049#issuecomment-5209304192 — the never-free model pinned so `detect_leaks=0` cannot quietly outlive its reason, the two refusal paths also run instrumented, and the *instrumented* and *armed* claims kept separate. The branch `agent/622-a1-host-boundary` is left in place so that text can be lifted from it verbatim.

Issue #943, opened 2026-08-02
Lambda spelling drift: Stage 2 accepts `(a, b) => e` and `x => e`, the grammar and SYNTAX.md describe neither


hjosugi:
## Metadata

- State: needs-decision
- Tracker role: curated
- Kind: spec-design
- Area: Lexer and Parser
- Priority: P2
- Milestone: M1-bootstrap
- Size: S
- Audited baseline: `7cdecd68dd42504da5881c6253e8f3d091636064`

## Summary

Stage 2 accepts three anonymous-function spellings. `spec/grammar.ebnf` defines
one of them, `docs/SYNTAX.md` shows that same one, and `spec/syntax/call-arguments-v1.md`
states there is no second form. The executable gate is the one that is right
about the language today, so either the documents are behind or two spellings
exist that nobody meant to keep.

## Current behavior and evidence

Measured on `main` at `7cdecd68dd42504da5881c6253e8f3d091636064`. Build the
Stage 2 compiler with the `kofun_stage2_build` function in
`bootstrap/stage2/build.sh`, then:

```sh
printf 'fn main() -> Int {\n    let bare = value => value * 3\n    let two = (a, b) => a + b\n    print(bare(14))\n    print(two(40, 2))\n    return 0\n}\n' > lam.kofun
kofun-stage2 lam.kofun lam.c lam.ir lam.tokens
cc -std=c11 -O2 -w lam.c -o lam && ./lam
# expected current result:
# 42
# 42
```

Both forms are deliberate, not accidental acceptance. The gate scopes them by
name:

```sh
sed -n '546,563p' tests/conformance/syntax/issues_35_47/run.sh
# "Lambdas were one of this checkpoint's unsupported fixtures until #703 lifted
#  them to top-level functions."
# "... the two-parameter form, a capture of an enclosing binding, a lambda
#  calling another lambda, and the bare single-parameter form `x => e`."
sed -n '1,8p' tests/conformance/syntax/issues_35_47/lambda_binding.kofun
# let double = fn(value: Int) => value * 2
# let add = (left, right) => left + right
# let bare = value => value * 3
```

The three documents disagree with that:

```sh
sed -n '138,139p' spec/grammar.ebnf
# lambda           = "fn", "(", parameters, ")",
#                    ( "=>", expression | block ) ;

sed -n '66,79p' docs/SYNTAX.md
# shows only fn(x) => x + 1, fn(x: Int, y: Int) => x + y, and the fn(x: Int) { } block form

[...]

Issue #882, opened 2026-08-01
[Umbrella] Complete call-arguments v1 fixed-slot lowering across pipelines and backends


hjosugi:
## Metadata

- State: in-progress
- Tracker role: planning umbrella
- Kind: planning
- Area: Code Generation / Call Arguments
- Priority: P1
- Size: L
- Parent: #625
- Blocked by: #1396, #1397, #1398, #1399
- Audited against: `origin/main@833cc752923b910225108e7a0fe115632e45a25e`

## Outcome

Coordinate the remaining cross-backend work for call-arguments v1 while
preserving the fixed-slot, exactly-once implementation already delivered for
labelled calls, the accepted trailing lambda, and the bounded one-stage direct
Stage 2/C11 pipeline.

This issue is an umbrella. It does not own a combined compiler/backend change.

## Delivered foundation

- #880: accepted syntax/formatter surface.
- #881: HIR, type, binding, and KIF identity.
- #1097: first direct top-level fixed-slot C11 lowering.
- #1107/#1189: Text/List[Int], Int?, enum, record, and labelled `take` carriers with ordinary temporaries.
- #1113/#1200: shared fixed-slot lowering machinery.
- #1191/#1237: accepted expression-bodied trailing lambda, source/ABI order, E2S167 duplicate-final-slot refusal, and a held lifted-lambda-body boundary.
- #1190/#1226/#1227/#1228: one-stage direct-call pipeline recognition, slot-zero binding, arity/type/`take` checking, and shared fixed-slot C11 lowering.

On current main, the lasting C11 gate proves fixed declaration/ABI slots,
subject-first and exactly-once source-order evaluation, no label map/string
dispatch, and no extra allocation for those shapes. Bare/member targets,
pipeline chains, pipeline-plus-trailing-lambda forms, block-bodied trailing
calls, labelled calls inside lifted lambdas, and lexical/indirect targets retain
their exact Stage 2 refusal boundaries.

## Remaining child DAG

```text
#1190 production pipeline construct (completed)
  -> #1226 bind subject to slot zero (completed)
      -> #1227 arity/type/take checking (completed)
          -> #1228 C11 fixed-slot lowering (completed)
              -> #1192 direct-native/Wasm differential (completed)
```

| Issue | State now | Deliverable |
| --- | --- | --- |
| #1190 | completed | recognize one direct-call pipeline production and fail closed downstream |
| #1226 | completed | bind the subject to the first declaration slot |
[...]

Issue #881, opened 2026-08-01
Call arguments v1 slice 2: bind labels in HIR, type checking, and KIF identity


hjosugi:
## Metadata

- State: ready
- Tracker role: curated child
- Kind: implementation
- Area: Type system / Syntax / HIR / KIF
- Priority: P1
- Size: M
- Parent: #625
- Surface predecessor: #880
- Downstream lowering slice: #882
- Audited against: `origin/main@60d0af56f2e37a5ce4750544b005eb72f1b73154` (2026-08-08 JST)

## Goal

Implement the production semantic slice for call-arguments v1: accept
declaration external labels and ordinary labelled calls, bind them to fixed
HIR parameter slots only after callable selection, run the existing type,
ownership, and effect checks on those slots, and make the public KIF identity
include the external-label/unlabelled shape.

This slice ends at checked HIR and compiler-authoritative interface identity.
Pipeline/trailing-lambda and backend lowering remain #882.

## Current behavior and evidence

Measured on the audited commit in a clean detached worktree.

The accepted contract and parser/formatter predecessor are complete:

```sh
sh spec/syntax/call-arguments/check-surface.sh
# PASS: 8 declaration shapes; 22 accepted forms; 26 named refusals;
#       formatting, ambiguity, and decision-corpus checks all pass

sh spec/syntax/call-arguments/check.sh
# PASS: static label binding; source evaluation versus ABI order;
#       binding diagnostics, interface fingerprint, and no-map model
```

Production still stops before this semantic slice:

```sh
./bin/kofun check tests/diagnostics/stage2/e2s158_declaration_external_label.kofun
# exit 1
# error[E2S158]: a declaration external label is specified by call-arguments v1 but not implemented at byte 548
```

The standalone decision model in
`spec/syntax/call-arguments/model.mjs` already proves the intended algorithm,
but the production Stage 2 HIR/checker does not consume it. PR #1052 deliberately
left that boundary named rather than misparsing the external label as the
internal parameter name.

The compiler-authoritative KIF checkpoint is also intentionally narrower than
this issue. `KofunKifFact` currently stores function parameter count/types and
result type, but no external labels; `encode_signature()` therefore cannot
change public semantic identity after an external-label rename. The production
interface snapshot likewise has no label field, and currently rejects published
ownership-mode signatures as unsupported (`EKI02`). The focused KIF v1 gate is
green at this narrower boundary.

## Scope

1. Parse each production parameter as distinct ownership mode, external label
[...]

Issue #838, opened 2026-08-01
grep assertions fail without saying what they looked for: 397 of them


hjosugi:
## Metadata

- State: ready
- Tracker role: curated
- Kind: planning
- Area: Verification gates
- Priority: P2
- Roadmap milestone: M2-alpha
- Size: L

## Why this exists

#814 made every `test` assertion name itself. It did not touch the other half
of the same defect.

**397 grep assertions across 28 files abort with no output at all.**

```sh
grep -q 'error[E2S64]: canonical import cycle' "$WORK/cycle-64.log"
grep '^function|dwarf_debug_strings_for|1|' "$WORK/encoder.ir" >/dev/null
```

Both forms are assertions — the gate is checking that an artifact says
something — and both are silenced on purpose, `-q` by the flag and the other by
the redirect. Under `set -eu` a failing one exits the script and prints
nothing, which is exactly the failure mode #814 spent five children removing.
The count is larger than #814's 459 was in its worst file: `bootstrap/native/check.sh`
alone has 76.

| form | count |
|---|---:|
| `grep … >/dev/null` | 213 |
| `grep -q …` | 184 |
| **total** | **397** |

## What a fix looks like

`tests/assertions/assert.sh` gains two helpers that keep the caller's grep
flags verbatim, because `-F`, `-x`, `-E` and `--` all change what the pattern
means:

```sh
assert_grep     "canonical import cycle in cycle-64.log" -F 'error[E2S64]: …' "$log"
assert_not_grep "no private name in the public sidecar"  -F 'internal_' "$out"
```

On failure each prints the label and the grep it ran, so the reader sees the
pattern and the file without opening the script.

`tests/assertions/check.sh` counts the silenced forms the same way it counts
`test`, using the rules #814 and #836 settled: no `||`/`&&` handler, not an
`if` condition, and not the last command of a function body.

## Children

- [x] **#839** — the two helpers, the counter change, and `bootstrap/` (160).
- [x] **#840** — the conformance gates (139).
- [x] **#841** — frameworks, tooling, spec, and integration (98).

160 + 139 + 98 = 397, so the children partition the count rather than sampling
it.

## Acceptance criteria

- [x] `assert_grep` and `assert_not_grep` exist, keep the caller's flags, and
      print the label and the grep on failure.
- [x] The counter sees both silenced forms and holds them to the same per-file
      budget as `test`.
- [x] Every child is closed and the budget records 0 for every file.
[...]

PR #1052, opened 2026-08-06
syntax: parse call-argument labels and the canonical trailing lambda (#880)

### LINKED ISSUE #880: Call arguments v1 slice 1: parse labels and canonical trailing lambdas
## Metadata

- State: ready
- Tracker role: curated child
- Kind: implementation
- Area: Syntax
- Priority: P1
- Roadmap milestone: M0-spec
- Size: M
- Parent: #625
- Contract: [`spec/syntax/call-arguments-v1.md`](https://github.com/hjosugi/kofun/blob/main/spec/syntax/call-arguments-v1.md)

## Goal

Implement the surface layer selected by #625 — declaration external labels,
mandatory labelled call arguments, and the single existing-fn trailing-lambda
spelling — plus the canonical formatter and the ambiguity corpus. Do not bind
types or lower calls in this slice.

Re-measured on `main` at `a50793ea6802411fd6f7c752578fd9421669236f`.

The #625 decision has landed. `spec/syntax/call-arguments-v1.md` is 220 lines
and states `Status: accepted design contract for issue #625`, so the blocker
this issue previously named is resolved.

The surface itself is not implemented. Stage 2 refuses a labelled argument on
an ordinary function rather than accepting the accepted grammar:

~~~sh
$ cat lbl.kofun
fn identity(value: Int) -> Int { return value }
fn main() -> Int { print(identity(value: 1)) return 0 }

$ ./kofun-stage2 lbl.kofun out.c out.ir out.tokens
error[E2S32]: ordinary function `identity` does not accept labelled arguments at byte 73
~~~

That refusal is the correct current boundary — `spec/records-v1.md` relies on
labelled call form distinguishing record construction from an ordinary call, so
today a labelled argument to a function is rejected deliberately. This slice
changes what the parser accepts, which is why the last acceptance criterion
below matters: the profiles that cannot lower the new form must keep refusing
it rather than reinterpreting it.

### In scope

- declaration external labels;
- mandatory labelled call arguments;
- the single existing-fn trailing-lambda spelling;
- the canonical formatter for both;
- the ambiguity corpus for pipelines, nested calls, comments, newlines, and
  automatic termination.

### Out of scope

- type binding and call lowering for the new form;
- any change to record construction, which already owns the labelled call form
  under `spec/records-v1.md`;
- brace lambdas and implicit parameters, which the contract excludes.

## Acceptance criteria

- [ ] The parser accepts the exact normative grammar in
      `spec/syntax/call-arguments-v1.md` and rejects duplicate or unknown
      surface forms without guessing.
- [ ] The formatter emits parentheses, one space, then `fn`; it never invents a
      brace lambda or an implicit parameter.
- [ ] The pipeline, newline, comment, and nesting corpus matches the contract.
- [ ] Record construction is unaffected: `task records` stays green, including
      the `stage2_labelled_call` fixture that asserts an ordinary function
      still refuses labels where this slice does not apply.
- [ ] Compiler profiles that cannot lower the new form continue to reject it
      rather than reinterpret it, with a named diagnostic.
- [ ] A focused task is wired into `task verify`.
- [ ] Every new assertion names what failed, per #814 and #838.

## Validation

|---|---|---|
| Focused | the task added by this issue | the corpus matches the contract |
| Records boundary | `task records` | green, `stage2_labelled_call` unchanged |
| Falsification | remove one corpus case's expectation | the focused gate fails |
| Regression | `task verify` | no regression |

## Dependencies

- Blocked by: none. #625's decision landed as
  `spec/syntax/call-arguments-v1.md` with `Status: accepted`.
- Related: `spec/records-v1.md`, which owns the labelled call form for record
  construction and must not change here.

---

*Refined against `docs/ISSUE_READINESS.md`: the previous body opened with
"Blocked until the #625 decision lands on main", which stopped being true when
the contract landed, and carried no `State:` line to agree with the `ready`
label. Scope and acceptance are the original ones; only the form and the
measurement stamp are new.*

*Promoted to `ready` on 2026-08-02 after re-measuring at `a50793ea6802411fd6f7c752578fd9421669236f`: the `E2S32`
refusal reproduces byte for byte and the contract still reads
`Status: accepted`, so the `State:` line now agrees with the `Blocked by: none`
this body already carried, and the label follows both.*

hjosugi:
Implements slice 1 of [`spec/syntax/call-arguments-v1.md`](https://github.com/hjosugi/kofun/blob/main/spec/syntax/call-arguments-v1.md): the surface parser, the canonical formatter, and the ambiguity corpus. It binds nothing and lowers nothing — #881 and #882 own those, and the contract's unsupported-current-compiler boundary is not lifted.

## Why

The issue's last acceptance criterion asks that profiles which cannot lower the new form *reject* it rather than reinterpret it. They were reinterpreting it, and the two fixtures this change adds show it. Run against the compiler on the parent commit:

```
$ ./kofun-stage2-parent tests/diagnostics/stage2/e2s158_declaration_external_label.kofun ...
error[E2S35]: unknown lexical binding `base` at byte 601

$ ./kofun-stage2-parent tests/diagnostics/stage2/e2s158_trailing_lambda.kofun ...
error[E2S17]: Core function `apply` expects 2 arguments, got 1 at byte 655
```

`fn add(to base: Int, ...)` was read as `to: <type base>`, so `base` never bound and the author was told a name they *did* write, in the position the contract gives it, does not exist. The trailing lambda was worse: it fell outside the call entirely, so the compiler counted one argument and blamed the author for an arity they had not got wrong.

## 1. The surface layer

`spec/syntax/call-arguments/parser.mjs` is a tokenizer and recursive-descent parser for the normative grammar — every production in the contract's `## Grammar` block. `format.mjs` is the canonical printer. `surface-corpus.json` drives both, and `task call-arguments-surface` is wired into `task verify`.

### Where the parse layer touches a signature, and why only there

The contract requires exactly one such contact: *"The grammar never inserts a trailing lambda before overload resolution: the callee's single resolved signature must establish that the final parameter is functional."* So the parser takes a signature environment and reads that one fact out of it. When it cannot resolve a callee it refuses — `trailing-callee-unresolved` — rather than guessing. Everything else about binding, including whether a label is known or mandatory, stays with #881 and the existing `bindCall` in `model.mjs`.

The signature environment is itself evidence: it is built by parsing ordinary Kofun declarations with the parser under test, so if declaration external labels did not parse, no trailing case could run at all.

### The formatter's two prohibitions are the reason it exists

*"Block lambdas use the existing block formatter and are not rewritten into expression lambdas"* and the absence of a brace lambda are prohibitions, and a printer that violated one would do it silently on code that already parsed. There is no such spelling in the printer at all: `fn` is always followed by a parenthesised parameter list, a block body is always printed as a block, and the checks run over every canonical string the gate emits rather than case by case.

Canonical output is also asserted to be a **fixed point** — reparsed, reformatted, and required to be byte-identical with the same kind and the same trailing count. A formatter whose output parses to something else is not canonical, and run over a repository it would keep producing diffs.

## 2. The Stage 2 boundary: `E2S158`

Added to **both** profiles — `bootstrap/stage2/compiler.c` and `bootstrap/stage2/compiler.kofun` — because the acceptance criterion is plural. A refusal that existed only in the C profile would leave the self-hosted one still misreading the form.

It runs first in `parse_program`, before every other check, because it is the only one that can name the form the author actually wrote. Once an external label has been read as a type, every later diagnostic is about a program nobody wrote.

### The span is the external label, and getting that wrong once was instructive

The label is word one of `to base` and word two of `take into file` — the ownership mode shifts everything right. A first draft of this change pointed at the *internal* name in the two-word case, which would have repeated the exact defect the check exists to remove, one token over. Both spans are now pinned by fixture.

### Deliberately not covered: `edit token: Token`

An ownership mode and an internal name, with no external label. That two-word shape predates this contract and its boundary belongs to the ownership work (`E2S121` and the `edit_parameter` fixture). It still reports `E2S35`, unchanged — moving it here would move a line this issue does not own.

### Not covered either: `let d = fn(x) { ... }`

The block-body lambda pinned by `unsupported_block_lambda` is #943's boundary, not this one. `E2S158` fires on `) fn(` — the trailing form — so that fixture is untouched and still records its misparse.

## 3. A defect found in the sibling corpus

`spec/syntax/call-arguments/corpus.json` wrote `fn next(value: Int) => value`. `spec/grammar.ebnf` derives a declaration body as `( "=", expression | block )` — `=>` is the lambda arrow, and a declaration is not a lambda. The decision gate never noticed because its regex is `/\)\s+fn\s+[A-Za-z_]/`, which stops at the identifier.

Fixed to `=`, which preserves exactly what that entry tests. To stop the two gates drifting again, the surface gate now cross-checks **every** decision-corpus entry against the real parser and requires `classifySurface` to agree. No entry is skipped — a skip list is where a disagreement would go to hide.

## 4. What is checked

| Check | Command | Result |
|---|---|---|
| Focused | `task call-arguments-surface` | 7 PASS lines; 4 declaration shapes, 15 accepted forms, 14 refusals, 7 cross-checked decision cases, 2 width cases |
| Records boundary | `task records` | green; `stage2_labelled_call` diagnostic byte-identical |
| Falsification | drop a canonical, flip a nesting count, restore the `=>` body | each fails, naming what broke |
| Regression | `task verify` | green |

Every new assertion names what failed, per #814 and #838; the `assertions` gate is green.

## Notes for #881

`parser.mjs` exports `parseExpression` for a single expression with the whole input consumed, and `parseProgram`/`parseFragment` for declarations and statement sequences. The tree carries `mode`, `external`, `internal`, and the resolved `type` per parameter, and `call.trailing` plus `call.signature` where a lambda attached — which is what binding will need.

Closes #880.


hjosugi:
## Converting to draft: an adversarial review found real defects

Two are high severity and I have reproduced the first myself.

### 1. E2S158 misses every keyword-spelled external label — including `in`, the contract's own headline example

The parameter scan counts a word only when `token_kind` returns `"identifier"`, but Stage 2's lexer classifies 16 words as `"keyword"` — including **`in`**. So `in text:` counts as one word, no refusal fires, and the misparse this PR exists to remove reproduces verbatim:

```
$ cat in-label.kofun
fn replace(in text: Int) -> Int {
    return text
}
...
$ ./kofun-stage2 in-label.kofun o.c o.ir o.tok
error[E2S35]: unknown lexical binding `text` at byte 45
```

That is byte-for-byte the failure mode this PR's own header comment claims to have eliminated. Meanwhile `parser.mjs` reserves only `fn`/`let`/`return`, so it happily reports `{"external": "in", "internal": "text"}` for the same signature.

**The two halves of this PR disagree about the spec's flagship example** — `fn replace(in text: Text, from old: Text, to replacement: Text)`, which `surface-corpus.json` itself declares. `compiler.kofun` has the identical gap, so the profiles are in parity on the bug. Even when the three-parameter form does fire, it blames `from` at byte 25 rather than `in` at byte 11.

### 2. The formatter silently changes the meaning of a higher-order parameter type

```
source   : fn f(g: (Int -> Int) -> Int) -> Int
formatted: fn f(g: Int -> Int -> Int) -> Int
```

`parseType` discards the `Group` node and `format.mjs` never re-parenthesises a function type in domain position. `spec/grammar.ebnf` derives this via `primary_type = "(", type_ref, ")"`. The output is a **stable** fixed point, which is exactly why the idempotence assertion cannot see it — it round-trips to a different type and then stays there.

### Also found

3. The formatter throws an uncaught `FormatError` — not a `SurfaceError` — on any block lambda outside the trailing position, so the gate would die with a stack trace rather than a named assertion.
4. `parseType` cannot parse `() -> Int` or `(Int, Int) -> Int`, both normative in `spec/grammar.ebnf` and both live in the repo today. The corpus had to mis-declare `fold` with a **unary** combiner to work around it, while its headline case passes a two-parameter lambda.
5. Two cases violate the contract's "Otherwise `fn` begins the next expression" clause: the second-trailing-lambda guard checks neither `newlineBefore` nor `commentBefore`, and an unresolved callee followed by a lambda statement hard-errors.
6. `read` written explicitly is deleted by the formatter, and the comment claiming otherwise is false.

### Verified sound

No false positives on any checked-in source — every `.kofun` in the repo compiles identically under main and this branch except the two new fixtures. Depth tracking, string/comment safety, and `.c`/`.kofun` parity are all exact. Idempotence held across 340 generated forms at three widths.

Not merging as-is.

hjosugi:
## All six findings fixed; `task verify` exit 0

| Finding | Fix |
|---|---|
| `in` and 15 other keywords could not be external labels | both profiles count keyword tokens as parameter words |
| `(Int -> Int) -> Int` printed as a **different type** | function types keep their parentheses in domain position |
| `() -> Int` / `(Int, Int) -> Int` unparseable | tuple domains in `parseType` |
| uncaught `FormatError` on non-trailing block lambdas | arguments format multi-line instead of via `inline` |
| second-lambda guard ignored the newline rule | honours the same trivia clause its sibling does |
| explicit `read` deleted by the formatter | parser records `modeWritten` |

### The contract's headline example now refuses correctly

```
fn replace(in text: Int, from old: Int, to replacement: Int) -> Int
  before: error[E2S35]: unknown lexical binding `text`
  after : error[E2S158] at byte 994 — the FIRST label, `in`
```

Pinned as `tests/diagnostics/stage2/e2s158_keyword_external_label.kofun`. Before the fix the three-parameter form did eventually fire, but on `from` — the first label spelled as an ordinary identifier — so it pointed past the one the author would fix.

### The type bug got the assertion that catches it

Text idempotence cannot see a *stable wrong* fixed point: `Int -> Int -> Int` reparses to a different tree and then formats to itself forever. The gate now compares the **reparsed tree** to the original, and a new corpus case exercises a function type in domain position. Reverting the fix turns it red.

### Correcting the corpus revealed it had been bent around the bug

`fold` declared `combine: Int -> Int` — a *unary* combiner — while its own headline accept case passes `fn(acc, item) => acc + item`. It now declares `(Int, Int) -> Int`.

### Regression evidence

Swept **2923 `.kofun` sources** against a compiler built from `main`: exactly 2 behavioural differences, both this branch's own fixtures. Zero false positives from the widened detection.

Gate went from 4 declaration shapes / 15 accept cases to **6 / 17**. Fixing my own brace-lambda prohibition also caught a false positive in it — it flagged a legitimate argument-position block lambda.

Ready for re-review.

hjosugi:
## Third review round: five more defects, all fixed. `task verify` exit 0.

The most severe was this work **reintroducing the bug its own first round fixed**.

Round one fixed a formatter that dropped parentheses. The fix was a special case — parenthesise the domain when it is a function. Round two then added `optional` and `moded` nodes, which are two more ways for a function type to sit inside a node whose kind is not `function`, and the same class of bug came straight back:

```
(Int?)?                   ->  Int??                    does not reparse
(Int -> Int)?             ->  Int -> Int?              different type
take (Int -> Int) -> Int  ->  take Int -> Int -> Int   different type
```

**29% of generated well-formed types**, and not cosmetic. `typeIsFunctional` decides trailing-lambda attachment, so `op: (Int -> Int)?` — which *refuses* a trailing lambda — printed as `op: Int -> Int?`, which *accepts* one. Running the formatter over a declaration changed what its callers were allowed to write.

Type printing is now precedence-driven: each kind declares how tightly it binds, and a child is parenthesised when it binds looser than its position allows. Exhaustively over 137 shapes at depth two, plus 6000 generated: **zero drifts, zero unreparsable**.

### The same mistake, twice, in a second place

`multiline()` — round two added `Group`, `Binary`, `Member` and missed `Lambda` with an *expression* body, so an expression lambda whose body held a block lambda still raised `FormatError`. Both bodies handled now; every block-lambda position × 3 widths formats and reparses.

### Three grammar boundaries were wrong in the permissive direction

| Now refused | Why |
|---|---|
| `take Int`, `Int -> take Int`, `List[take Int]` | a mode belongs to `callable_domain` only |
| `(Int, Int)`, `(Int, Int)?`, `()?` | a parenthesised parameter list is a domain: it must be followed by `->` and cannot be optional |
| `(Int,)` | not derivable at arity below two, and had no printable form |

Still accepted: `take Int -> Int`, `(read Int, Int) -> Int`, `(Int?)?`, `List[(Int -> Int)?]`, `(Int, Int, Int,) -> Int`.

### And one where knowing less gave the stricter parse

`trailing-callee-unresolved` fired regardless of trivia while the check four lines below honoured a newline. So `consume(1)` + newline + `fn(y) => y` was two statements, but `mystery()` in the same shape was refused.

### Corpus

4 declaration shapes / 15 accept / 14 reject → **8 / 21 / 21**.

### What I would want a reviewer to know

Fifteen defects across three independent adversarial reviews; three found by my own review. **Each review found something wrong with the previous round's fix.** The pattern was not about this language — it was that I kept fixing by enumerating cases, and enumeration does not survive new shapes. The precedence table is the only fix so far that generalised.

A fourth pass is running against this state. I would not merge before it reports.

Issue #303, opened 2026-07-20
Decide PackageId, ModuleId, and semantic interface digests


hjosugi:
## Metadata

- State: complete
- Tracker role: curated
- Kind: spec-design
- Area: Modules
- Priority: P0
- Roadmap milestone: M1-bootstrap
- Size: M
- Generated range: KOFUN-01981–KOFUN-02000
- Active rows: KOFUN-01981 0c933006a77f; KOFUN-01982 077b4b4f3a29; KOFUN-01983 02f848066109; KOFUN-01984 2a8bb47c69d9; KOFUN-01985 9b30a35b8518
- Decision owner: repository maintainer

## Summary

Define stable package/module/symbol identities and three distinct digests:

1. public semantic digest for dependent packages;
2. package-internal semantic digest for other modules in the same package;
3. target ABI digest for linker/backend compatibility.

Do not use one hash for all three jobs. A private body edit should not rebuild dependent packages; an internal signature edit should invalidate same-package consumers but not external packages; a target layout change should invalidate target artifacts without changing target-neutral semantic identity.

## Current boundary

The project build has deterministic target/action caching through Frost, but no language-level compiled interface or semantic module graph. #575 recommends separate compiled-interface and typed-sidecar artifacts. #285 defines public/internal/private levels. #284 and #300 fix PackageId, FileId, and ModuleId inputs. #290/#589 fixes stable value, type, module, and meta namespaces.

## Proposed identity hierarchy

- PackageId: accepted schema from #284.
- ModuleId: hash(identity schema, PackageId, normalized semantic module path).
- NamespaceId: accepted #290 schema with stable value, type, module, and meta tags.
- SymbolId: hash(symbol schema, ModuleId, NamespaceId, declared name, declaration-kind discriminator).
- TypeId: SymbolId of the nominal declaration plus accepted generic arity/identity fields.
- ExportBindingId: hash(export schema, exporting ModuleId, exported local name, target SymbolId).
- ImplementationId: package/trait/type/declaration identity accepted by #403.
- FileId and source spans are provenance, not nominal identity.

No ID contains an absolute path, source discovery order, timestamp, inode, compiler process ID, or target machine.

### Public semantic digest

Hashes only facts required to type-check an external dependent package:

- schema, language edition, and compatibility version;
- public modules, names, namespaces, SymbolIds, canonical types, generic constraints, traits/implementations visible externally, effects, ownership modes, and public constants that affect typing;
- re-export edges and canonical target identities;
[...]

PR #1053, opened 2026-08-07
lsp: filter completion and navigation by caller context (#1033)

### LINKED ISSUE #1033: LSP visibility: filter completion and navigation by caller context
## Metadata

- State: ready
- Tracker role: implementation child
- Kind: implementation
- Area: Tooling / LSP
- Priority: P2
- Milestone: M3-beta
- Size: M
- Parent: #584
- Audited baseline: `f7c3cfa3f21632a6f0ce18fb94745922e2b47e31`

## Goal

Make completion, definition, hover, and references apply the compiler's accepted visibility decision using the requesting file/package as caller context.

## Current behavior

`tooling/lsp/server.js` `visibleSymbols()` only applies single-document lexical scope and shadowing. It carries no caller `PackageId` and performs no dependency visibility filtering. The existing `visibility-restricted` sidecar reason is a diagnostic token, not an access check. No `tests/lsp/visibility.sh` exists.

## Scope

- add caller file/module/package context to the bounded LSP symbol query;
- hide dependency-internal/private symbols from completion and navigation;
- keep same-file private symbols available in that file's editor session;
- return a bounded non-disclosing result for inaccessible targets, without private path/content leakage;
- keep output deterministic under source order and path remap.

## Non-goals

Compiler resolver rules, documentation-index filtering, sidecar storage authority, macros/generated declarations, or visibility fuzzing.

## Acceptance criteria

- [ ] completion hides dependency internals and still shows legitimate same-file private symbols;
- [ ] definition/hover/references cannot disclose inaccessible paths, spans, content, or candidate names;
- [ ] caller context cannot be supplied by an untrusted sidecar field;
- [ ] path-remapped and source-reordered workspaces produce identical public identities/results;
- [ ] focused and existing LSP gates plus full regression pass.

## Validation

- `sh tests/lsp/visibility.sh`
- `sh tests/lsp/check.sh`
- `task documentation-index`
- `task repository-check`
- `task verify`

## Dependencies

- Satisfied: #582, #583, #601, #604, #825.
- Blocked by: none.

hjosugi:
Makes completion and definition apply the compiler's accepted visibility decision, using the requesting document as caller context.

## Why

Measured against the real `tooling/lsp/kofun-lsp` on the parent commit, with one file declaring all three modifiers and the caret on an empty prefix:

```
completion offered 32 items:
  secret_helper   detail="Int -> Int"     # private
  package_only    detail="Int -> Int"     # internal
  public_api      detail="Int -> Int"     # pub
```

All three identically. `buildIndex` recorded no modifier, `visibleSymbols` carried no caller identity, and no layer of the server had a notion of `pub`, `internal`, or `private`. The issue's own summary is confirmed.

**Worth stating plainly, because it is a user-visible change beyond filtering:** the server was single-document, so no declaration from another file could appear anywhere, and a filter alone would have been a no-op. Satisfying "completion hides dependency internals *and still shows legitimate same-file private symbols*" requires that foreign declarations be reachable in the first place. So this PR adds a bounded cross-file symbol query and applies the access rule to it. Completion and definition now reach other open files — and only what the caller may see.

The change is conservative by construction: an omitted modifier is private, so an existing workspace of unannotated files gains nothing across file boundaries.

## 1. The decision

`tooling/lsp/visibility.js` is a pure function over a declaration's identity and a caller context. It reads no document, no sidecar, and no filesystem, which is what lets the gate check the rule directly rather than inferring it from editor behaviour.

The clause that makes the two halves of the criterion one rule rather than two competing ones is `spec/modules/visibility.md`'s own sentence: **"a top-level declaration is visible only within its source file"**. So `private` is per *file*. Same file sees everything it declares; `internal` crosses within a package; `pub` crosses packages.

Anything unrecognised resolves to `private`. `public`, `protected`, `pub(crate)`, and `pub(super)` are explicitly not aliases — the spec says so — so reading one as the modifier it resembles would widen an API on a spelling the language does not have. `pub(to path)` is in the table but recorded as follow-up work, so it fails closed too.

## 2. Caller context, and the two identities that were wrong

**Not `doc.fileId`.** It is assigned from `result.snapshot.fileId` — it comes out of the *sidecar*, which is analysis of the very file whose access is being decided. `spec/modules/visibility.md` closes that door by name: "Imports, generated declarations, macros, and sidecar readers cannot bypass access by constructing an ID directly."

**Not `doc.logicalPath` either**, which is what this PR first used and is wrong for a quieter reason. It is a *display* path and it is not injective: it falls back to the basename for a file outside the workspace root, and to the constant `editor-buffer.kofun` for an untitled buffer. Two such documents collapse to one identity, and the same-file clause then hands each of them the other's private declarations — the exact disclosure this change exists to prevent, through the field that looked most like an identity.

Reverting to it fails the gate that now covers it:

```
FAIL lsp visibility: two files outside the workspace root that share a basename
     must stay separate files; completion offered ["collide_private"]
```

Identity is the document **URI**: unique by construction (it is the key of the `documents` map), transport state the buffer's content cannot reach. Only equality is ever asked of it, so its absolute prefix never enters a decision and two checkouts at different paths still agree on what is offered.

## 3. Package identity fails closed

`spec/modules/package-roots.md` rule 2 is the only root rule an editor can apply — there is no source operand in a session — and rule 4 forbids the nearest-ancestor manifest search that would otherwise be the obvious implementation: "Kofun v1 performs no nearest-ancestor manifest search."

With no `kofun.toml` at the workspace root, every open file is an anonymous single-file package, which that document defines as **non-importable**. Nothing crosses, including `pub`. An editor that cannot identify the package must fail closed, because the failure mode of guessing is disclosure.

| declaration | same file | same package | no manifest |
|---|---|---|---|
| no modifier / `private` | visible | hidden | hidden |
| `internal` | visible | visible | hidden |
| `pub` | visible | visible | hidden |

## 4. Navigation refuses indistinguishably

Definition on an accessible declaration in another open file returns its location. On an inaccessible one it returns `null` — byte-identical to the answer for a name that exists nowhere. A distinguishable refusal would disclose that the name exists, which file holds it, and by elimination roughly where. The gate asserts the serialised refusal contains no path, no span, and no candidate name.

## 5. Determinism

Documents are ordered by display path and then by URI. The second key is not decoration: `logicalPath` is not injective, so a comparator without it returns 0 for two distinct documents and leaves their relative order to sort stability and the session's open order. The gate opens the same workspace at a second absolute path, and a third time with the files opened in the opposite order, and requires identical results.

## 6. One bug found in the work itself

`rank()` originally routed lattice levels through `declaredVisibility`, which collapsed `restricted` onto `private` and made `rank('private') < rank('restricted')` false. Reading a source modifier and ordering the visibility lattice are different questions — an unrecognised *spelling* must read as private, but `restricted` is a real rung — and they now have different functions.

## What is checked

| Check | Command | Result |
|---|---|---|
| Focused | `sh tests/lsp/visibility.sh` | decision function + 6 end-to-end scenarios |
| Existing LSP | `task lsp` | green; protocol test unchanged |
| Performance | same gate | definition p95 0.51ms, completion p95 cold 4.71ms / warm 2.52ms |
| Regression | `task verify` | green |

The focused gate runs inside `tests/lsp/check.sh`, so it is reached by `task lsp` and therefore by `task verify`, and it also runs standalone as the issue's validation line specifies.

## Deliberate boundaries

- **Hover and references stay single-document.** They disclose nothing today because they never look outside the buffer, which satisfies the non-disclosure criterion; extending them to reach accessible foreign declarations is a coherent follow-up, not a fix.
- **The workspace view is the session's open documents.** No directory walk and no filesystem read beyond the one `kofun.toml` stat, so a closed file cannot inject a symbol and the cost stays bounded by what the editor already has.
- **Package identity is presence-of-manifest, not a parsed `PackageIdPayload`.** `spec/modules/package-roots.md` leaves the digest to #303 and says the compiler does not expose `PackageId` yet, so this layer must not invent one.

Closes #1033.

### DIFF
--- a/tests/lsp/check.sh
+++ b/tests/lsp/check.sh
@@ -42,6 +42,7 @@ node --check "$ROOT/tests/lsp/performance_test.js"
 node --check "$ROOT/tests/lsp/bridge_fallback_test.js"
 node --expose-gc "$ROOT/tests/lsp/semantic_sidecar_test.mjs"
 node "$ROOT/tests/lsp/protocol_test.js" "$SERVER"
+sh "$ROOT/tests/lsp/visibility.sh"
 KOFUN_LSP_REVISION="$REVISION" \
     node "$ROOT/tests/lsp/performance_test.js" "$SERVER" "$RESULTS"
 
--- a/tests/lsp/visibility.sh
+++ b/tests/lsp/visibility.sh
@@ -1,0 +1,87 @@
+#!/bin/sh
+set -eu
+
+ROOT=$(CDPATH= cd -- "$(dirname -- "$0")/../.." && pwd)
+SERVER="$ROOT/tooling/lsp/kofun-lsp"
+
+fail() {
+    printf '%s\n' "lsp visibility: $*" >&2
+    exit 1
+}
+
+command -v node >/dev/null 2>&1 ||
+    fail "node is required to drive the language server"
+test -x "$SERVER" ||
+    fail "the language server is not executable: $SERVER"
+
+node --check "$ROOT/tooling/lsp/visibility.js" ||
+    fail "tooling/lsp/visibility.js is not valid JavaScript"
+node --check "$ROOT/tests/lsp/visibility_test.js" ||
+    fail "tests/lsp/visibility_test.js is not valid JavaScript"
+
+# The decision function on its own, before any transport is involved. A rule
+# this small is worth checking directly: every server-level assertion below
+# depends on it, and a failure here says "the rule is wrong" rather than "the
+# editor answered oddly".
+node -e '
+const assert = require("assert");
+const { accessible, declaredVisibility, rank } =
+    require("'"$ROOT"'/tooling/lsp/visibility.js");
+
+const here = { fileId: "a.kofun", packageId: "p" };
+const sameFile = (visibility) =>
+    ({ fileId: "a.kofun", packageId: "p", visibility });
+const otherFile = (visibility) =>
+    ({ fileId: "b.kofun", packageId: "p", visibility });
+const otherPackage = (visibility) =>
+    ({ fileId: "b.kofun", packageId: "q", visibility });
+
+for (const visibility of ["private", "internal", "pub", undefined]) {
+    assert.ok(accessible(sameFile(visibility), here),
+        `same file must reach its own ${visibility} declaration`);
+}
+assert.ok(!accessible(otherFile("private"), here),
+    "another file`s private declaration must not be reachable");
+assert.ok(!accessible(otherFile(undefined), here),
+    "an omitted modifier is private, so it must not cross a file boundary");
+assert.ok(accessible(otherFile("internal"), here),
+    "internal must reach another file in the same package");
+assert.ok(!accessible(otherPackage("internal"), here),
+    "internal must not cross a package boundary");
+assert.ok(accessible(otherPackage("pub"), here),
+    "pub must cross a package boundary");
+assert.ok(!accessible({ fileId: "b.kofun", packageId: null, visibility: "pub" }, here),
+    "an anonymous single-file package is non-importable, so even pub must not cross");
+assert.ok(!accessible(otherFile("pub"), { fileId: "a.kofun", packageId: null }),
+    "a caller with no package identity must reach nothing outside its own file");
+
+// spec/modules/visibility.md: these are explicitly not aliases. Reading one as
+// the modifier it resembles would widen an API on a spelling the language does
+// not have.
+for (const spelling of ["public", "protected", "pub(crate)", "pub(super)", "PUB"]) {
+    assert.strictEqual(declaredVisibility(spelling), "private",
+        `${spelling} is not an alias and must resolve to private`);
+    assert.ok(!accessible(otherFile(spelling), here),
+        `${spelling} must not grant access across a file boundary`);
+}
+
+// pub(to path). The table lists it; the same document records restricted
+// visibility as follow-up work, so it must not be widened here.
+assert.ok(!accessible(otherFile("restricted"), here),
+    "restricted visibility is undecided, so it must fail closed outside its file");
+
+assert.ok(rank("private") < rank("restricted"), "private ranks below restricted");
+assert.ok(rank("restricted") < rank("internal"), "restricted ranks below internal");
+assert.ok(rank("internal") < rank("pub"), "internal ranks below pub");
+
+assert.ok(!accessible(null, here), "a missing declaration must not be reachable");
+assert.ok(!accessible(otherFile("pub"), null), "a missing caller must reach nothing");
+
+console.log("PASS visibility: the decision function matches spec/modules/visibility.md");
+' || fail "the visibility decision function disagrees with spec/modules/visibility.md"
+
+node "$ROOT/tests/lsp/visibility_test.js" "$SERVER" ||
+    fail "the language server does not apply the visibility decision to completion or navigation"
+
+printf '%s\n' \
+    "PASS: LSP completion and navigation filter by caller context"
--- a/tests/lsp/visibility_test.js
+++ b/tests/lsp/visibility_test.js
@@ -341,6 +341,53 @@ async function scenarioPackageIdentityIsPerDocument() {
   await session.stop();
 }
 
+// Go-to-definition on a declaration this file owns must stay put. The semantic
+// adapter answers null on a declaration site by design, and the cross-file
+// fallback then fired on the bare identifier — so the caret on a parameter's
+// own name jumped to an unrelated file that happened to declare the same word.
+async function scenarioDefinitionOnDeclarationStaysPut() {
+  const root = makeWorkspace(true);
+  const session = new Session(root);
+  await session.start();
+  const dep = 'pub fn shared(value: Int) -> Int {\n    return value\n}\n';
+  const caller = 'fn probe(shared: Int) -> Int {\n    return shared\n}\n';
+  const depUri = await session.open('demo/dep.kofun', dep);
+  const callerUri = await session.open('demo/caller.kofun', caller);
+
+  const onDeclaration = await session.request('textDocument/definition',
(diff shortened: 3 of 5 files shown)


hjosugi:
## Converting to draft: an adversarial review found real disclosure defects

I ran an adversarial security review over this branch and it found leaks I missed. Two are confirmed reproductions against the real server over stdio, and I have verified the first myself independently.

### 1. A modifier is read from a non-modifier position — private declarations leak

`declaredVisibilityBefore` takes the token immediately before a declaration at the same depth and treats it as a visibility modifier, without checking it is in a modifier position. `spec/modules/visibility.md` warns about exactly this:

> `pub`, `internal`, and `private` are contextual declaration modifiers. The lexer continues to emit ordinary identifier tokens outside a position where a declaration or member may begin.

A file with **no modifier anywhere** leaks:

```kofun
module demo.internal

fn leaked(value: Int) -> Int {
    return value
}
```

```
offered: ["leaked"]
```

Renaming the last path segment to `core` closes it; writing `private` explicitly closes it. So the *default* way to write a private declaration is the vulnerable one, and `import demo.internal` reproduces it too.

### 2. Package identity is a session global, so `internal` restricts nothing

`callerContext` and `declarationContext` both return the module-level `workspacePackageId`, so `declaration.packageId === caller.packageId` is **always** true whenever a manifest exists. A neighbouring checkout with its own `kofun.toml` exchanges `internal` declarations in both directions, with full URI and span, and an untitled buffer — a non-importable anonymous package — both sees and is seen.

This also makes a claim in the PR description false. It says an anonymous package means "nothing crosses out of it and nothing crosses into it — not even `pub`". The decision function implements that; the operand fed to it does not.

### Also found

3. Effective visibility ignores imports, so completion offers names the server's own diagnostics reject in the same cycle with `error[E2S16]: unknown Core function`.
4. Foreign completions vanish while another document is analysing, returned with `isIncomplete: false` so the client caches the empty list.
5. The absolute checkout path decides which foreign declaration definition lands on, contradicting the determinism claim in the description.
6. Go-to-definition on a parameter's own declaration is hijacked to another file.
7. The feature silently no-ops for clients that send `workspaceFolders` rather than the deprecated `rootUri`.

### Root cause worth naming

`visibility.js` is correct — every one of its own assertions passes — and that is precisely why the gate stayed green while the server disclosed private declarations. The decision function was fed two badly sourced operands: visibility scraped positionally from the token stream, and a package identity that is a session constant rather than a per-document property.

Not merging this as-is. Fixing the operands and extending the gate to cover each reproduction.

hjosugi:
## Both disclosures fixed

### 1. A modifier must be in modifier position

A visibility modifier now has to be first on its line, which a path segment never is. Reverting the check reproduces the original leak across all three header shapes:

```
FAIL lsp visibility: a path segment spelled like a modifier must not make a file's
     unmarked declarations public; completion offered
     ["hidden_import","hidden_module","hidden_pub_segment"]
```

Control held: `internal fn marked_visible(...)` after a `module demo.core` header is still offered, so the fix does not close legitimate modifiers along with the false ones. The rule is deliberately conservative — two declarations crammed onto one line lose the modifier and fall back to private, erring toward hiding.

### 2. Package identity is per document

`documentPackageId` returns the workspace package only for documents under the selected root, and `null` — an anonymous, non-importable package — for anything else. Reverting it reproduces the cross-checkout leak:

```
FAIL lsp visibility: a neighbouring checkout with its own manifest is a different
     package, so its internal API must not cross; completion offered ["neighbour_internal"]
```

Both directions are now closed, verified against the review's own reproductions:

| Probe | Before | After |
|---|---|---|
| neighbouring checkout → workspace | `["vendor_internal"]` | `[]` |
| workspace → untitled scratch buffer | `["app_internal"]` | `[]` |
| unmarked file after `module demo.internal` | `["leaked"]` | `[]` |

### Gate

10 scenarios, up from 7. The existing seven still pass, so in-package `internal` and `pub` still cross and a file's own private declarations are still available to it.

### Still open from that review

Findings 3–9 are not addressed here: effective visibility ignores imports (so completion can offer a name the server's own diagnostics reject), foreign completions vanish during another document's analysis while `isIncomplete` says the list is complete, definition on a declaration site can be hijacked, the absolute path decides which foreign declaration definition lands on, `workspaceFolders`-only clients get no feature, and per-request cost is linear in open documents. Leaving the PR as a draft until those are triaged.

hjosugi:
## Fourth-round fixes: the navigation regression and two coherence defects

**Go-to-definition on a declaration this file owns jumped to another file.** A regression this branch introduced: the semantic adapter answers `null` on a declaration site by design, and the cross-file fallback then fired on the bare identifier, so a caret on a parameter's own name navigated to an unrelated file declaring the same word. The fallback is now skipped when the caret sits on a declaration the document holds.

**A list emptied by pending analysis claimed to be complete.** `reachableForeignSymbols` skips unsettled documents, so a keystroke in one split pane emptied the cross-file completions in the other — with `isIncomplete: false`, telling the client to cache the empty list and stop asking. The reply now carries the same incompleteness the truncation flag already did.

**The absolute-path tiebreak turned out to be already closed.** The `logicalPath` collision that made the absolute checkout path decide a definition target only happens *outside* the workspace root — and those files are now anonymous single-file packages under the per-document identity fix, so no colliding pair is reachable at all. Verified and pinned as a scenario rather than assumed.

### Gate

12 scenarios, up from 10. `task lsp` green including the performance budgets — definition p95 0.43ms, completion p95 cold 5.81ms / warm 2.25ms.

### Still open, and why they are not in this PR

| Finding | Status |
|---|---|
| effective visibility ignores imports | **needs its own issue** — completion can offer a name the server's own diagnostics reject with `E2S16`; fixing it means modelling `import`, which this PR does not do at all |
| `workspaceFolders`-only clients get no feature | fail-closed, so not a disclosure; a small follow-up |
| manifest deleted mid-session keeps sharing on | minor, and defensible under the "fixed for the invocation" reading |
| per-request cost linear in open documents | measured and bounded; the gate covers the single-document case only |

The first is the one I would not fold in here — it is a design gap, not a bug in this change.

Issue #1133, opened 2026-08-09
Type-level kinds v1: Nat, Symbol, and Bool as type-level data


hjosugi:
## Decision requested

Accept type-level data kinds — `Nat`, `Symbol`, and `Bool` — as the second
axis of the decision split recorded on #1130. The accepted profile in
[`spec/type-level-programming-v1.md`](https://github.com/kofun-lang/kofun/blob/main/spec/type-level-programming-v1.md)
is Type-only by design, and DD-031 records higher-kinded computation as
deferred, not refused. This issue owns the kind axis of that deferral:
literals of the three data kinds, erased `Nat`/`Symbol`/`Bool` parameters on
nominal type declarations, constant-step builtin type functions, and the two
inductive views (`Succ` on `Nat`, `Cons` on `Symbol`) that keep induction over
numbers and strings inside the structural termination discipline.

## Why this is its own decision

The analysis on #1130 measured that two of its three motivating examples are
blocked on kinds, not on termination: `Add[A: Nat, B: Nat]` is unspellable in
a Type-only profile at any fuel budget, and the same holds for every
symbol-consuming parser. Measured against the
[kofun-type-challenges](https://github.com/kofun-lang/kofun-type-challenges)
corpus, the kind axis gates most of the value people cite when they ask for
type-level programming: the typed router, `printf`, and the units library are
all kind-blocked first.

Under this issue alone — with #1130 undecided — those applications already
become expressible, because the views grant structural-subterm status and the
recursion they need is ordinary structural descent. What still needs #1130 is
the non-structural residue: fixpoint transformations and recursion whose
measure is a counter that is not a subterm.

## Design under review

[RFC-0009](https://github.com/kofun-lang/kofun/blob/main/rfcs/0009-type-level-kinds-v1.md)
is the proposed design for this issue. In one breath: three data kinds beside
`Type`; decimal `Nat` literals bounded to the checked 64-bit range; NFC
`Symbol` literals bounded to 65,536 UTF-8 bytes; `Bool` literals; nominal
declarations may take data-kind parameters as erased indices with equality
decided on reduced normal forms; builtin functions (`Add`, `Monus`, `Mul`,
`Div`, `Mod`, `NatEq`, `NatLt`, `NatLe`, `Concat`, `Length`, `SymbolEq`) at
one logical step each; kind-directed pattern namespaces with residual arm
semantics; every refusal a named diagnostic in the `E394`–`E399` band.

[...]

Issue #1130, opened 2026-08-09
Type-level programming v2: accept Turing-complete type computation, bounded by fuel rather than structure


hjosugi:
## Decision requested

Accept Turing-complete type-level computation as a language goal, superseding the row in [`spec/type-level-programming-v1.md`](spec/type-level-programming-v1.md) that reads:

| Turing-complete type programming | rejected as a language goal | conflicts with deterministic checking and bounded diagnostics |

That specification states that a change to the accepted forms, reduction limits, or trace contract requires a separately reviewed, versioned specification change. This issue is the request for that change: **type-level programming v2**.

## What v1 gives up, and why it matters

V1 selects named, kinded, **structurally terminating** type functions. Structural termination is what makes the profile decidable, and it is also what removes an entire class of programs:

- **type-level arithmetic** — `Sum`, `Multiply`, `Subtract`, `Fibonacci`, comparison of numeric literals. These need recursion whose measure is a counter, not a shrinking structure.
- **type-level parsing** — a route pattern, a `printf` format string, a JSON schema, a query string. The recursion consumes a string whose remainder is not a structural sub-term of the input type.
- **fixpoint transformations** — deep transformations that must iterate to a fixed point rather than descend once.

Measured against [kofun-lang/kofun-type-challenges](https://github.com/kofun-lang/kofun-type-challenges), which restates all 190 type-challenges problems, this is not a corner case: a large share of `medium`, most of `hard`, and effectively all of `extreme` need exactly what v1 rejects.

More importantly, it removes the applications people actually want a strong type system for. A typed router that derives parameter types from `"/users/:id/posts/:slug"`, a typed SQL builder that derives a row type from a query string, a units library that adds and subtracts exponents — all of these are type-level computation over unbounded structure.

## What acceptance costs, stated plainly

The v1 rejection was not arbitrary. Accepting this means giving up three properties, and the design has to say what replaces each:

| Lost | Replacement this issue proposes |
|---|---|
| Checking terminates | A **fuel budget**: a per-declaration reduction-step ceiling, with a diagnostic naming the declaration and the step count when exhausted. Non-termination becomes a bounded, reported error rather than a hang. |
| Diagnostics are bounded | A **trace depth cap** independent of the fuel budget, so an error at step 900,000 still prints a readable trace: the first N frames, the last N frames, and a count of what was elided. |
[...]

PR #1134, opened 2026-08-09
rfc: split the type-level v2 decision into a termination axis and a kind axis (#1130)

hjosugi:
Closes nothing. #1130 and #1133 are decisions, and this is the design each one decides on — two `proposed` RFCs, no amendment to anything accepted.

## Why two documents

#1130 asks to supersede v1's rejection of Turing-complete type computation with a fuel-bounded v2. Its own analysis then records that two of its three motivating examples — the typed router and the units library — are blocked on kinds rather than on termination, and recommends splitting the decision rather than answering it yes or no. Deciding only the termination axis would ship the cost without the headline benefit.

So the split is now on the record as two decisions with two designs. #1133 was opened for the kind axis.

| | Axis | Design | Delivers |
|---|---|---|---|
| #1130 | termination | [RFC-0008](https://github.com/kofun-lang/kofun/blob/agent/typelevel-v2-rfcs/rfcs/0008-type-level-general-v2.md) | fixpoints and non-subterm measures |
| #1133 | kinds | [RFC-0009](https://github.com/kofun-lang/kofun/blob/agent/typelevel-v2-rfcs/rfcs/0009-type-level-kinds-v1.md) | the router, `printf`, the units library |

Each is independently acceptable, and each says in its own opening what it does *not* deliver alone. RFC-0008 states plainly that accepted by itself it produces neither headline application.

## RFC-0008 — the termination axis

`type fn general` opts a declaration out of structural termination. A root that can statically reach one reduces under `kofun.type-reduction/general-v2`; everything else keeps the v1 profile verbatim, and a structural declaration may not call a general one, so the v1 guarantee stays compositional rather than becoming a comment.

The one place this departs from the languages usually cited: **the budget is versioned language semantics, not an implementation setting.** GHC's `-freduction-depth` and TypeScript's instantiation depth are both implementation-defined, which means the same source can check under one compiler and fail under another. That, and not general recursion, is where deterministic checking is actually lost — so the numbers belong to the named profile and change only with a new profile name.

It also carries the correction #1130 asked for. A fuel-bounded reducer *terminates*: every reduction ends in at most 2²⁰ steps. What v2 gives up is that a legitimate program can be refused for exhausting the budget — which is why the budget has to be declared and versioned, and is a different and more demanding cost than "checking might not terminate".

Two rules are semantics for the same reason, because both decide which code refuses a program: tail replacement, and a 64-state tail witness ring scoped to the frame being replaced. The ring exists because tail replacement would otherwise blind cycle detection — a two-declaration tail loop keeps one live frame — and it is frame-scoped rather than root-scoped so that a repeated *dead* state is not mistaken for a cycle. The resulting seam is stated rather than hidden: period 1 through 64 is `E393`, period 65 and above exhausts fuel and is `E390`, and both sides get a fixture.

## RFC-0009 — the kind axis

`Nat`, `Symbol`, and `Bool` beside `Type`; literals; erased data-kind indices on nominal declarations with equality on reduced normal forms; eleven constant-step builtins; and two kind-directed views, `Succ` and `Cons`, whose shrinking binder counts as a strict subterm — which is what puts induction over numbers and strings inside v1's existing structural discipline rather than needing fuel for it.

It needs a profile of its own, and for a measured reason rather than a preference: **v1's 32 active frames, not its 256 logical steps, are what a scalar-by-scalar descent reaches first.** V1 has no tail replacement, so a non-tail `Cons` descent costs a frame per scalar plus one for the base case and caps out at 31 scalars — `"/api/v1/organizations/:org/projects/:project"` is 44. Shipping the kinds without the budget would have made the summary's own claim false.

Two boundaries are drawn hard, because both are the difference between adding type-level data and accidentally proposing a dependent-type system. No term informs an index and no index constrains a term. And a builtin applied to a non-closed operand is a stuck term compared syntactically, so `Add[N, 1]` is not equal to `Add[1, N]` — anything else is a solver, which refinement retains its own owner for.

## What this does not do

It does not amend DD-031. The rejection stands as written; a supersession still requires the versioned specification change the spec describes, and taking either decision is what starts that. The spec, the ledger, and both documents are consistent about this being a proposal.

Trace v2 is defined jointly across the two documents over disjoint field sets, so either can be accepted alone and land the schema carrying only its own contribution. Neither touches the v1 trace: a `default-v1` root still emits it byte for byte.

## Two smaller corrections

`rfcs/README.md` said every recorded decision was `migrated`. That stopped being true when RFC-0007 landed, and this adds two more `rfc` rows. RFC-0007's diagnostic-band note claimed `E390`+ was unclaimed; it now names both new bands. RFC-0007 is still `proposed`, so it is editable — the immutability rule attaches at acceptance.

## Validation

| Check | Command | Result |
|---|---|---|
| Decision ledger | `task rfc-registry` | passes — 34 decisions, 3 proposed, 23 mutations refused |
| Trace contract | `task type-reduction-trace` | passes, including its assertion that documentation claims no active type-function implementation |
| Repository | `task verify` | exits 0 |

Compatibility for both proposals is `additive` and measured at `a654f7fe`. Four queries return `0`: no tracked Kofun source declares a `type fn`, annotates a type parameter with a data kind, or declares a type whose name RFC-0009 reserves.

The design itself was put through three adversarial review passes before this was opened. The first found that RFC-0009's headline claim was wrong about which v1 limit binds; the second found that the tail rule as first written defeated cycle detection for any period above 1; the third found that a root-scoped witness ring would refuse an honest re-computation. Each is fixed above, and each fix is pinned by a fixture in the document's Validation table so it cannot quietly regress into an implementation detail.

### DIFF
--- a/rfcs/0007-derive-v1.md
+++ b/rfcs/0007-derive-v1.md
@@ -173,8 +173,9 @@ same file beyond the two rules above.
 | `E383` | the same derive named twice on one type | the derive and both attribute spans |
 | `E384` | a generated declaration colliding with an existing one | both declarations and both spans, in canonical order, with the remedy being to remove one |
 
-`E385`–`E389` stay unallocated in this band. `E370`–`E372` are RFC-0005's and
-`E390`+ are unclaimed.
+`E385`–`E389` stay unallocated in this band. `E370`–`E372` are RFC-0005's,
+`E390`–`E393` are RFC-0008's, `E394`–`E401` are RFC-0009's, and `E402`+ are
+unclaimed.
 
 ## Ownership and effects
 
--- a/rfcs/0008-type-level-general-v2.md
+++ b/rfcs/0008-type-level-general-v2.md
@@ -1,0 +1,547 @@
+# RFC-0008: Type-level programming v2 — general type functions under a versioned fuel profile
+
+- Shepherd: hjosugi
+- Opened: 2026-08-09
+- Status: proposed
+
+Proposal for [#1130](https://github.com/kofun-lang/kofun/issues/1130). This
+document records target semantics only. No parser, reducer, or trace producer
+implements it, and the rejection row in
+[`spec/type-level-programming-v1.md`](../spec/type-level-programming-v1.md)
+stands until this proposal is decided.
+
+This is one half of the decision split recorded on #1130. This document owns
+the **termination axis**: whether recursion whose measure is not a structural
+subterm is admissible, and what bounds it. The **kind axis** — `Nat`,
+`Symbol`, and `Bool` as type-level data — is
+[RFC-0009](0009-type-level-kinds-v1.md), proposed for
+[#1133](https://github.com/kofun-lang/kofun/issues/1133). Stated plainly, as
+the #1130 analysis requires: **accepted alone, this proposal does not deliver
+the typed router or the units library.** Those are kind-blocked first. What
+this proposal alone delivers is the recursion class that no kind can:
+fixpoint transformations, and descent on a measure that shrinks without being
+a subterm.
+
+## Summary
+
+A `type fn` may opt out of structural termination by declaring
+`type fn general`. A general declaration may recurse on constructed
+arguments and may be mutually recursive with other general declarations. A
+root reduction that can reach a general declaration runs under a new named
+profile, `kofun.type-reduction/general-v2`, whose frame, step, and node
+limits are fixed by this document — versioned language semantics, not
+implementation settings. Exhausting a limit is a deterministic type error
+that names the declaration, the limit, the measured counts, and the top
+consumers. The unmarked `type fn` keeps every v1 rule and every v1 guarantee,
+checked exactly as before, and a structural declaration cannot reach a
+general one, so the old guarantee stays compositional rather than becoming a
+comment.
+
+## Motivation
+
+V1 selects structurally terminating type functions, and the audit merged as
+[#1131](https://github.com/kofun-lang/kofun/issues/1131) made the three
+surfaces describing that profile say what is decided, what is implemented
+(nothing), and what is under challenge (#1130). This document is the
+challenge's design.
+
+What structural termination removes is the recursion class where the
+decreasing measure is not a subterm: iterate-to-fixpoint transformations,
+`gcd`/division-style descent, worklist passes over nominal graphs, and —
+once RFC-0009 exists — every string or numeric algorithm whose remainder is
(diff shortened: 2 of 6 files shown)


hjosugi (commented):
設計レビューです。対象ゲート（`task rfc-registry`、`task type-reduction-trace`、`task verify`）は commit `1e4c8bccec56218156528436a9b942e6bbef91df` で通過しました。ただし、accept 前に仕様上のブロッカーが4点あります。

1. **RFC-0009 の 1,025 深度の accumulator rewrite は v1 の構造再帰規則に適合しません。** v1 は direct self-call の「全引数」を matched strict subterm と要求しますが、accumulator は構築値です。design validation の「accumulator-passing rewrite succeeds」を成立させるなら、designated decreasing parameter 等へ終了規律を変える必要があります。規律を変えないなら、この成功主張と 1,025 深度の acceptance case を外してください。tail replacement はフレーム消費を変えても、この受理条件自体は変えません。

2. **RFC-0008 と RFC-0009 が同じ `kofun.type-reduction-trace/v2` を別々に拡張しつつ、独立 accept を許しているため schema freeze が成立しません。** RFC は accept 後 immutable、reader は unknown field を拒否する契約です。一方だけ accept した後でもう一方の field を同じ v2 に追加すると、accepted schema の破壊的変更になります。両軸で使う完全な v2 schema を最初から凍結するか、axis ごとに別 version / composition rule を定義してください。

3. **literal 上限違反の診断が未定義です。** RFC-0009 は `Nat` と `Symbol` literal 自体に上限を置きますが、E397/E399 は builtin result overflow だけです。上限外 literal の named diagnostic、phase、required structured facts を追加してください。

4. **open operand で stuck になる builtin の会計が未定義です。** 「builtin application は one step / one node」と、open operand は reduction せず stuck の両方が書かれています。stuck 判定を試みた application が step/node/trace entry を消費するかを固定してください。ここが実装間で違うと profile budget と trace が観測可能にずれます。

非ブロッカーですが、用語も2点直すと明確です。

- 固定 fuel の下で受理可能な計算は全て有限なので、language goal は文字どおりの “Turing-complete” ではなく「固定 reduction budget 下の general recursion」と表現する方が、RFC 自身の意味論に一致します。
- 「no index constrains a term」は indexed type equality や typed router の主張まで否定して読めます。境界は「runtime term が index を決めない／index だけを根拠に runtime observation を導かない」程度に狭めるのがよいです。

以上は branch を変更せず、既存 claim 所有者の修正待ちとします。

Issue #274, opened 2026-07-20
[Umbrella] Close B6 with policy-defined independent clean-builder evidence


hjosugi:
## Metadata

- State: blocked
- Tracker role: implementation/evidence umbrella
- Kind: planning
- Area: Self Hosting / Supply-chain verification
- Priority: P1
- Roadmap milestone: M2-alpha
- Size: L
- Parent roadmap: #32
- Direct children: #1289, #1290, #1530, #1291, #1292
- Blocked by: #1530, #1291, #1292
- Audited against: `origin/main@20cee8101c2792e0725fb6dd2211d1b0bfcec500`

## Outcome

Close B6 with one policy-compliant independent external builder reproducing the
frozen B4/B5 chain, and bind that reviewed attestation through the
#1530-selected identity/closure topology into normal verification,
bootstrap/roadmap truth, and bounded release wording without making the
attestation stale in the act of publishing it.

This parent coordinates the child sequence. It receives no direct code and
cannot produce, authenticate, review, or promote the external attestation by
itself.

## Current behavior and evidence

Current `main@20cee8101c2792e0725fb6dd2211d1b0bfcec500` already contains the
producer-owned foundation: the frozen self-host profile, C2 == C3 and A2 == A3
fixed point, content-addressed declared inputs with offline reconstruction and
fail-closed drift checks, and B7 gcc/clang diversity. B7 deliberately remains
separate from B6 because both chains still share one producer checkout,
machine, libc, and kernel.

#1289 and #1290 are completed; the remaining children are open:

```sh
for n in 1289 1290 1530 1291 1292; do
  gh issue view "$n" --repo kofun-lang/kofun \
    --json number,state,stateReason \
    --jq '"#\(.number) \(.state) \(.stateReason)"'
done
# #1289 CLOSED COMPLETED
# #1290 CLOSED COMPLETED
# #1530 OPEN
# #1291 OPEN
# #1292 OPEN
```

The completed children delivered distinct foundations:

- #1289 closed through #1335 with the delegated offline reproduction runner,
  canonical `kofun.selfhost-b6-report/v1` report/validator, mutation corpus,
  and `task selfhost-b6-report`. The producer report is a structural self-check,
  not an independent attestation.
- #1290 closed after #1414 landed
  `bootstrap/selfhost/b6/POLICY.md`, `producer-identities.tsv`, the policy gate,
  and `task selfhost-b6-policy`. It selected an independent external party and
[...]

PR #1137, opened 2026-08-09
bootstrap: close B7 with a diverse-double-compilation gate (#1136)

### LINKED ISSUE #1136: Reproducible bootstrap B7: diverse double compilation of the fixed point
## Goal

Close roadmap **B7, diverse double compilation**, with an executable gate:
build the generation chain under two distinct host C compilers and require
that what the resulting Kofun compilers *emit* is byte-identical, even though
the executables are not.

## Scope

**In:** a gate that runs `bootstrap/selfhost/build-a1-a2.sh` under two declared,
distinct host C compilers; asserts C1 and C2 byte equality across them;
asserts the full driver corpus agrees across the two A2 compilers in emitted
C, stdout, stderr, and exit status; records path-independent provenance; and
carries a refusal corpus including a simulated Thompson payload.

**Out:** B6, independent clean-builder reproduction ([#274](https://github.com/kofun-lang/kofun/issues/274)).
Both chains run on one machine, one libc, and one kernel, from one checkout,
so a payload below the C compiler is shared by both and survives. This issue
does not narrow that gap and must not be read as doing so.

Also out: requiring the two toolchains to produce identical executables. That
is a reproducible-build claim about GCC and Clang, not about Kofun.

`bootstrap/manifest.json` at `a654f7fe79aa2c32c8d0f7bfb3c77c0ef3f52292` records B7 as the one open gate:

```sh
git grep -n 'diverse_double_compilation' -- bootstrap/manifest.json
# expected current result: "diverse_double_compilation": "open"
```

and `docs/ROADMAP.md` lists it as remaining bootstrap path item 2.

The host has two independent toolchains available, so the gap is tooling
rather than availability:

```sh
gcc --version | head -1    # gcc (GCC) 16.1.1 20260728
clang --version | head -1  # clang version 22.1.8
```

### Why the existing gates do not already cover this

Measured at `a654f7fe79aa2c32c8d0f7bfb3c77c0ef3f52292`. The existing chain gates pin artifacts against
checked-in evidence, and that evidence was recorded by one toolchain. If that
toolchain carried a payload when the evidence was recorded, the pins encode
the payload and every later run reproduces it and passes.

That is not hypothetical here — it reproduces. With a source-preserving
payload wrapper around `gcc` (it alters the compiler it builds and restores
the source it found, so every `SHA256SUMS` still verifies) and the checked-in
evidence regenerated from the payload chain:

| Gate | Result |
|---|---|
| `sh bootstrap/selfhost/build-a1-a2.sh` under the payload compiler | **exit 0**, every check green, including "every declared source, seed, profile, corpus, and runtime digest verified" |
| the same tree under two diverse toolchains | **exit 1**, `derived C1 differs from the checked-in evidence` at char 8675, naming `clang version 22.1.8` as the toolchain that disagrees |

The single-toolchain gate cannot see the payload because it never runs a
toolchain that did not produce the baseline. That is precisely the gap B7
names.

## Definition of done

- a gate that passes on the honest `gcc`/`clang` pair and is wired into
  `task verify`
- a refusal corpus proving the gate cannot pass by comparing a toolchain
  against itself, and that it refuses a host compiler carrying a payload
- `bootstrap/manifest.json` `diverse_double_compilation` reads `working`, and
  `truthful_status` states the B6 boundary that remains
- `docs/ROADMAP.md` and `bootstrap/README.md` agree with the manifest

## Metadata

- State: ready
- Size: M
- Blocked by: none

hjosugi:
Closes #1136.

`diverse_double_compilation` was the one gate `bootstrap/manifest.json` still recorded as `open`. It is now `working`, behind `task selfhost-diverse-double-compilation`, wired into `task verify` at ~65s.

## What the gate does

Builds the generation chain twice, under two host C compilers that are different binaries reporting different identities, and requires the two resulting Kofun compilers to **emit the same bytes**:

- C1 and C2 byte-identical across the toolchains
- all 45 driver corpus cases agreeing in emitted C, stdout, stderr, and exit status

On this host the pair is GCC 16.1.1 and Clang 22.1.8. The C is identical; the executables are not. **Executable equality is reported and never required** — requiring it would be a reproducible-build claim about GCC and Clang rather than anything about Kofun, and requiring inequality would fail the day two toolchains legitimately converge.

## Why the existing gates do not already cover this

This is the part worth reviewing closely, because if it were false the gate would be ceremony.

Every other chain gate compares the checkout against evidence checked into it, and that evidence was recorded by one toolchain. A payload present in that toolchain **when the evidence was recorded** is pinned along with it, and reproduces forever.

That is demonstrable, not theoretical. With a payload wrapper that alters the compiler it builds while restoring the source it found — so every `SHA256SUMS` still verifies — and the checked-in evidence regenerated from that chain:

| Gate | Result on the poisoned tree |
|---|---|
| `sh bootstrap/selfhost/build-a1-a2.sh` under the payload compiler | **exit 0**, every check green, including *"every declared source, seed, profile, corpus, and runtime digest verified"* |
| the same tree, two diverse toolchains | **exit 1** — `derived C1 differs from the checked-in evidence` at char 8675, naming `clang version 22.1.8` |

B7 is the only chain gate that runs a compiler which did not produce the baseline. That is the whole of what it buys.

## What it does not do

**It does not close B6 (#274).** Both chains run on one machine, against one libc and one kernel, from one checkout, so a payload below the C compiler is shared by both and survives untouched. B7 narrows the trusted set to what the two toolchains have in common; it does not empty it.

The gate says so in its own output, `bootstrap/README.md` has a section on it in both directions, and `bootstrap/manifest.json`'s `truthful_status` states it. A security gate that overstates its reach is worse than none.

## Refusals

Four, each pinning its diagnostic. Three are vacuity guards, because this gate has one specific way to become worthless — comparing a toolchain against itself would make every byte equality a comparison of a run with its own output, all green and all meaningless:

| Case | Refusal names |
|---|---|
| one compiler named twice | `resolve to the same binary` |
| two binaries reporting one identity | `report the same identity` |
| a compiler that is not installed | `KOFUN_DDC_CC_A and KOFUN_DDC_CC_B` |
| a host compiler carrying a Thompson payload | `did not reproduce the pinned evidence` |

The payload case is backed by the *second* toolchain rather than the honest one, because a payload wearing the honest compiler's identity is refused by the diversity guard before it compiles anything — which is that guard working, but would leave the payload case untested.

## Drift

`spec/roadmap-31-34/verify-current-gates.sh` asserted the manifest read `"open"` — the gate that caught this change and had to move with it. It now asserts `"working"` **and** that the two scripts behind that word exist, because a manifest claiming `working` with no gate behind it is exactly the drift that row is for. `docs/ROADMAP.md`, `spec/roadmap-31-34/README.md`, and `bootstrap/README.md` all move with the manifest.

## Validation

| Check | Command | Result |
|---|---|---|
| The gate | `task selfhost-diverse-double-compilation` | passes in 65s — C2 identical across toolchains, 45 corpus cases agree, 4 refusals fire |
| Roadmap gate | `sh spec/roadmap-31-34/verify-current-gates.sh` | passes |
| Repository | `task verify` | exits 0 |


hjosugi:
Added `docs/SECURITY.md`, which I had missed. Its §Bootstrap security ended with *"Diverse double compilation remains open in `bootstrap/manifest.json`"* — a sentence the rest of this branch makes false, in the one document where a stale sentence matters most.

It now states what the gate removes from the trusted computing base (a payload present in only one of the two host compilers) and, at more length, what it leaves behind:

- a payload in the **checked-in seed** — both chains build it from the same `bootstrap/stage2/compiler.c`, so it is shared and reproduces identically
- a payload **shared by both host compilers, or below them** in libc, the kernel, or the OS — both chains run on one machine
- **B6 (#274)** stays open, and is what would narrow the machine-shared part

`task verify` exits 0 with the change.

PR #1150, opened 2026-08-09
benchmark: run the native-function regression gate (#1139)

### LINKED ISSUE #1139: The native-functions regression gate is fully implemented and invoked by nothing
## Goal

Make the performance regression gate that already exists actually run.

Measured on `main@a654f7fe`.

`benchmarks/native-functions/benchmark.sh` implements a complete regression
gate. Its own header states the contract:

> With `BASELINE` set to a git revision, the same workloads are also built by
> that revision's `bootstrap/native/core_compiler.c` and the recorded budgets
> are enforced: every workload named in `IMPROVE` must reduce its median by
> the percentage declared there, no workload may regress by more than 5%, and
> neither corpus compile time nor emitted text size may regress by more than
> 10%.

The enforcement is real — `benchmark.sh:317` is `at_most "$regression" 5 || { … exit 1 }`.

**Nothing invokes it.**

```sh
grep -n "benchmarks/" Taskfile.yml          # expected current result: no matches, exit 1
grep -rn "benchmark" .github/workflows/     # expected current result: no matches, exit 1
```

And without `BASELINE` it disables itself by design (`benchmark.sh:273-275`):

```sh
test -n "$BASELINE" || {
    printf '%s\n' \
        "budgets=not evaluated (set BASELINE to a git revision to enforce them)"
    exit 0
}
```

`task benchmark-summary` is in `task verify` and is a false friend: it runs
`tests/stdlib/benchmark-summary/check.sh`, which gates the stdlib quantile and
MAD statistics library. It measures no compiler performance.

The recorded numbers are also stale, and were taken on a laptop:

```sh
git log -1 --format=%cI -- benchmarks/native-functions/results.json   # 2026-07-25
git log -1 --format=%cI -- benchmarks/http/results.json               # 2026-07-20
```

both `AMD Ryzen 3 7330U`, pinning `implementation_commit` values that are no
longer `HEAD`.

## Why this is the cheapest item on the 1.0 page

`docs/ROADMAP.md` §M4 lists "performance regression gates" as an exit
criterion. Exactly one performance assertion runs in CI today —
`tests/lsp/performance_test.js`, reached through `task roadmap` — and it
asserts **fixed absolute ceilings**, not a baseline:

| Metric | Budget |
|---|---|
| `diagnosticP95Ms` | ≤ 100 ms |
| `diagnosticMaxMs` | ≤ 250 ms |
| `definitionP95Ms` / `hoverP95Ms` / `completionColdP95Ms` / `completionWarmP95Ms` | ≤ 50 ms |
| `residentGrowthRatio` | < 0.10 |

A change that makes hover ten times slower — 4 ms to 40 ms — passes green.
That gate catches catastrophe, not regression.

So the repository has already built the machinery the exit criterion asks for
and is not running it. Every other M4 item needs work that does not exist yet;
this one needs wiring.

## Scope

**In:** a decision on how the baseline is supplied in CI, and the wiring that
follows from it. `BASELINE` today means "rebuild this git revision's compiler
and measure it", which doubles the work and needs a revision to name.

**Out:** new benchmarks, and the LSP gate's ceilings. Converting those to
ratio-vs-baseline is a separate change with the same baseline problem.

## The decision this needs

Naming a baseline revision in CI makes every run rebuild an old compiler.
Committing a results baseline makes the numbers machine-dependent, and shared
GitHub runners are noisy — the Taskfile already serialises the LSP gate for
exactly that reason. Neither is obviously right, and the choice determines
whether this is a `task` in `verify` or a scheduled lane.

## Metadata

- State: needs-decision
- Size: M
- Blocked by: none

hjosugi:
## Summary

- read the baseline pin from `results.json`, which already records `baseline_revision` and `budgets.declared_improvements`
- add `benchmarks/native-functions/check.sh` with a cheap `structure` mode and an enforcing `measure` mode
- put `native-functions-baseline` in `task verify`; give `native-functions-benchmark` its own scheduled lane
- register both in the task help groups

## Why

`benchmark.sh` implements a complete regression gate and **nothing invoked it**. Without `BASELINE` it disables itself by design, and no task or workflow set `BASELINE` — so `docs/ROADMAP.md` §M4's "performance regression gates" exit criterion was met by machinery that never ran.

## The decision

The issue asked which baseline CI names, and neither obvious answer is right on its own:

- **Committed numbers** measure the hardware. The recorded ones came from a Ryzen 3 7330U laptop; runners are neither that laptop nor each other.
- **A moving baseline** (`HEAD~1`) is worse than either: the 5% per-comparison allowance lets performance decay without limit while every run stays green.

So: a **named, pinned revision**, which is what the harness is already built for — it builds that revision's producer and measures it in the same run, on the same machine, interleaving rounds so drift lands on every variant.

And the pin is **not a new file**. `results.json` already records both halves, so the gate and the published numbers cannot describe different comparisons. Recording a new measurement is what moves the pin — which puts the move at the moment a person looked at the numbers.

## The split

Enforcing costs two producer builds and needs the baseline commit in the object store, so it does not belong in `verify`:

| task | where | does |
|---|---|---|
| `native-functions-baseline` | in `task verify` | proves the pin is coherent — real revision, claims naming measured workloads. No benchmark, no history. |
| `native-functions-benchmark` | `.github/workflows/benchmark.yml`, daily + dispatch, `fetch-depth: 0` | the real gate |

Not on `pull_request`, for the reason the backlog lane is not either: a timing gate that goes red for a reason a PR author cannot fix teaches people to ignore it. On `main` the same failure is a true signal.

## Validation

`task native-functions-baseline`, `task task-help`, `task documentation-index`, `task repository-check`, `task release-claims`, `task backlog`, `task assertions` — all pass.

Refusals verified by mutating `results.json`: short revision, claim naming an unmeasured workload, empty claim set, and non-positive percentage each exit 2 with a distinct diagnostic; the restored pin passes.

`task-help` initially failed on the two new tasks, which is the gate doing its job.

Closes #1139

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,50 @@
+name: Benchmark
+
+# The compiler performance regression gate (#1139). `docs/ROADMAP.md` §M4 lists
+# performance regression gates as a 1.0 exit criterion; the enforcement has
+# existed in `benchmarks/native-functions/benchmark.sh` since it was written and
+# nothing invoked it, so it enforced nothing.
+#
+# Its own lane rather than a job in CI, for two reasons that both come from what
+# the gate does. It builds the pinned baseline revision's producer in addition
+# to this revision's, so it costs about twice a normal build — too much to add
+# to every pull request for a signal that changes on almost none of them. And it
+# needs the baseline commit in the object store, which the default shallow
+# checkout does not provide.
+#
+# Not on `pull_request`, for the reason the backlog lane is not either: a shared
+# runner's neighbours are not this repository's business, and a timing gate that
+# goes red for a reason a pull request's author cannot fix teaches people to
+# ignore it. On a schedule the same failure is a true signal about `main`.
+
+on:
+  schedule:
+    # 04:00 UTC daily, off the hour that everything else on shared runners
+    # starts on.
+    - cron: "0 4 * * *"
+  workflow_dispatch:
+
+# A slow run is better than two interleaved ones: concurrent jobs would be
+# measuring each other.
+concurrency:
+  group: benchmark
+  cancel-in-progress: false
+
+jobs:
+  native-functions:
+    name: Native function budgets
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          # The gate builds the pinned baseline revision from source; a shallow
+          # checkout does not contain it, and the gate says so rather than
+          # reporting the absence as a regression.
+          fetch-depth: 0
+      - name: Install go-task
+        uses: arduino/setup-task@c0bc642852239c2689f73f4ea6459c29405f3c52 # v3.0.0
+        with:
+          version: 3.x
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Enforce the budgets against the pinned baseline
+        run: task native-functions-benchmark
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -488,6 +488,27 @@ tasks:
     cmds:
       - "sh tests/stdlib/benchmark-summary/check.sh"
 
+  # `benchmark-summary` above gates the stdlib quantile library and measures no
+  # compiler performance; these two are the compiler regression gate (#1139).
+  # They are split because the enforcing half costs two producer builds and
+  # needs full git history, which is a scheduled lane's budget, not `verify`'s.
+  #
+  # This half is `verify`'s: it proves the pinned comparison is still coherent —
+  # a real revision, and claims that name workloads the harness measures — which
+  # is what rots silently. It runs no benchmark and reads no history.
+  native-functions-baseline:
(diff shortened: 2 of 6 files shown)


hjosugi:
Closing in favour of #1154: the decision on #1139 went the other way.

Rather than wiring this gate into CI, the benchmark corpus moves out of the repository entirely — to [`kofun-lang/kofun-benchmarks`](https://github.com/kofun-lang/kofun-benchmarks), with full history — and `docs/ROADMAP.md` records the M4 exit criterion as owned there and **not met**.

This is no reflection on the work here. The baseline analysis in this PR is the best statement of the problem anyone has written down: committed numbers measure the hardware, a moving `HEAD~1` baseline lets performance decay without limit while every run stays green, and the pin belongs beside the published numbers rather than in a new file. That reasoning should carry over to `kofun-benchmarks`, where the harness now lives and where a scheduled lane costs this repository's `task verify` nothing.

Issue #637, opened 2026-07-23
Discovery query v1: project inferred types and callable operations from semantic facts


hjosugi:
## Metadata

- State: ready
- Tracker role: curated parent
- Kind: implementation
- Area: IDE Debugger and Profiler
- Priority: P1
- Roadmap milestone: M1-bootstrap
- Size: M
- Parent: #635
- Completed implementation children: #1080, #1094, #1100
- Open bounded child: none; the remaining work landed directly as PR #1159.
- Evidence commit: `225ae59c`
- Generated backlog ID: none

## Summary

Implement the compiler-backed `kofun.discovery.request/v1` and `kofun.discovery.result/v1` query defined by `docs/DEVELOPER_DISCOVERY.md`. Given a file identity/digest and expression position, project inferred type and deterministic callable-operation facts from authoritative semantic analysis without evaluating user code or adding runtime metadata.

This is the semantic service shared by later REPL, CLI, and editor presentation. It does not implement a REPL or runtime reflection.

## Current behavior and evidence

Measured on `origin/main@225ae59c`, after PR #1159 landed the remaining
semantic-fact completeness:

```sh
sh tests/conformance/discovery/run.sh
# PASS: parse / emit / boundaries / facts / facts-refused
# PASS: analysis-key / staleness / select / type / operations
# PASS: live-query / nominal-typeid / bounded-typeid
# PASS: discovery-disabled-artifact
# PASS: AddressSanitizer and UndefinedBehaviorSanitizer
# discovery: OK

sh tests/typed-sidecar/stage2-events.sh   # exit 0
sh spec/typed-sidecar/check.sh            # exit 0
sh tests/backlog/check.sh                 # exit 0
```

The live golden is no longer partial:

```text
authoritative=false
status=complete
reason=null
type.status=validated
type.identity=TypeId 125ac3100e11a7150fb950247e5ecde776975fb10c5fb7a0deaaefdb38bd82c5
type.display=List[Text]
candidate rows=5; callable=5; unavailable=0
```

## Completed topology

- #608 — authoritative semantic producer: closed and consumed.
- #1080 — live Stage 2 discovery adapter and real source fixture: closed and
  focused-gated.
- #1094 — closed; current-file non-generic nominal identities propagate onto
  resolved reference snapshots.
- #1100 — closed; the compiler-owned unary `Int` constructor `Count` is
  validated and callable.
- PR #1159 — the remaining semantic-fact completeness, described below.
[...]

PR #1173, opened 2026-08-10
discovery: close two ways a row claimed a closure it did not have (#637)

hjosugi:
An adversarial review of the #637 candidate-validation change ([PR #1159](https://github.com/kofun-lang/kofun/pull/1159), merged) found two
rows that answer `complete` without the closure that would justify it,
plus one contract rule the emitter stated but did not enforce.

All three produce well-formed output, which is the point: none is
visible from the shape of the answer. Each is now observed as a status.

## 1. A result type is in no dependency list

The signature fact's dependencies are its parameter bindings and nothing
else, so walking them proves nothing about what the function returns.

Measured on `main` before this change, with `Wibble` declared nowhere
and absent from the bounded catalog:

```text
open_result:  status=validated signature=Int -> Wibble
open_nullary: status=validated signature=() -> Wibble
result-closure: status=complete open-result-row-callable=yes
```

That contradicts the contract directly — *"`validated` requires the
complete transitive dependency closure to be validated"* — and my own
commit message's claim about "the whole committed closure".

`producer_type_reference_is_identified` now asks the two owners that
actually issue identities (a current-file declaration, or the bounded
builtin/constructed catalog). A result type neither answers for keeps
the candidate provisional with **no signature disclosed at all**, since
a rendered signature naming an unidentified type is exactly the thing
this slice must not hand out.

The nullary form escaped the same way and predates PR #1159; both are
checked now.

## 2. An unreadable record was counted as a completed analysis

An enum this build does not define is withheld. That is the right
*disclosure* decision — it leaves the same `hidden-by-visibility`
omission a private declaration leaves, deliberately, so the two stay
indistinguishable from outside.

But PR #1159 exempted that omission from forcing `partial`, and a record
the provider cannot read is analysis that did not complete. Corrupting
one candidate's status flipped the answer:

```text
pre-fix:  top-level status=complete  name-disclosed=no  omission=hidden-by-visibility
fixed:    top-level status=partial   name-disclosed=no  omission=hidden-by-visibility
```

Nothing about disclosure changed; only the claim of completeness did.

**The exemption itself stays, and is still right.** A genuinely private
declaration does not make the disclosed facts incomplete, and without
the exemption `complete` is unreachable for any file containing one.

Worth noting for reviewers: `live_query_test.c` already corrupted these
same enums, but asserted only non-disclosure and never the status — so
the existing gates structurally could not catch this.

## 3. Effect uniqueness was by sort key, not by identity

The contract says requirements are *"unique by identity when an identity
is present"*, which is stricter than the sort key. Two requirements
sharing a `SymbolId` but rendering `io.read` and `io.write` sorted
adjacent, compared unequal, and were both emitted.

Unreachable through today's provider, which only emits null-identity
effects — but `facts_are_permitted` exists to distrust providers, and
PR #1159's message claimed repeated requirements were refused.

I also corrected a comment that claimed the identity-kind enum matches
ASCII order. It does not in general (`FileId` is enum 4 but ASCII-first);
it is currently harmless only because effect identities are restricted
to `SymbolId`/`TypeId`.

## Measured

```sh
sh tests/conformance/discovery/run.sh
# PASS: parse / emit / boundaries / facts / facts-refused
# PASS: analysis-key / staleness / select / type / operations
# PASS: live-query / nominal-typeid / bounded-typeid / closure
# PASS: discovery-disabled-artifact
# PASS: AddressSanitizer and UndefinedBehaviorSanitizer
# exit 0
```

**The live golden is byte-identical.** Every type that fixture names is
identified, so nothing there was resting on either defect — which is
also why the original gates went green over both.

The new `closure` lane was proven to refuse by building it against
pre-fix `main` and watching each case fail there.

### DIFF
--- a/bootstrap/stage2/discovery_query.c
+++ b/bootstrap/stage2/discovery_query.c
@@ -88,7 +88,12 @@ static KofunDiscoveryVisibility query_visibility(
     }
 }
 
-static bool query_visible_candidate(
+/*
+ * Every enum in this caller-owned snapshot must be one this build defines.
+ * An unrecognized one is a record the provider cannot read, not a record
+ * that says something.
+ */
+static bool query_readable_candidate(
     const KofunStage2DiscoveryCandidate *candidate
 ) {
     bool valid_kind = candidate->kind == KOFUN_STAGE2_DISCOVERY_FUNCTION ||
@@ -97,10 +102,20 @@ static bool query_visible_candidate(
         candidate->status == KOFUN_SEMANTIC_PROVISIONAL ||
         candidate->status == KOFUN_SEMANTIC_ERROR ||
         candidate->status == KOFUN_SEMANTIC_UNAVAILABLE;
+    bool valid_visibility =
+        candidate->visibility == KOFUN_STAGE2_INTERFACE_PRIVATE ||
+        candidate->visibility == KOFUN_STAGE2_INTERFACE_INTERNAL ||
+        candidate->visibility == KOFUN_STAGE2_INTERFACE_PUBLIC;
+    return valid_kind && valid_status && valid_visibility;
+}
+
+static bool query_visible_candidate(
+    const KofunStage2DiscoveryCandidate *candidate
+) {
     bool visible = candidate->visibility ==
             KOFUN_STAGE2_INTERFACE_INTERNAL ||
         candidate->visibility == KOFUN_STAGE2_INTERFACE_PUBLIC;
-    return valid_kind && valid_status && visible;
+    return query_readable_candidate(candidate) && visible;
 }
 
 static bool query_source(
@@ -233,6 +248,7 @@ size_t kofun_stage2_discovery_query(
     size_t omission_count = 0u;
     bool truncated = false;
     bool complete = true;
+    bool unreadable = false;
     size_t written;
 
     if (analysis == NULL || request_bytes == NULL || output == NULL) return 0u;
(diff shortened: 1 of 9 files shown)


hjosugi:
## Added: a fourth fix, and a note on why the first version of its test was worthless

The review that found the closure defects also noted `sort_rejections` had no guard where `sort_effects` had one. I checked rather than filing it as a nit, and it overflows:

```text
==ERROR: AddressSanitizer: stack-buffer-overflow
    #0 sort_rejections bootstrap/stage2/discovery_v1.c:783
    #1 kofun_discovery_result_emit bootstrap/stage2/discovery_v1.c:1152
```

`facts_are_permitted` does check the count against its array — but it runs *after* the ordering pass, and a sort reads and writes the whole count it is handed. The refusal was at the wrong end of the function to prevent the overrun it exists to catch. Both counts are now bounded before anything sorts.

This predates the branch (`sort_rejections` has been unguarded since the contract layer landed), but it is one line from code this branch already changes, and leaving the guarded and unguarded twins side by side reads as deliberate to the next reader.

### The test I nearly shipped would have passed on the bug

Worth recording, because two separate versions of this case looked fine and proved nothing.

**Attempt 1** set the count to `MAX + 1`. That still lands inside the same struct, so the sanitizer sees nothing — and `facts_are_permitted` refuses an over-limit count in *both* the fixed and unfixed code, so the `refused` line was byte-identical either way.

**Attempt 2** raised the count to 4096 but reused a row from the local `operations[2]` array. Reading 16 KB past that row lands inside the neighbouring element, which is one allocation — again invisible to ASan. It passed on the defect.

**What works** is a row with its own allocation, so there is a redzone immediately after it, plus a count far past the struct rather than one past the array. Verified by building that case against the parent commit and watching it fail as a heap overflow there.

The lesson generalises to the other cases on this branch: a shape assertion that reads the same before and after the fix is not a gate.

---
_Generated by [Claude Code](https://claude.com/claude-code)_

Issue #1183, opened 2026-08-10
Stage 2 C11 aggregate bridge increment 5: List[Int] record fields — the carrier and the layout spec disagree


hjosugi:
## Goal

Let a nominal record carry a bounded `List[Int]` field, so `type Bag = { samples: List[Int], count: Int }` lowers and runs.

Child of #868, the last of its five increments. #1181 landed increment 4 (`Text` fields) and established the record-field carrier, which is the condition the parent set for splitting this one out.

## Measured current boundary

On `origin/main@43d53d38`, with `Text` now admitted:

```
type Bag = { samples: List[Int], count: Int }
error[E2S32]: record `Bag` has a field type outside the Stage 2 Int/Bool/Text slice at byte 0
```

## This is not increment 4 with a different type name

`Text` was two edits because a `Text` value is one reference — `const char *`, eight bytes on this target, which is exactly what `spec/aggregate-layout-v1` already says it is. The placement arithmetic never moved.

`List[Int]` is different in a way that needs a decision before any code:

**The Stage 2 backend's carrier is by value.**

```c
typedef struct { uint64_t length; int64_t elements[64]; } KofunIntListValue;
```

520 bytes, copied whole, no pointer.

**`spec/aggregate-layout-v1` says a List value is a reference.** `layout.mjs`:

> A Text or List value is exactly one reference. The object it addresses is described separately, so the value layout does not depend on the payload.

returning `size: pointerSize`, `align: pointerAlign`, `pointers: [0n]`, `drop: "managed"`.

So for a `List[Int]` **record field** the two disagree about what is stored: eight bytes with one pointer at offset 0 and managed drop, or 520 bytes with no pointers and trivial copy. They give different offsets for every field after it, and different answers to whether a record containing one is Copy.

## The decision, made

**Option 1 is adopted**, recorded as [RFC-0011](https://github.com/kofun-lang/kofun/blob/main/rfcs/0011-bounded-inline-list-layout.md) and the ledger amendment `DD-033/A01`, merged through #1184.

[...]

PR #1187, opened 2026-08-10
stage2: take a record field's alignment from its type, not its size (#1183)

hjosugi:
RFC-0011 **step 2**, on merged `main` (#1184 accepted the RFC, #1186 landed step 1). Behaviour-neutral: no field type is admitted or refused that was not before, and the emitted C is byte-identical.

## The defect

The record emitter derived a field's alignment from its size:

```
let offset = record_align_up(extent, field_size)
if field_size > record_alignment { record_alignment = field_size }
```

That is correct only by coincidence. Every field type admitted today is 1 or 8 bytes, and for those, size and alignment happen to be equal. The coincidence ends at the next admitted type: RFC-0011's `bounded_list[Int, 64]` is **520 bytes and 8-aligned**, so aligning by the size would align a record to 520 and place every following field wrong.

This is a defect the new kind *exposes*, not one it introduces, which is why the RFC sequences it before admitting the field rather than inside that change.

## Both walks, both surfaces

Two walks carry the split, because they must agree with each other — the struct emitter that computes `extent` and `record_alignment`, and the `_Static_assert` offset walk that re-derives the same offsets. Two walks that agree only because both derive alignment from size would drift apart the moment one of them stopped.

The C side already had a `field_alignment` name at the first site, bound to `= field_size` — a decoupling that read as done and was not. It now derives from the field type like the Kofun side, and the second site gains the same.

## Behaviour-identity, proved

Built the Stage 2 compiler from this branch and from `main`, then emitted C for the two executable record fixtures and compared:

| Fixture | Result |
|---|---|
| `production_read_parameter.c` | **byte-identical**, 7822 bytes |
| `production_take_accepted.c` | **byte-identical**, 8559 bytes |

`bootstrap/stage2/SHA256SUMS` and the manifest's `trusted_seed_sha256` are regenerated, as every compiler-pair change does — `sha256sum -c` verifies.

## Validation

```
task stage2            PASS   (the pinned compiler pair)
task records           PASS
task list-int-values   PASS
task text-results      PASS
task const-generics    PASS   (shares the struct emitter via validate_struct_identity)
task adt               PASS
```

## Next

Step 3 admits `List[Int]` record fields in the Stage 2 profile; step 4 is the `E2S32` message sweep across its 17 sites; step 5 flips the four stdlib gates (`array`, `binary_heap`, `set`, `vector`) from asserting refusal to asserting execution.


hjosugi:
CI caught a fourth pinned digest I had missed. "Release evidence pack" failed while every compiler gate was green.

`artifacts/release-evidence/index.json` records a hash of `bootstrap/manifest.json`, and the job regenerates the pack then asserts `git diff --exit-code -- artifacts/release-evidence`. Moving `trusted_seed_sha256` therefore invalidated it:

```
-    "bootstrap/manifest.json": "f9c9da7c...f446"
+    "bootstrap/manifest.json": "4d9ab808...c537"
```

So a Stage 2 pair change moves **four** recorded digests, not three: both lines of `bootstrap/stage2/SHA256SUMS`, the manifest's `trusted_seed_sha256` (which is the `compiler.c` digest), and the release-evidence pack that hashes the manifest. The first three fail locally in `task stage2`; the fourth has no local gate in the usual set and only shows up here.

Fixed by regenerating with `task release-evidence` rather than hand-editing — that command owns the file. Only the manifest digest line moves; claim states and every other recorded hash are unchanged.

Issue #1189, opened 2026-08-10
Call arguments v1 slice 3a: lower wider and ownership-bearing carrier temporaries


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation child
- Kind: implementation
- Area: Code Generation / Call Arguments
- Priority: P1
- Size: M
- Parent: #882
- Blocked by: none
- Audited against: `origin/main@1c27baa804f825b93c031648c1c627f477236316`

## Goal

Lower labelled calls whose arguments are wider than `Int` — including `Optional` and ownership-bearing carriers — through ordinary temporaries and fixed declaration/ABI slots.

#1097 delivered the bounded all-`Int` direct slice. This is the first widening of the carrier, and nothing else.

## Current behavior and evidence

Measured on exact `origin/main@1c27baa804f825b93c031648c1c627f477236316` with the committed `bin/kofun` seed.

```
tests/conformance/call-arguments/reordered_optional.kofun
# error[E2S158]: labelled-call ABI lowering is owned by #882;
#   fixed-slot checked HIR is available at byte 138
# exit 3
```

`task call-arguments`, `task call-arguments-spec`, and `task call-arguments-surface` are green on that commit and prove the bounded all-`Int` direct slice from #1097 only.

## Scope

**In scope:** wider and ownership-bearing carrier temporaries in labelled calls; the existing `reordered_optional` and `typed_and_owned` shapes.

**Out of scope:** pipeline subjects (3b), the trailing-lambda form and lifted-lambda bodies (3c), direct-native and Wasm lowering (3d), lexical callable shadowing, default arguments, and optional call-site labels.

## Acceptance criteria

- [ ] `reordered_optional.kofun` lowers, builds under the same `-Wall -Wextra -Werror -pedantic` as every other accepted call-arguments program, and runs.
- [ ] A source-order golden proves exactly-once evaluation for a wider carrier, in the shape of `source_order_carriers.stdout`.
- [ ] An ownership-bearing carrier is moved into its slot exactly once, and a double move is refused by the existing registered code.
- [ ] Shapes outside this slice keep refusing with `E2S158` and publish no accepted C artifact.
- [ ] No runtime label map, string dispatch, or extra allocation appears in the emitted C.

## Validation

| Check | Command | Expected result |
| --- | --- | --- |
| This slice | `task call-arguments` | the new shapes lower, execute, and evaluate exactly once in source order |
[...]

Issue #1107, opened 2026-08-09
Stage 2 C11: lower labelled calls with Text and List[Int] carriers


hjosugi:
## Metadata

- State: ready
- Tracker role: bounded implementation child
- Kind: implementation
- Area: Code Generation / Call Arguments
- Priority: P1
- Size: M
- Parent: #882
- Blocked by: none

## Outcome

Widen the #1097 labelled-call C11 lowering from all-`Int` carriers to direct top-level calls whose parameters and result may also be `Text` or `List[Int]`. Evaluate written arguments exactly once in source order into function-local typed temporaries, then invoke the declaration-order ABI slots — with no runtime label data and no allocation or deep copy introduced by the temporaries beyond what the same call already performs positionally.

## Current behavior and evidence

Measured on exact `origin/main@862bb4b7acdea40ac01a4efa4a6e723adec02275` on 2026-08-09 with `kofun-stage2 --compile-outcome`:

- All-`Int` labelled call (`add3(third: 3, first: 1, second: 2)`): exit 0, C emitted — the #1097 slice.
- `Text` carrier (`tag(name who: Text)`; call `tag(name: "hi")`): `error[E2S158]: labelled-call ABI lowering is owned by #882; fixed-slot checked HIR is available at byte 93`, exit 3, no C artifact.
- `List[Int]` carrier (`pick(items xs: List[Int], index i: Int)`; call `pick(items: xs, index: 1)`): same exact `E2S158`, exit 3, no C artifact.
- `Bool` parameter: refused upstream by `error[E2S15]: Core parameters must have type Int, Text, a concrete enum, or a nominal record` — `Bool` is not a legal parameter carrier, so it is not part of this slice.
- Positional call to a labelled declaration: `error[E2S164]: labelled parameter requires its external label` — the #881 binding contract, unchanged here.

The prerequisites landed: #1097 owns the all-`Int` temporary lowering (`labelled_int_call_supported` / `emit_labelled_int_call`, `bootstrap/stage2/compiler.kofun:3776` and `:4596`), and #1103 (commits `744e5f47`, `5463ccf9..cadf9a00`) made `List[Int]` parameters and results pass across Stage 2 C11 calls positionally. `Text` parameters and results are established (`task text-results`). Only the labelled path refuses these carriers.

## Required artifact

Edit the Stage 2 canonical pair/checksum and `tests/conformance/call-arguments/**`:

[...]

Issue #784, opened 2026-07-28
ownership: an affine handle whose owned state cannot be duplicated (#644)


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation child
- Kind: implementation
- Area: Ownership and Borrowing
- Priority: P2
- Roadmap milestone: M2-alpha
- Size: M
- Parent: #644
- Blocks: #644
- Blocked by: none
- Audited against: `origin/main@1639a8deb811d3f04acc5d84a4fd73f7630e8be0`

## Summary

Give Kofun one affine resource handle whose owned state cannot be duplicated across write, read, drain, close, cancel, and reuse transitions, with terminal state observed by an executed fixture. This is the remaining producer prerequisite for #644's bounded HTTP client.

## Current behavior and evidence

Kofun now has three relevant bounded precedents, but no artifact satisfies this issue:

- `task records` executes whole-record moves through Stage 2 C11, binds and lowers `read` parameters, and rejects partial move, double take, and use after move. It is still not a reusable resource-state protocol and does not define write/read/drain/close/cancel/reuse transition authority.
- `task clock-adapters` executes a type-specific affine clock handle and deterministic cancellation through reference and C11 backends. Its own README names this issue as the open general affine-handle decision.
- `task affine-resumption` provides an executable checker/model for one-shot continuation consume, transfer, drop, escape refusal, and a runtime double-resume backstop. #735 is complete, but that contract deliberately does not define transport write/read/drain/close/reuse states.

`docs/MVP_IMPLEMENTED.md`, `docs/ROADMAP.md`, and `release/claims.json` still state that no general ownership pass exists. No RFC ledger row names #784. The parent #644 still names this handle as a blocker, but its accepted response-body stream contract is also specified-only: no stream operators, scheduler, or adapters exist, and #644 must refine whether it owns that projection or names another producer. Its HTTP client corpus is not present.

## Decision required before implementation

Choose the smallest ownership authority that may define this handle:

1. a bounded per-type affine protocol with a closed declared transition table;
2. an extension of the executable record move checker to one named resource type; or
3. the first slice of a general ownership pass.

The decision must specify:

- the declaration and transition spelling;
- ownership of state before and after write/read/drain/close/cancel;
- whether the terminal observation borrows a consumed handle, returns a separate copyable summary, or uses another bounded rule;
- how normal completion and cancellation choose reuse;
- how the bounded design is subsumed by a later general ownership pass;
[...]

PR #1199, opened 2026-08-10
stage2: widen labelled-call carriers and move a bare take binding (#1189)

hjosugi:
Slice 3a of #882, and the first widening of the call-argument carrier since #1107. Picked up from an unpushed working tree — the implementation is the prior agent's; this PR rebases it onto current `main`, regenerates the digests, and gates it.

## What lands

Direct top-level labelled calls now lower `Int?`, a concrete enum, and a nominal record in their fixed parameter and result slots, alongside the existing `Int`/`Text`/`List[Int]` set. A bare binding placed in a slot declared `take` becomes one source-order semantic transfer.

Each written argument still goes to a function-local temporary of its own carrier type, sequenced through the C11 comma operator before the declaration-order ABI vector.

## Evidence

`source_order_wide.kofun` writes `Int?`, enum, record, and `Int` slots in the order 3, 2, 0, 1 and prints `1 2 3 4` before the callee's `10` — the widening did not buy source order at the cost of exactly-once evaluation. `reordered_optional.kofun` pins the Optional narrowing path on its own.

`owned_carrier.kofun` proves a bare record binding in a `take` slot is one move even though the C carrier assignment is mechanically a copy. `double_move_carrier.kofun` pins the later transfer to the existing exact `E2S123`, use site primary and move site named.

## What this deliberately is not

The ownership half is the smaller rule the spec now states: it reads source-order semantic `take` facts only. No inference from compound expressions, and no alias, branch, loop, lifetime, destructor, or cleanup analysis. `docs/MEMORY_MODEL.md` keeps every one of those denials and adds `task call-arguments` as the second gate on the E2S122/E2S123 row.

Pipeline subjects, the trailing-lambda form, labelled calls inside lifted lambdas, indirect/lexical callees, and the direct-native/Wasm backends stay at the explicit `E2S158` boundary owned by #882.

## Rebase notes

Rebasing onto `main` after #1198 conflicted only in the two generated digest files. Both were regenerated from the merged bytes rather than hand-resolved: `bootstrap/stage2/SHA256SUMS` via `sha256sum`, `bootstrap/manifest.json`'s `trusted_seed_sha256` from the merged `compiler.c`, and the release evidence pack via `task release-evidence`. The stage2-events census is unchanged at 339 — the new refusal companion is a `.diagnostic`, which that count does not read.

Refs #1189.


hjosugi:
## Full `task verify` caught a regression in the handover; pushed a fix

`task call-arguments` passed on its own, but the full run refused a program the RFC-0010 gate requires the compiler to **accept** — `tests/ownership/affine-resource-handle/transport.kofun`:

```
error[E2S123]: `initial` was already moved by `take` at bytes 2841..2853;
  first moved by `take` at bytes 2825..2832
```

That fixture writes:

```kofun
let written: AffineTransport = affine_transport_write(initial, 5)
take initial
```

`affine_transport_write` declares `take handle`, so the new rule counted the **positional** call as the move and the written `take initial` as a second one. The accepted #784 model is the reverse: the call is not the transfer, the explicit `take` is.

The rule was already documented as labelled-only — both `spec/syntax/call-arguments-v1.md` and `docs/MEMORY_MODEL.md` say "labelled" — but `move_call_binding` asked only whether the resolved slot was `take`, and `call_argument_parameter_property` resolves positional arguments too. The code shipped wider than its own contract; 1d1d73ac restores the contract rather than changing it.

Labelled-ness now comes out of that same walk as a third property beside `mode` and `type`, so the ownership rule and the slot it names cannot disagree about which argument matched. Widening to positional `take` calls would amend the accepted affine model and stays owned by #882.

`owned_carrier` and `double_move_carrier` both call `consume(into: ticket)`, so the bounded transfer and its `E2S123` are unaffected. `task affine-resource-handle`, `task call-arguments`, and `task records` now pass; full `task verify` is re-running.

The earlier `selfhost generations: A1 and A2 exit differently` line in the same run was a `--failfast` teardown artifact, not an independent failure.

Issue #1205, opened 2026-08-11
[Umbrella] Remove repeated C compilation from task verify


hjosugi:
## Metadata

- State: in-progress
- Tracker role: planning umbrella
- Kind: planning
- Area: Quality / Build system
- Priority: P2
- Size: L
- Blocked by: none
- Completed foundations: #1238, #1319, #1326, #1239
- Audited against: `origin/main@50f79e38c1b273a5870905b693b1e9b3267c7a37`

## Goal

Remove measured avoidable C compilation from `task verify` without reducing warning, sanitizer, analyzer, compiler-diversity, semantic, or gate coverage.

This issue is the planning umbrella for the measured implementation slices and their final remeasurement. It does not own one combined implementation PR and is never directly implementation-ready.

## Measured baseline

The original serial full-suite census measured:

| Measure | Value |
| --- | ---: |
| suite wall time | 1658 s |
| C compilation | 1075 s (64.8%) |
| repeated already-seen source sets | 872 s (52.6% of suite) |
| bounded `VERIFY_JOBS=3` wall time | about 607 s |

The original basename aggregation overstated Stage 2 `compiler.c` as 64 builds; full-path aggregation found 37. The suite-level C time was unaffected. Every child must therefore classify full source paths and effective compiler argv rather than basenames.

## Delivered foundations

- #1202 / PR #1206 made the verify runner, conformance adapter, and owned gates reuse the one exported Stage 2 compiler.
- #1207 made `bin/kofun` reuse that compiler.
- #1238 is merged on the audited main. It delivered the unique verify-run directory, immutable semantic object bundle, fail-closed source-or-object selection, standalone fallback, compiler/source/profile/member provenance, content-keyed semantic links, cleanup, and actual full-run census.
- #1238's measured full run was green in 519.597 seconds with standard semantic producer main/library exactly once each, standard events and SHA exactly once each, and unexpected/failed producer builds zero. That focused evidence is not a claim that all remaining suite time is removable.
- The optimized `-Werror` build and every deliberately different sanitizer/analyzer/PIC/compiler profile remain present.

## Disproved approach

PR #1208 measured ccache in CI and was closed without merge:

```text
Cacheable calls:      13 / 975 (1.33%)
Uncacheable calls:   962 / 975 (98.67%)
```

The repository normally compiles several sources and links them in one invocation. ccache does not serve that shape. Adding an apt dependency, PATH masquerade, cache restore, and guards for 1.33% coverage would be negative value.

[...]

Issue #1202, opened 2026-08-11
verify: reuse one Stage 2 compiler and one self-host profile scan


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation issue
- Kind: maintenance
- Area: Build system / CI / verification
- Priority: P1
- Size: M
- Blocked by: none
- Audited against: `origin/main@b6947cc9ecc54ab7465bc4ea7166c4a960d1d04b`

## Goal

Make `task verify` reuse one ordinary optimized Stage 2 compiler build and perform the self-host profile's shared evidence scan once, without removing a gate, fixture, sanitizer/analyzer build, diagnostic, or supported toolchain check.

This is deletion of redundant build/scan code. It is not a compiler refactor, a parallelism change, or selective-test policy.

## Current behavior and measured evidence

Measured on exact `origin/main@b6947cc9ecc54ab7465bc4ea7166c4a960d1d04b`.

`task verify` already builds `build/verify/kofun-stage2` once and exports it as `KOFUN_STAGE2_COMPILER`. `bootstrap/stage2/build.sh` copies that binary for gate-local paths and documents that a fresh ordinary compile costs roughly 4.6 seconds.

Seven ordinary consumers still bypass that helper and compile the same `bootstrap/stage2/compiler.c` again:

- `tests/conformance/adt-exhaustiveness/run.sh`
- `tests/conformance/call-arguments/run.sh`
- `tests/conformance/const-generics/run.sh`
- `tests/conformance/discovery/run.sh` for its unmodified comparison arm
- `tests/conformance/syntax/issues_48_60/run.sh`
- `tests/conformance/backends/c11-stage2.sh`
- `spec/roadmap-31-34/verify-current-gates.sh`

The discovery macro-variant compile, sanitizer builds, analyzer builds, compiler-pair mutations, and diverse-host builds are semantically distinct and must remain.

The self-host profile also repeats its shared 46-row/10-corpus evidence scan four times: default, `--phase frontend`, `--phase c11-text`, and `--phase c11-control`. Local exact-main measurements were 4.399s, 4.364s, 4.591s, and 4.729s. The phase-only assertions run after the same base scan.

Recent successful GitHub Actions runs spent 664–899 seconds inside `task verify`; qemu installation cost only 10–14 seconds. Eliminating repeated repository work is therefore the first no-coverage-loss optimization.

### In scope

- route the seven ordinary Stage 2 consumers through `kofun_stage2_build`;
[...]

PR #1208, opened 2026-08-11
ci: front the compilers with a cache, and prove it stays invisible (#1205)

hjosugi:
## Why a cache and not more reuse

Measured on `main@0af511f2` by wrapping `cc` and running all 117 gates serially:

| Measure | Value |
| --- | --- |
| Serial suite | 1658 s |
| C compilation | 1075 s — **64.8%** |
| Recompiling source sets already built in the same run | 872 s — **52.6% of the suite** |

#1202 removed the repeated builds of `bootstrap/stage2/compiler.c`. It could not remove the rest: `imports_selective.c`, `re_exports.c`, `kif_v1_tool.c`, `semantic_producer.c`, and `incremental_graph.c` are each compiled **from source in every one of the ~15 gate scripts that needs them**, dragging `kofun_unicode.c`, `sha256.c`, `kif_v1.c`, and `visibility_access.c` along each time — roughly 350 s per run.

No individual site can fix that, because no site knows what the others already built. A cache does not need to know. Restored across runs, the usual case — one translation unit edited — stops rebuilding the others at all.

## `CC` is deliberately not touched

`bootstrap/selfhost/generations-lib.sh` derives `compiler_identity` from `"$CC" --version`, and `bootstrap/manifest.json` records that string for the closure measurement. Pointing `CC` at a launcher would rewrite a recorded fact — and would break every gate that invokes `"$CC"` as a single word.

A PATH masquerade leaves `CC` alone. ccache forwards `--version` to the compiler it fronts.

## That claim is proved, not assumed

CI reads `cc --version` **through** the masquerade and fails before any gate if the answer names ccache:

```
- name: Prove the cache is transparent to the toolchain identity
```

`tooling/ccache-masquerade.sh` also links only compilers that are actually present — the diverse-double-compilation gate picks its two compilers by what it finds, and inventing a `clang` would change which two it picks.

The cache cannot change what a gate observes: ccache returns an object only for an identical compilation, so the byte-identical artifact comparison in `tests/conformance/discovery/run.sh` and the generation digests in the selfhost chain see exactly what they see today.

## What is and is not proven locally

ccache is **not installed** on the machine this was written on, so **the cache-hit path is proven by this PR's CI run, not locally**. `ccache --show-stats` is printed after `task verify` so the hit rate is visible rather than claimed.

Locally verified: the script refuses with a diagnostic when ccache is absent, and `repository-check`, `build-system`, `documentation-index`, and `task-help` all pass.

Refs #1205.


hjosugi:
## CI refused the first version, correctly

```
FAIL: selfhost diverse double compilation: `gcc` and `clang` resolve to the
same binary (/usr/bin/ccache): the pair is not diverse
```

I fronted `cc`, `gcc`, and `clang`. That collapsed the two-toolchain defence into one binary, and the gate caught it.

The gate is right. Diverse double compilation exists because a payload living in one host compiler cannot be caught by a chain that only ever runs that compiler, so it needs the two to be *genuinely different binaries*. The cache is transparent to what a compiler **emits** — but that gate is about **which compiler ran at all**, and it cannot see through a symlink to answer that.

I had anticipated the identity risk (`--version` reaching the recorded `compiler_identity`) and guarded it. I did not anticipate the distinctness risk, which is a different property of the same masquerade.

## Fix: front only `cc`

Almost all the benefit survives. Measured over a full serial run:

| compiler | invocations | time |
| --- | --- | --- |
| `cc` | 1004 | **987 s** |
| `clang` | 394 | 55 s |
| `gcc` | 31 | 33 s |

`cc` alone is 92% of the compile time. `gcc` and `clang` are left completely alone, so the diverse pair stays diverse.

The workflow now also **proves** the pair is still distinct — next to the identity check, before any gate runs — so this cannot come back as a suite failure forty gates later.

hjosugi:
## Closing: the CI run disproves this PR's premise

The run went green, and the guards worked — but the cache statistics show it buys almost nothing:

```
Cacheable calls:     13 / 972 ( 1.34%)
Uncacheable calls:  959 / 972 (98.66%)
  Hits:               1 /  13
```

**98.66% of compiler invocations are uncacheable.** ccache caches preprocessing-and-compilation to an object file; it does not cache linking, and it does not support several source files in one invocation. This repository compiles and links in a single command almost everywhere:

```
$ grep -rhE '\-std=c11' --include='*.sh' tests bootstrap spec | wc -l
272
$ ... of which use -c
2
```

270 of 272 build lines produce an executable directly, often from several `.c` files at once. That is precisely the shape ccache cannot cache, so the 872 s of repeat compilation I measured is out of its reach.

I was wrong about the mechanism. The measurement that motivated this PR is still correct — compilation is 64.8% of the suite and most of it is repeats — but a compile cache is the wrong instrument for it while the builds are compile-and-link.

Merging this would add an apt package, a cache action, a PATH masquerade, and two guard steps to every CI run in exchange for a 1.34% cacheable rate. That is complexity with no payoff, so I am closing it rather than landing it.

What the two guards proved is worth keeping on the record:

```
cc --version through the cache: cc (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0
gcc:   /usr/bin/x86_64-linux-gnu-gcc-13
clang: /usr/lib/llvm-18/bin/clang
```

The masquerade is transparent to the recorded toolchain identity, and fronting `cc` alone keeps `gcc` and `clang` distinct for `selfhost-diverse-double-compilation`. If a cache is ever revisited, those two checks are the ones it needs.

The finding and what a real fix would require are recorded on #1205.

hjosugi:
The latest green run disproves the implementation premise rather than validating it. `ccache --show-stats` reported 13 cacheable calls out of 975 (1.33%) and 962 uncacheable calls (98.67%). All 13 cacheable calls hit, so the wrapper works, but the repeated frontend bundles are multi-source compile+link invocations that ccache cannot cache. This does not meet #1205’s compile-time or repeat-compilation acceptance criteria, so it must not land. Run: https://github.com/kofun-lang/kofun/actions/runs/31457825679

PR #1247, opened 2026-08-11
ci: add reproducible scheduled fuzz lane

### LINKED ISSUE #1210: Sustained fuzzing: run the generators on a schedule with a rotating seed, and record findings
## Metadata

- State: ready
- Tracker role: curated child
- Kind: test-quality
- Area: Quality / Fuzzing
- Priority: P2
- Roadmap milestone: M4-1.0
- Size: M
- Parent: #1119
- Blocked by: none
- Audited against: `origin/main@8fd0b25497c1fd4c61085260a416af619beb3c3f`

## Goal

Run the existing fuzz generators on a schedule with a seed that changes per run, and record what they find, so accumulated machine time buys coverage instead of repeating one run.

This is the executable half of the M4 exit criterion "sustained fuzzing without unresolved critical findings". It does not widen any generator's grammar.

Measured on `origin/main@8fd0b254`.

`task fuzz` runs ten generators. Nine already take a seed from the environment and print the one they used:

| generator | seed variable |
| --- | --- |
| `grammar.sh` | `KOFUN_GRAMMAR_FUZZ_SEED` |
| `semantic_differential.sh` | `KOFUN_SEMANTIC_FUZZ_SEED` |
| `value_if.sh` | `KOFUN_VALUE_IF_FUZZ_SEED` |
| `match_guard.sh` | `KOFUN_MATCH_GUARD_FUZZ_SEED` |
| `match_value.sh` | `KOFUN_MATCH_VALUE_FUZZ_SEED` |
| `match_value_invalid.sh` | `KOFUN_MATCH_VALUE_INVALID_FUZZ_SEED` |
| `enum_match.sh` | `KOFUN_ENUM_MATCH_FUZZ_SEED` |
| `optional_narrowing.sh` | `KOFUN_OPTIONAL_NARROWING_FUZZ_SEED` |
| `visibility-artifacts.sh` | `KOFUN_VISIBILITY_ARTIFACTS_FUZZ_SEED` |
| `semantic_protocol_test.sh` | **none** |

`hm_levels.sh` and `pure_io.sh` run under sibling tasks and take a case count (`KOFUN_HM_LEVELS_CASES`, `KOFUN_PURE_IO_CASES`) but no seed.

The override works today and the seed is reported:

```
$ sh tests/fuzz/enum_match.sh | head -1
enum_match fuzz: seed=1516993677

$ KOFUN_ENUM_MATCH_FUZZ_SEED=987654321 sh tests/fuzz/enum_match.sh | head -1
enum_match fuzz: seed=987654321
```

`enum_match.sh` says why the default is fixed:

> The default is the seed this corpus was recorded with, so `task verify` […] the same seed means accumulated machine time buys no coverage.

So the substrate exists and the gap is narrow. What is missing:

- **No schedule.** `grep -rlE 'schedule:|cron' .github/workflows/` returns nothing; all three CI jobs are `push`/`pull_request`/`workflow_dispatch` only. Every run therefore explores the same inputs as every other run.
- **No findings register.** `git ls-files | grep -iE 'finding|triage'` returns nothing, so a failure has nowhere to be recorded or triaged, and the exit criterion's second half — "without unresolved critical findings" — is not merely unmet but unmeasurable.

## Scope

**In scope:**

- a scheduled workflow that runs the fuzz generators with a per-run seed;
- a seed derived from the run so a failure is reproducible from the log alone;
- giving `semantic_protocol_test.sh` a seed variable, or excluding it from the lane with the reason stated in the workflow;
- a findings register with a defined row format, and the workflow raising a finding when the lane fails.

**Out of scope:**

- widening any generator's grammar or case count;
- coverage instrumentation;
- changing the default seeds, which must stay fixed;
- the other M4 exit criteria.

## The defaults must not move

`task verify` is deterministic today because each generator's default seed is the one its corpus was recorded with. A scheduled lane must vary the seed **only in that lane**. If the defaults change, or the lane's seed leaks into `task verify`, every pull request becomes intermittently red for reasons its author cannot fix — which is the same failure mode the `backlog` job is deliberately kept off pull requests to avoid.

## Acceptance criteria

- [ ] A scheduled workflow runs the fuzz generators on a fixed cadence, with a seed that differs between runs.
- [ ] Every generator the lane runs prints the seed it used, including `semantic_protocol_test.sh` or an explicit statement of why it is excluded.
- [ ] A failing scheduled run names the exact seed, and re-running that generator locally with that seed reproduces the failure. Demonstrated by forcing one failure and reproducing it.
- [ ] A findings register exists, with a stated row format covering at least date, generator, seed, and resolution, and the workflow records a finding when the lane fails.
- [ ] `task verify` remains byte-identical in behaviour: the committed default seeds are unchanged, and a pull-request run is unaffected by the lane.
- [ ] The lane's failure is visible without watching CI — it does not fail silently into a green-looking repository.

## Validation

| --- | --- | --- |
| Determinism preserved | `task fuzz` twice with no environment | byte-identical output both times |
| Seed plumbing | `KOFUN_<GEN>_FUZZ_SEED=N sh tests/fuzz/<gen>.sh` | prints `seed=N` and generates from it |
| Reproduction | re-run a failed scheduled seed locally | same failure |
| Repository | `task verify` | green, unchanged |

## Dependencies

- Blocked by: none.
- Parent: #1119, which records this as an M4 exit criterion.
- Touches `.github/workflows/` and `tests/fuzz/`, not the canonical `compiler.kofun`/`compiler.c` pair, so it does not serialise against the compiler-pair issues.

hjosugi:
Closes #1210

## Summary

- add a read-only daily fuzz workflow with deterministic run-derived 31-bit seeds for all nine randomized generators
- retain exact logs, seeds, a schema-valid finding artifact, and executable reproducers on failure
- keep ordinary `task fuzz`, committed default seeds, existing generators, `ci.yml`, and `Taskfile.yml` unchanged
- explicitly exclude the deterministic semantic protocol self-test from only the scheduled randomized lane

## Validation

- `sh tests/fuzz/scheduled-check.sh`
- scheduled run across all nine generators with bounded test budgets
- forced generator failure followed by byte-identical local reproduction
- `task fuzz` twice with byte-identical stdout/stderr
- `task release-claims repository-check`
- `git diff --check`
- `graphify update .`

## Safety

The workflow grants only `contents: read`, checkout credentials are not persisted, no issue/commit mutation occurs in Actions, and a failed generator is converted back into a failed workflow after artifacts and the step summary are published.


hjosugi:
Superseded by merged PR #1259 on main. This draft would add a second daily workflow over the same nine generators, so it must not merge unchanged. The stronger machine-readable findings/artifact ideas should return only as a separate non-duplicating hardening slice.

Issue #1291, opened 2026-08-11
B6: obtain and validate one policy-compliant independent clean-builder attestation


hjosugi:
## Metadata

- State: blocked
- Tracker role: curated external-evidence child
- Kind: test-quality
- Area: Bootstrap / independent reproduction
- Priority: P1
- Size: M
- Parent: #274
- Blocked by: an eligible independent builder willing to run the finalized
  reproduction packet on hardware they control and author the attestation pull
  request from their own GitHub account. This non-issue blocker is recorded in
  `tests/backlog/debt.tsv` as `capability-blocker\t1291\t-`.
- Pre-run prerequisite: #1530 must close by a merged implementation on `main`
  before the exact run contract is issued or the packet is executed.
- Policy authority: `bootstrap/selfhost/b6/POLICY.md`
- Audited against: `origin/main@20cee8101c2792e0725fb6dd2211d1b0bfcec500`

## Goal

After #1530's non-self-referential acquisition-identity contract has landed
on `main`, have one policy-eligible external builder independently run the
exact finalized B6 packet for one exact `main` commit and author an ordinary
pull request containing its canonical attestation report.

The pull request must also retain or durably link the complete run logs and
identify their digests, so a maintainer can review the report, its subject,
the policy version, and the human independence claim without treating a green
mechanical validator as authentication.

This issue tracks obtaining, reviewing, and merging that external evidence. It
does not design the acquisition identity, authenticate a builder by automation,
or promote B6 into manifest and release truth. It is not by itself remote-write
authorization for an agent; any agent performing a future PR or ref write still
needs a current user request authorizing the exact internal target and action.

## Current behavior and evidence

On the audited revision, no independent attestation exists. The one committed
report is the repository's historical producer-side self-check, and its old
acquisition key does not describe the audited checkout:

```sh
test "$(git rev-parse HEAD)" = \
  20cee8101c2792e0725fb6dd2211d1b0bfcec500 || exit 1
test -z "$(git status --porcelain --untracked-files=no)" || exit 1

work=$(mktemp -d)
trap 'rm -rf "$work"' EXIT HUP INT TERM
sh bootstrap/selfhost/declare-inputs.sh "$work/current" >/dev/null

attestations=$(find bootstrap/selfhost/b6 -type f \
  -name 'attestation-*.tsv' | wc -l | tr -d ' ')
retained_key=$(awk -F '|' '$1 == "result" &&
  $2 == "acquisition_set_sha256" { print $3 }' \
[...]

Issue #1256, opened 2026-08-11
Decide the bounded executable carrier profile for the HTTP/1.1 core


hjosugi:
## Metadata

- State: needs-decision
- Tracker role: curated profile decision
- Kind: spec-design
- Area: HTTP / Bytes / stream / C11 Stage 2
- Priority: P1
- Size: M
- Decision owner: hjosugi
- Parent: #644
- Blocked by: none
- Normative inputs: `docs/stdlib/http-client.md`, `docs/stdlib/stream-protocol.md`, `stdlib/CONTRACT.md`
- Audited against: `origin/main@3f72c7df8f5c87a7c92a96ff62042bb6cca2745e`

## Goal

Choose the concrete bounded source/runtime profile that can express the accepted HTTP core on C11 Stage 2 without weakening its Bytes, header, typed-error, affine-stream, or 64-KiB limit semantics.

## Current missing choices

- Shared `Bytes[65536]` lowering is split across #1315, #1321, and #1322. This HTTP decision consumes the existing RFC-0004 Managed, non-Copy value through that bounded alias-free C11 storage profile and does not redesign it.
- It lowers only bounded `List[Int]`; no `List[Header]`/generic collection carrier exists.
- the product generic path needed for `Stream[Bytes, HttpError]` is not implemented (#31); the affine-resource gate is a monomorphic Int/Bool checkpoint.
- the accepted client contract names a pre-parsed `Url`, validated headers, `Result`, `Bytes`, and stream protocol, but does not freeze their first executable ABI.
- default maximum header bytes is 64 KiB, which cannot be represented by the current 64-element `List[Int]` profile.

## Decisions required

1. **Bytes carrier (fixed input).** Consume #1315 + #1321 + #1322: the existing builtin RFC-0004 Managed, non-Copy `Bytes`; bounded alias-free C11 storage; exact capacity 65,536; arbitrary bytes; nonescaping read/edit, proven-unique take, and explicit checked Text conversion. #1256 does not choose GC/RC/COW, make Bytes Owned, or reimplement it; it chooses only how HTTP-specific carriers compose the completed shared profile.
2. **Header collection.** Choose a bounded `List[Header]` product carrier or an explicitly named `HeaderList` value; freeze count, name/value byte limits, order/duplicate behavior, and ownership. Raw unindexed header bytes are not an acceptable public substitute.
[...]

Issue #1060, opened 2026-08-07
backlog: 24 of 28 blocked issues name their blockers where the gate cannot read them


hjosugi:
## Metadata

- State: needs-decision
- Tracker role: curated child
- Kind: test-quality
- Area: quality / backlog tooling
- Priority: P2
- Size: S
- Parent: #1048
- Blocked by: none
- Audited against: `03d8a7099fa14c62c0fa8d250c0d5d8c2c3ffc38`

## The door #1048 named but did not open

#1048 fixed the `State:` line and closed with one deferral, quoted exactly:

> The `Blocked by:` prose problem as a general fix — worth its own issue if the
> split shows it is widespread.

It is widespread. Measured on `main@03d8a709` across all 70 open issues.

`tests/backlog/check.mjs` reads blockers with one regex in
`tests/backlog/extract.mjs`, and the rule keyed on it is
`blocked issues with named blockers still have an open blocker or recorded debt`.
A blocker stated in prose does not parse, so the issue gets `blocked_by: []` and
the rule passes it vacuously.

**Of 28 open `blocked` issues, 4 carry a `Blocked by:` line naming an issue.
The rule reaches 4 of 28 — 14%.**

That is the same shape #1048 corrected for `State:`, on a different door: the
gate prints `PASS: 4 blocked issues with named blockers still have an open
blocker or recorded debt` and the `4` reads as a total when it is a sample.

## What the unreached 24 are hiding

Fifteen of the 24 name a blocker somewhere in prose. Resolving every issue
number those sentences mention:

### Blocker closed, nothing else named — candidates for re-triage

| Issue | Prose says | Blocker state |
|---|---|---|
| **#314** Null coalescing | `Unblock condition: **fulfilled.**` — all three prerequisites on `main`, run 30728562695 green | #312, #120 both closed |
| **#915** ensure_move v3 | `**Blocked on #904.**` and explicitly `Not blocked by #891` | #904 closed as completed |
| **#576** Koka reuse | `**Blocked by #920**, which owns kofun.reuse-candidate/v1` | #920 closed as completed |

**#314 is the sharpest case: its own body states the unblock condition is
fulfilled and cites the green CI run, and it is still labelled `blocked`.** No
gate could notice, because the sentence that says so is prose.

### Correctly blocked, but on something no issue number can express

| Issue | Blocked on |
|---|---|
| #26 wasm32 target | "the still-unowned WASI capability profile and engine/browser/WASI support matrix" |
[...]

Issue #644, opened 2026-07-23
[Umbrella] Deliver the bounded HTTP/1.1 scripted-transport client core


hjosugi:
## Metadata

- State: blocked
- Tracker role: implementation umbrella
- Kind: planning
- Area: Core Standard Library / HTTP / Ownership
- Priority: P2
- Roadmap milestone: M2-alpha
- Size: umbrella
- Parent contract: #638 — complete
- Direct delivery children: #1256, #1257, #1258, #1260, #1261, #1262, #1263, #1264
- Audited against: `origin/main@78b409c7ba69ea8fe21bdaea911197430f0e05b8`
- Contract: `docs/stdlib/http-client.md`

## Outcome

Deliver and certify the first bounded HTTP/1.1 client core over an injected scripted byte transport. The accepted protocol contract is preserved; executable carrier, generic stream, serializer, parser, body-lifecycle, and release-evidence work are now owned by separate reviewable children.

This parent owns coordination and closes only after all delivery children close. It must not receive production code directly.

## Why the old issue was not startable

The previous issue treated records, payload ADTs, and the affine-resource checkpoint as sufficient prerequisites. Exact source/backend measurement shows four missing executable surfaces:

- Stage 2 C11 does not lower the stdlib `Bytes` carrier used by the HTTP contract.
- The current bounded `List[Int]` profile cannot represent ordered headers or the accepted 64-KiB header limit.
- `Stream[T,E]` is specified but has no production implementation; #784 proves only a monomorphic Int/Bool affine state table.
- There is no normalized HTTP origin carrier, incremental request/response implementation, or client vertical fixture.

A host JavaScript parser/model, the existing HTTP server, or the affine table alone cannot satisfy the implementation claim.

## Delivery DAG

```text
#1257 independent HTTP oracle/corpus ───────┬─────────> #1261 request serializer ──┐
                                            └─────────> #1262 head/framing parser ─┼─> #1263 body stream ─┐
#1256 executable carrier/profile decision ─> #1258 Bytes/headers/URL/errors ──────┘                    │
#31 product-generic compiler slice ─────────┐                                                        │
#1256 bounds/profile ───────────────────────> #1260 Stream[T,E] + Subscription ────────────────────────┘
                                                                                                      │
#1261 request serializer + #1263 body stream ────────────────────────────────────────────────> #1264 vertical/release
```

Implementation lanes:

1. #1257 is closed; its oracle/corpus landed as `task http-client-model`.
[...]

PR #1295, opened 2026-08-11
backlog: record four unnamed blockers so the live gate can land green

hjosugi:
## main is red

The `Backlog issue state` job — which only runs on `main` — is failing on `dc7b49bf`:

```
FAIL: backlog: #30  is blocked but names no blocker where the gate reads one
FAIL: backlog: #31  ...
FAIL: backlog: #274 ...
FAIL: backlog: #644 ...
```

All four carry the `blocked` label with no `Blocked by:` line and no `debt.tsv` row. The rule is not new — it landed in #1060 (`4fb5947b`) — so this is **new input, not a new gate**. `debt.tsv`'s own header names the cause: *"Most of these date from a bulk relabel to `blocked` that did not update the bodies."*

## Why `unnamed-blocker` and not `capability-blocker`

`capability-blocker` asserts that **no issue number could ever express** what the issue waits on. Three of these four contradict that:

- **#644** says it "waits for the smallest product-generic compiler child from **#31 and #1256**" — named blockers, just not on the line the gate reads.
- **#274** routes its external dependency through **#1291**, an issue.
- **#30** and **#31** name nothing either way.

`unnamed-blocker` is the kind that **retires itself**: an issue that gains a `Blocked by:` line stops reaching that branch, and its row then fails as unused. So this records the drift, keeps it countable, and does not assert a permanent exemption on someone else's behalf — which `debt.tsv` explicitly says is a decision per issue rather than something a gate may guess.

I did not relabel any of the four, and I am not editing their bodies. Naming their blockers is the owner's call, and each row disappears when they do it.

## Verified

```
$ GITHUB_TOKEN=$(gh auth token) task backlog-refresh
PASS: 75 of 80 blocked issues name their blockers where the gate reads them;
      4 recorded as unnamed, 1 exempt by design
EXIT=0
```

### DIFF
--- a/tests/backlog/debt.tsv
+++ b/tests/backlog/debt.tsv
@@ -90,3 +90,7 @@
 #                     is worse than none: its author believes the signal is
 #                     published.
 capability-blocker	885	-	Typed upgrade patches slice 1: validate schema and compute
+unnamed-blocker	30	-	[Umbrella] Deliver recursive ownership-aware ADT matching ac
+unnamed-blocker	31	-	[Umbrella] Deliver production generics, traits, KIF, strateg
+unnamed-blocker	274	-	[Umbrella] Close B6 with policy-defined independent clean-bu
+unnamed-blocker	644	-	[Umbrella] Deliver the bounded HTTP/1.1 scripted-transport c


hjosugi:
Superseded by naming the real child blockers directly in #30, #31, #274, and #644. Adding debt.tsv exemptions would preserve stale issue bodies after the dependency DAGs were already designed.

hjosugi:
## Closing: the condition fixed itself at the source while this was in review

Re-measured against pristine `origin/main@dc7b49bf`, with no change from this PR:

```
GITHUB_TOKEN=$(gh auth token) task backlog-refresh   -> EXIT=0
task backlog                                          -> EXIT=0
```

Both checkers are clean, so no debt row is owed and this PR would now fail as *unused* — `debt.tsv` refuses in both directions, and a row recording drift that no longer exists is itself drift.

What happened: all five issues gained proper `Blocked by:` lines while I was preparing this, as part of a restructure that split each umbrella into children.

| issue | now names |
| --- | --- |
| #30 | #1281–#1288 |
| #31 | #1265–#1280 |
| #274 | #1289–#1292 |
| #644 | #1256–#1264, #1269–#1271 |
| #26 | #1293–#1304 |

That is the better outcome by some distance. A `Blocked by:` line says *what* the issue waits on; a debt row only says *that* nobody wrote one. I would rather this PR be unnecessary than merged.

Worth recording for the next person who sees the backlog job red, because it cost me three cycles here:

**The two backlog checkers can disagree, and a debt row has to satisfy both.** `task backlog-refresh` reads live GitHub; `task backlog` inside `verify` reads the committed `artifacts/backlog/issue-state.json` fixture. A row that the live state needs can be *unused* against the fixture, which is exactly how this PR first failed CI:

```
FAIL: backlog: debt names #274 as unnamed-blocker, which no longer applies
```

So adding a debt row while the live backlog is being actively edited is a race you lose: each refresh surfaced a different set, and by the fourth pass every row I had written had retired itself.

Issue #1119, opened 2026-08-09
M4 tracker: the ten 1.0 deliverables and six exit criteria


hjosugi:
## Metadata

- State: blocked
- Tracker role: planning
- Kind: planning
- Area: Release / Project
- Priority: P2
- Roadmap milestone: M4-1.0
- Size: L
- Blocked by: #31, #30, #274
- Audited against: `origin/main@fd680a7e630ad85e51940062601728c4c1a1f6d2`

## Goal

Track the ten M4 deliverables and six exit criteria `docs/ROADMAP.md` defines for 1.0, so the distance to 1.0 is recorded rather than implied. The M4-1.0 milestone currently holds zero open issues while the roadmap names sixteen requirements; that gap is what this umbrella exists to close.

This is a planning umbrella. It is not implementation-ready and must not be marked `ready`.

## Why this is filed as one tracker rather than sixteen issues

`docs/ISSUE_READINESS.md` requires a `ready` issue to carry measured current behavior and an evidence stamp. Most M4 requirements have no executable substrate to measure: there is no proof kernel to audit, no package registry to stabilize, no second platform to release for. Filing sixteen issues whose "Current behavior" sections are empty would manufacture debt rows rather than track work. Each becomes its own issue when its substrate exists, and this tracker records the unlock condition until then.

## Measured distance, 2026-08-14

Re-measured on `origin/main@30f18e4d85dcd23232c46f73fb0d20fac301815e`. The
2026-08-09 numbers below it are kept in the git history rather than here; five
of them had moved, and one of them — "every published release carries zero
assets" — had become the opposite of true.

`release/claims.json` carries 53 published claims:

| State | Count |
|---|---|
| `implemented` | 9 |
| `checkpoint` | 40 |
| `open` | 3 |
| `design` | 1 |

The same four capabilities still have no executable gate:
`general-parser-type-checker`, `general-ownership-checking`,
`general-native-lowering` (all `open`), and `formatter-and-repl`
(`design`). The 40 checkpoints are bounded slices — the compiler accepts the
integer Core, not general Kofun, and seven more checkpoints since 2026-08-09
does not change which four capabilities are ungated.

`tests/conformance/capabilities.tsv` records 17 supported and 23 unsupported
rows across 40 data rows, each unsupported one with a stated reason.

Milestone state: M0-spec 1 open, M1-bootstrap **0 open**, M2-alpha 14 open,
[...]

Issue #1228, opened 2026-08-11
Call arguments v1 slice 3b.4: lower direct-call pipelines through C11 fixed slots


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation child
- Kind: implementation
- Area: Code Generation / Call Arguments
- Priority: P1
- Size: M
- Parent: #882
- Completed prerequisite: #1227, arity/type/take checking landed
- Blocked by: none
- Audited against: `origin/main@037b78cc5456b294e4daf1a15f2555fccb3296ae`

## Goal

Lower one-stage direct-call pipelines through the shared C11 fixed-slot family, evaluating the subject first and exactly once, then explicit arguments once in source order, before ABI-slot invocation.

## Current behavior and evidence

PR #1200 merged as `037b78cc` and provides `fixed_slot_call_shape`, `fixed_slot_call_supported`, `emit_fixed_slot_call`, and `emit_fixed_slot_call_temporaries`. Pipeline syntax has no production lowering and must extend that family, not create a second emitter/walker.

## Scope

- admit `subject |> f()`, positional-rest, and labelled-rest direct top-level calls within the checked carrier matrix;
- evaluate subject first into slot-zero temporary;
- evaluate remaining explicit values once in source order;
- invoke only after all fixed declaration slots are assigned, in ABI declaration order;
- use the existing shared fixed-slot predicate/emitter/walker and temporary namespace;
- execute representative scalar and existing aggregate carriers;
- delete the now-obsolete valid-shape refusal branch/fixture and any helper made dead by admission.

Out of scope: chains, member/bare/indirect targets, trailing lambdas, native/Wasm backends, runtime label structures, allocations, and general emitter refactoring.

## Acceptance criteria

- [ ] Empty-rest, positional-rest, and labelled-rest pipelines build and run on Stage 2/C11.
- [ ] Stdout proves subject first/exactly once, then explicit values once in source order.
- [ ] Call invocation occurs only after every slot assignment and uses ABI declaration order.
- [ ] Original expressions do not reappear in the call expression.
- [ ] The #1200 fixed-slot family is extended; no pipeline-only emitter/walker or temporary family exists.
- [ ] Int and at least one existing aggregate carrier execute with typed slot declarations.
- [ ] Emitted C contains no labels, runtime map/dictionary, string dispatch, allocation, or `|>`.
- [ ] Excluded shapes keep exact E2S158/no-artifact behavior.
[...]

Issue #1227, opened 2026-08-11
Call arguments v1 slice 3b.3: check pipeline arity, type, and take transfer


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation child
- Kind: implementation
- Area: Type System / Call Arguments
- Priority: P1
- Size: M
- Parent: #882
- Completed prerequisite: #1226, slot-zero binding landed
- Blocked by: none
- Audited against: `origin/main@037b78cc5456b294e4daf1a15f2555fccb3296ae`

## Goal

Check effective pipeline arity, each bound slot's declared carrier type, and the bounded RFC-0010 take-transfer rule before lowering.

The checked pipeline still ends at E2S158 and emits no C.

## Current behavior and evidence

`validate_core_calls` counts only parenthesized arguments, so the four canonical pipeline shapes currently report E2S17 or E2S164. No production path checks the subject against slot zero.

## Scope

- effective arity = one subject plus explicit arguments;
- type-check the subject against slot 0 and every explicit value against its bound declaration slot;
- use the subject span as the primary type-error span;
- for a bare binding flowing to a declared `take` slot 0, record exactly one move and preserve existing E2S123 later-use behavior;
- never infer a move from a compound subject;
- keep ordinary positional calls outside the RFC-0010 move path;
- restrict the slice to the current `call_slot_carried` carrier matrix and explicit direct top-level callable;
- refuse zero-parameter, lexical/member/indirect, chain, and other excluded shapes without C.

## Acceptance criteria

- [ ] The four canonical shapes count the subject and reach the lowering boundary instead of E2S17/E2S164.
- [ ] Subject and explicit arguments are checked against bound slots, not source positions.
- [ ] A slot-zero mismatch reports the subject span using the existing type diagnostic.
- [ ] Bare-binding subject to a `take` slot moves once and later use is E2S123.
- [ ] Compound subject creates no synthetic move.
- [ ] Ordinary positional calls retain the existing RFC-0010 boundary.
- [ ] Unsupported target/carrier shapes remain explicit E2S158 refusals and write no C.
- [ ] Mutations excluding the subject from arity or checking against another slot fail the gate.

## Validation

| Check | Command | Expected result |
| --- | --- | --- |
| Call checking | `task call-arguments diagnostics` | arity/type/unsupported cases are exact |
| Ownership | `task affine-resource-handle move-assertion` | bounded take transfer and ordinary-call boundary hold |
[...]

Issue #1226, opened 2026-08-11
Call arguments v1 slice 3b.2: bind the pipeline subject to slot zero


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation child
- Kind: implementation
- Area: Syntax / Call Arguments
- Priority: P1
- Size: M
- Parent: #882
- Completed prerequisite: #1190, recognition landed
- Blocked by: none
- Audited against: `origin/main@037b78cc5456b294e4daf1a15f2555fccb3296ae`

## Goal

Bind a recognized pipeline subject to declaration/ABI slot zero, before binding the explicit call arguments, without using labels for callable selection.

This slice produces checked binding observations and then keeps the lowering boundary at `E2S158`; it emits no C.

## Current behavior and evidence

Production Stage 2 currently initializes every call slot unbound and processes only expressions inside the parentheses. A pipeline subject never reaches `validate_declared_call_arguments`, so the first parameter remains missing.

The accepted call-arguments model already decides that a pipeline subject supplies the first declaration parameter, including an externally labelled first parameter, without spelling that label.

## Scope

- after callable resolution, create one synthetic binding from the subject to parameter/ABI slot 0;
- give the subject semantic `source_index=0`; explicit arguments begin at source index 1;
- start positional explicit binding at slot 1; labelled explicit arguments still map by declaration label;
- treat an explicit argument for slot 0 as a duplicate;
- preserve existing E2S162/E2S163/E2S164/E2S165/E2S166 precedence and related spans;
- keep the post-binding E2S158 lowering boundary and publish no C.

Out of scope: subject type/ownership checking, effective arity, C11 lowering, chains/member/indirect calls, non-C11 backends, and parser refactoring.

## Acceptance criteria

- [ ] The subject binds slot 0 for both unlabelled and externally labelled first parameters.
- [ ] Observation records `slot=0, source_index=0` and the pre-pipe subject span.
- [ ] Positional explicit arguments begin at slot 1; labelled arguments use declaration slots.
- [ ] Explicitly supplying slot 0 again emits E2S163 with the declaration parameter as related span.
- [ ] Existing missing/unknown/duplicate/positional-after-labelled/omitted-slot diagnostics keep their precedence.
- [ ] Labels and pipeline presence do not participate in overload/callable selection.
- [ ] Valid bound shapes stop at E2S158 and write no C artifact.
- [ ] A mutation resetting next positional slot to zero is caught.

## Validation

| Check | Command | Expected result |
[...]

PR #1306, opened 2026-08-11
tests: gate the tests/kofun corpus instead of leaving it unread

hjosugi:
## Eight sources nothing runs, six of which do not compile

`tests/kofun/` holds eight `.kofun` files. **No script, task, or workflow runs them.** `bootstrap/native/README.md` says why:

> The obsolete `tests/kofun/*.kf` acceptance path no longer exists.

The path went; the sources stayed. Measured:

| | |
| --- | --- |
| sources | 8 |
| scripts that run them | **0** |
| that compile today | **2** (`null`, `recursion`) |
| that do not | 6 — they call `assert_eq`, `map`, `is_open`, and bind with `let own` |

## Why not simply repair them

They are not stale tests. They describe the language Kofun intends to become, and **two RFC artifacts cite them as corpus evidence**:

- `rfcs/0004-ownership-kind-classification.md` reads `ownership.kofun` as one of five tracked sources that bind an owned value — *"Each was read."*
- `rfcs/index.json` counts sites in `scientific.kofun`.

Rewriting them until they compile would delete the exact property those records cite, and leave recorded RFC results describing files that no longer say what was measured. **No `.kofun` source is modified here.**

## What the gate does

Each source already declares its contract in the `# expect:` line the obsolete harness read. The gate honours it:

- a source that runs must print that line — `null` prints `9`, `recursion` prints `55`, both verified;
- a source that does not must produce its recorded `.unsupported` diagnostic **on stderr, byte for byte**, and write nothing to stdout.

The second half is load-bearing. When the language grows to accept one of the six, its recorded refusal stops matching and the gate fails *saying so*, naming the golden to delete so the `# expect:` line is checked instead. The corpus stops being a directory nobody reads and becomes a list of boundaries that announce themselves when they move — the same shape `tests/conformance/` already uses.

## Both failure paths are proven, not described

```
changed diagnostic  -> FAIL: tests/kofun: fp refuses differently than recorded
starts compiling    -> FAIL: tests/kofun: fp now compiles and runs. That is
                             progress: delete <golden>
```

The count is pinned at eight, so the corpus cannot quietly shrink to nothing and still pass. The split is expected to move toward `executed`, and moving it is what the recorded refusals force someone to do deliberately.

`.unsupported` is deliberately not `.stdout`/`.stderr`, so the stage2-events companion census does not count these. Registered in `Taskfile.yml` **and** `tooling/task-help.mjs`.

Refs #1119.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -712,6 +712,11 @@ tasks:
     cmds:
       - "sh tests/digest/check.sh"
 
+  tests-kofun:
+    desc: "Verify the tests/kofun corpus runs or refuses as recorded"
+    cmds:
+      - "sh tests/kofun/check.sh"
+
   call-arguments-spec:
     desc: "Verify labelled call binding and the single trailing-lambda form"
     cmds:
@@ -1011,6 +1016,7 @@ tasks:
             artifact-qualification \
             syntax \
             digest \
+            tests-kofun \
             call-arguments \
             call-arguments-spec \
             call-arguments-surface \
--- a/artifacts/release-evidence/index.json
+++ b/artifacts/release-evidence/index.json
@@ -705,7 +705,7 @@
     }
   ],
   "evidence_digests": {
-    "Taskfile.yml": "ca430ff9deb20ccef967aaa7ee6ac613dc0d46c4612e73ac52e2886e5c515b93",
+    "Taskfile.yml": "28bf1b2af6a03f045fd5f6079e7750e5532345ba71a62fa0e5545249e1b756d5",
     "bootstrap/native/check.sh": "dfa876ae496b87ce1a3f9535d09a5f54d650a1efbe747ddc527a78f2114b4fae",
     "bootstrap/native/fixtures/function_all_traps_pressure.kofun": "8ecd1a2a6956db3812210a60b2f027bee406777db14b2faea3fc7008c280528a",
     "bootstrap/selfhost/check-diverse-double-compilation-refusals.sh": "55fc39166070b9bc612fca6898086938b1a8ca122b1c1fb8e928f1a5d4cde572",
--- a/tests/kofun/check.sh
+++ b/tests/kofun/check.sh
@@ -1,0 +1,95 @@
+#!/bin/sh
+set -eu
+
+# `tests/kofun/` was written for an acceptance path that no longer exists --
+# `bootstrap/native/README.md` says so in as many words -- and nothing has run
+# these eight sources since. Six of them do not compile: they call `assert_eq`,
+# `map`, `is_open`, and bind with `let own`, none of which the language has.
+#
+# They are not stale tests. They are sources describing the language Kofun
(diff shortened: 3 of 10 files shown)


hjosugi:
## Strengthening the central claim's method

This PR asserts that no script, task, or workflow runs `tests/kofun/`. I first established that by grepping for the directory string — which does **not** establish it: a gate could glob `*.kofun` broadly and pick the files up without ever naming the directory.

Re-checked properly, by enumerating every place in the repository that globs `.kofun` files at all and asking whether each could reach the corpus:

| site | scope | reaches `tests/kofun/`? |
| --- | --- | --- |
| `examples/check.sh` | `(cd "$EXAMPLES" && find . -name '*.kofun')` | no — the `cd` bounds it |
| `stdlib/tests/verify.sh` | `find "$stdlib_dir/linux_x86_64"` | no |
| `tests/conformance/optional-*/run.sh` | `find "$CASES"` | no |
| `bootstrap/selfhost/check-declared-inputs.sh` | `ls bootstrap/selfhost/driver/corpus_*.kofun` | no |
| `tooling/kotest/run.sh` | `find "$path"` — explicit PATH arguments | no; `tests/stdlib/kotest/check.sh` passes its own fixtures |

`examples/check.sh` was the one that could have invalidated the claim, since `find . -name '*.kofun'` reaches everything below the working directory — but it runs inside `cd "$EXAMPLES"`.

So the conclusion holds, and now by a method that supports it. Recording this because the weaker method was in my working notes and would have looked equally confident if it had happened to be wrong.

hjosugi:
## Proven in both directions on the day it was written

**It fires when a boundary genuinely moves.** #1307 taught the front end to recognise pipeline chains, and rebasing onto it produced:

```
FAIL: tests/kofun: fp refuses differently than recorded
--- recorded
error[E2S16]:  unknown Core function `map` at byte 56
--- actual
error[E2S158]: a pipeline chain is specified by call-arguments v1 but not recognized at byte 53
```

Re-recorded all six refusals, and exactly one moved — `fp` is the only source in the corpus containing `|>`.

**It will not fire for #1226, #1227, or #1228** — and that is the more interesting half, because a gate that fired on every neighbouring change would be noise rather than signal.

The reason is structural: #1190 split pipeline refusal into an **early shape pass** (`validate_pipeline_shapes`, called at `compiler.kofun:2692`, before scope construction) and a **late production pass**. `fp.kofun` is a chain, so it is refused by the early pass and never reaches the binder, checker, or emitter. The three successor slices only touch the recognised production, so the chain stays on the early path throughout.

So the gate distinguishes those two paths **without being told they exist** — it pins observed behaviour, and the split falls out of that. If `fp.unsupported` does move under one of those three, it means the early path was changed by accident, which is exactly the signal worth having.

That also narrows the obligation this gate creates on neighbouring work: **a slice owes a golden update only if it touches the path its fixture is on**, not merely because it touches the same feature area.

Issue #1348, opened 2026-08-12
RFC-0013 step 1: implement the eight Int bit operations on the Stage 2 pair


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: implementation
- Area: Language / Stage 2 compiler
- Priority: P1
- Size: L
- Decision: PR #1351 merged as `ce416fb50f84e95a54bac7839cdc97de88697698`, but the published P1/P2 review remains unresolved; fix-forward required
- Blocked by: none
- Audited against: `origin/main@ce416fb50f84e95a54bac7839cdc97de88697698`

## Goal

Implement RFC-0013's eight `Int` bit operations on the canonical Stage 2 pair,
with C11 lowering, so the language can express a 32-bit SHA-256 round.

This is RFC-0013 step 1. It does not write SHA-256, touch the
native/AArch64/wasm32 backends, or publish a capability.

PR #1351 merged the reviewed-NO-GO head despite the published P1/P2 findings. This issue remains open for a fix-forward that closes the non-Int dispatch, unknown-chain, diagnostic-order/span, registry, boundary-evidence, and truth-restamp gaps, followed by focused/full validation and independent review.

## Current behavior and evidence

Measured on `80f79798361c9513a637c521ec03d718d3842dfc`, across 1114 tracked
`.kofun` sources:

```sh
git ls-files '*.kofun' | wc -l
git grep -nE '^[[:space:]]*fn (and|or|xor|not|shl|shr|rotr|wrapping_add)\(' -- '*.kofun'
git grep -nE '\.(and|or|xor|not|shl|shr|rotr|wrapping_add)\(' -- '*.kofun'
```

The inventory is 1114; both collision queries exit 1 with no output. The
language has no direct bit primitive or tracked in-language SHA-256
implementation. It does have tracked digest helpers, so those are not claimed
absent.

The canonical compiler has no general `Int` method-dispatch path. Host builtins
are prefix functions; `.join()`/`.spawn()` are a par-handle special case.
`postfix = primary, { call | member | index }` admits the syntax, but current
main does not type or lower these eight methods for an `Int` receiver.

## Frozen implementation boundary

In scope:

- consume `<Int expression>.name(args)` on both canonical Stage 2 surfaces;
- type and lower `and`, `or`, `xor`, `not`, `shl`, `shr`, `rotr`, and
  `wrapping_add`, with `Int` receiver/arguments/result;
[...]

Issue #1220, opened 2026-08-11
Scoped parallelism HIR: emit par, spawn, and join identities


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: implementation
- Area: Concurrency / Compiler
- Priority: P2
- Size: M
- Parent: #1161
- Blocked by: none
- Audited against: `origin/main@bbbdc439bd277176d1efdab1d288a15eecd8186c`

## Goal

Emit canonical scoped-parallel identity records from the production Stage 2
resolver through an analysis-only CLI. The closed document and all derived
IDs obey `spec/concurrency/scoped-captures-v1.md`; ordinary compilation still
refuses `par` with E2S154.

## Before implementation (measured 2026-09-23)

Historical baseline measured in a clean worktree at `04003d9b153a09967723a0313016b7a78e55d2cd`:

```sh
cc -std=c11 -O1 -w bootstrap/stage2/compiler.c -o /tmp/kofun-par-main
/tmp/kofun-par-main --emit-scope-hir tests/diagnostics/stage2/e2s154_scoped_parallelism.kofun /tmp/par.scope-hir
# exit 1, stdout: error[E2S154]: scoped parallelism `par` is specified but not implemented at byte 88
# empty stderr; output absent
```

The four existing v1 unsupported shapes also return exit 1 and no HIR:
`e2s154_par_dynamic_scope_token` at byte 688, `e2s154_join_alias` at 627,
`e2s154_par_nested_scope` at 324, and `e2s154_par_spawn_in_loop` at 526.
Malformed missing-token/block fixtures likewise retain their current refusals.
The standalone spawn/join-member fixtures can produce lexical v1 HIR; their
ordinary compilation refusal is a later boundary, not a reason to claim every
E2S154 fixture fails at the scope-HIR walk.

SHA-256 is already in the canonical pair (#1382), and the current driver has
`stage2_unicode_scalar_at`. The work required at that baseline was resolver integration, streaming
framing, and canonical v2 output. The completed implementation supplies these
without a host digest or generic JSON library.

## Frozen implementation profile

- Add `kofun-stage2 --emit-scope-hir-v2 INPUT OUTPUT LOGICAL-PATH`. The physical
  input/output paths are I/O locations only. `LOGICAL-PATH` is explicit,
  1–4,096 bytes of valid UTF-8 already in NFC, relative, with `/` separators.
[...]

PR #1349, opened 2026-08-12
rfcs: RFC-0013, bit operations on `Int` are named methods

hjosugi:
Kofun cannot compute a digest, and not because nobody wrote the code.

The language's entire operator set is `+ - * / // % **` and unary `+ - !`
(`spec/grammar.ebnf:121-124`). There is no `&`, `|`, `^`, `<<`, or `>>`, and no
way to obtain one — a bit operation has to be built from `//`, `%`, and `**`,
one bit at a time. So the repository's own SHA-256 is in C, and not one of the
1105 tracked `.kofun` sources implements a digest, because none can.

That is load-bearing. Every identity the compiler assigns is a SHA-256
preimage — `ModuleId`, `FileId`, `SymbolId`, and the scope-HIR v2
`ParId`/`TaskId`/`JoinId` frozen in `spec/concurrency/scoped-captures-v1.md`.
**The compiler cannot compute the identities it is specified to assign**, which
is where #1220 stopped.

## What this proposes

Eight operations on `Int`, spelled as postfix methods: `and`, `or`, `xor`,
`not`, `shl`, `shr`, `rotr`, `wrapping_add`. No grammar change — `postfix =
primary, { call | member | index }` already admits the form and `handle.join()`
already uses it. No reserved word: the eight stay ordinary identifiers
everywhere, exactly as `join` does.

Semantics are fixed rather than left to the implementation: two's complement,
arithmetic `shr`, shift counts `0..63` and widths `1..64` trapping as a new
`R011`, `shl` trapping as the existing `R010`.

## Why no operator symbols

`&`, `^`, `<<` and `>>` are free. `|` is not — it delimits the `par` scope
token and separates ADT variants — so an operator set could cover four of the
five and would leave `a & b` next to `a.or(b)` in every dense expression.

Haskell met the same collision and answered with `.&.` and `.|.`. Rust kept
single `|` only because its enums use commas, and still pays for the token
meaning three things by position. And symbols would not finish the job: Rust
has every operator and still spells `rotate_right` and `wrapping_add` as
methods, because checked arithmetic needs a named escape — and Kofun's `+`
traps, so modular addition needs a name here too.

Symbols stay available as a later amendment defined as sugar for the methods,
after there is an implementation to define them against.

## Corpus evidence

`1105` tracked sources, `0` declaring any of the eight names, `0` calling a
member with them, at `ffa3d8a7`. No spelling collides and nothing acquires new
behaviour.

## Two questions left open

Whether `shl` should trap or wrap is deferred to writing the SHA-256 and seeing
which form the code wants. Whether `width` belongs on the operation or in a
sized type is a decision about Kofun's numeric tower that this does not open.

## Plan

Four steps, tracked separately. Step 1 is #1348. Step 2 — a Kofun SHA-256 gated
against the NIST vectors *and* differentially against `bootstrap/stage2/sha256.c` —
is deliberately a standalone gate before step 4, because a wrong hash produces
stable, self-consistent, entirely incorrect identities.

`task rfc-registry` passes: 38 decisions, 31 accepted.

### DIFF
--- a/rfcs/0013-int-bit-operations.md
+++ b/rfcs/0013-int-bit-operations.md
@@ -1,0 +1,239 @@
+# RFC-0013: Bit operations on `Int` are named methods, and the operator symbols stay free
+
+- Shepherd: hjosugi
+- Opened: 2026-08-12
+- Status: accepted
+
+## Summary
+
+`Int` gains eight bit operations, spelled as postfix methods: `and`, `or`,
+`xor`, `not`, `shl`, `shr`, `rotr`, and `wrapping_add`. No operator symbol is
+added, no keyword is reserved, and no existing program changes meaning. Someone
+writing Kofun can express a 32-bit hash round for the first time.
+
+## Motivation
+
+Kofun cannot compute a digest, and the reason is not that nobody has written the
+code.
+
+The language's entire operator set is `+ - * / // % **` and unary `+ - !`
+(`spec/grammar.ebnf:121-124`). There is no `&`, `|`, `^`, `<<`, or `>>`, and no
+way to obtain one — a bit operation has to be built from `//`, `%`, and `**`,
+one bit at a time. SHA-256 performs roughly sixty-four rounds of six such
+operations over 32-bit words; expressed arithmetically that is thousands of
+loop iterations per block, in a bounded Core, written twice because the Stage 2
+compiler is a hand-transliterated pair.
+
+So the repository's own SHA-256 is in C. `bin/kofun-digest` is a shell script
+that builds `bootstrap/stage2/sha256_tool.c` against `bootstrap/stage2/sha256.c`,
+and its header says why: to stop depending on GNU `sha256sum`. Not one of the
+1105 tracked `.kofun` sources implements a digest, because none can.
+
+That is load-bearing rather than cosmetic. Every identity the compiler assigns
+is a SHA-256 preimage — `ModuleId`, `FileId`, `SymbolId`, and the scope-HIR v2
+`ParId`/`TaskId`/`JoinId` frozen in
+`spec/concurrency/scoped-captures-v1.md`. **The compiler cannot compute the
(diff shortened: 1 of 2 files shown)


hjosugi:
Adversarial review: **do not merge the current head**. P1=0, P2=4.

1. **Semantics conflict / missing cases.** The RFC normatively makes `shl` trap with R010, but later calls trap-vs-wrap unresolved. Keep trapping as the settled `shl` rule and move the open question to a possible separate `wrapping_shl`. Also define `rotr` for `n >= width` and define the signed two's-complement representative returned for width-64 results (or narrow the supported widths).
2. **Diagnostics are not implementable from the text.** Pin exact code, message, and offending token/span for invalid shift/width and for unknown method, non-Int receiver, wrong arity, and wrong argument type. Distinguish the R011 count and width messages and add these negatives to Validation.
3. **Evidence/provenance prose is inaccurate.** The tree already has in-language digest functions, so say “no tracked in-language SHA-256 implementation/direct bit primitive,” and add a SHA-256-specific inventory query. Say “bitwise primitives,” not the entire operator set. ModuleId/FileId/SymbolId/ParId/TaskId/JoinId are SHA-256 outputs over framed preimages, not preimages themselves; correct both RFC and index wording.
4. **Accepted-state evidence is stale.** Add the ledger's Review closed / Decided dates and remeasure the exact current base: the corpus is 1114 (not 1105), with 0 declarations / 0 member calls under the stated queries. Rebase/amend so the accepted bytes enter current main consistently.

Remote integrity itself is healthy: head `56a1f8c2c0037de6b96c48101865ea77b9b562be`, base/main `833cc752923b910225108e7a0fe115632e45a25e`, one commit/two files, MERGEABLE/CLEAN, required verification and release-evidence checks green.

Correct dependency order: fix and merge this RFC first, then re-audit/rebase #1348 against that exact main before continuing implementation.

Issue #1350, opened 2026-08-12
rfcs: close RFC-0013 bit-operation semantics and evidence truth


hjosugi:
## Metadata

- State: completed
- Tracker role: curated correctness fix
- Kind: spec-design
- Area: Language / Stage 2 compiler
- Priority: P0
- Size: S
- Parent: none
- Blocked by: none
- Blocks: #1348 and the #1205/release close path
- Audited against: `origin/main@80f79798361c9513a637c521ec03d718d3842dfc`

## Goal

Record `RFC-0013/A01`, a process-compliant amendment that closes the accepted
RFC's contradictory or missing bit-operation semantics, diagnostic contract,
and repository evidence before #1348 may continue.

The accepted `rfcs/0013-int-bit-operations.md` is immutable. Preserve every
original word and add only the required amendment announcement; put the
superseded wording verbatim, current delta, compatibility analysis, evidence,
date, and #1350 change pointer in `rfcs/index.json`.

This issue does not implement an operation, change a compiler, publish a
capability, or choose a second numeric type.

## Reproduced current behavior

At the audited main, accepted RFC-0013 has these correctness defects:

1. Lines 118-121 settle trapping `shl`, while 230-234 call trap versus wrap
   unresolved. `rotr(n, width)` leaves `n >= width` and the signed
   representative of width-64 results unspecified. The overflow analogy also
   names `a * 2**n` as existing behavior even though Stage 2 does not fully
   parse/lower `**`; the rule must instead use mathematical `a × 2^n` and the
   signed-Int range.
2. Lines 130-141 do not give implementable exact diagnostics and precedence for
   runtime count/width failures or compile-time member dispatch. Runtime
   diagnostics use the existing generated-C ABI and have no source span;
   compile-time E2S diagnostics do have byte anchors. The accepted labelled-call
   grammar also admits `1.and(value: 2)`, but the RFC says nothing about labels,
   trailing lambdas, or their precedence.
3. Lines 16-37 deny tracked digest functions, misclassify unary Boolean `!` as
   arithmetic by implication, call a subset the entire operator set, and call
   identity digests preimages. The current provenance is also split: module
   identity tools/producers use C SHA-256, while the scoped-captures reference
   model computes Par/Task/Join IDs with `node:crypto`; compiler emission of the
   latter remains unimplemented.
4. The accepted document does not surface the ledger review/decision dates,
   and both RFC and ledger report the pre-merge `ffa3d8a7` corpus count 1105.
[...]

PR #1351, opened 2026-08-12
Int gains eight bit operations, spelled as postfix methods

### LINKED ISSUE #1348: RFC-0013 step 1: implement the eight Int bit operations on the Stage 2 pair

hjosugi:
RFC-0013 step 1. `Int` gains eight bit operations — `and`, `or`, `xor`, `not`,
`shl`, `shr`, `rotr`, `wrapping_add` — spelled as postfix methods, on the
canonical Stage 2 pair with C11 lowering.

Before this, the language's entire operator set was `+ - * / // % **` and unary
`+ - !`. There was no way to express a bit operation, which is why the
repository's own SHA-256 is in C and why no tracked `.kofun` source implements a
digest. `./bin/kofun run` now executes a 32-bit hash round.

## What is here

- **Parsing.** `field_postfix_end` is the single point every primary form
  reaches, so one insertion there gives every receiver: literal, binding, record
  field, call result, parenthesised expression, and chains of all of them.
  `(a + b).and(255)` and `state.h0.rotr(4, 32).xor(mask)` both parse.
- **Lowering.** Each operation becomes one call on a checked runtime helper,
  never a bare C operator. That is not indirection for its own sake — see below.
- **Refusals.** `E2S168` for a non-`Int` receiver or argument, `E2S169` for a
  wrong argument count, `R011` for a shift count outside `0..63` or a width
  outside `1..64`, and the existing `R010` for `shl` overflow.
- **The names stay free.** `fn and()`, `fn xor()`, and `type P = { xor: Int }`
  all still work, in the same file that uses the operations. A name is
  meaningful only as a member of an `Int` receiver, exactly as `join` is.

## Why the operations do not lower to C operators

C will not answer the questions these operations ask. A shift by the operand
width is undefined; a signed right shift is implementation-defined; and
converting a bit pattern above `INT64_MAX` back to a signed type is
implementation-defined as well. The whole reason for naming these operations was
to give them answers the language states, so none of the three is left to the
host compiler — `kofun_from_bits` computes an out-of-range pattern from its
complement rather than casting it.

## Two defects this found, both mine, both silent

Writing the gate found them; running the compiler did not.

**`to_text`, `len` and `text_slice` never checked their argument.** Each appended
the result of `emit_expression` without inspecting it, so a refused argument
became C source: the diagnostic text was emitted *inside* the call, the compiler
exited 0, and only `cc` reported it. The general call path already carries this
check with a comment describing the same defect; these three direct branches
never had it. Pre-existing, and reachable for the first time here because
`E2S168`/`E2S169` are the first refusals an argument of `to_text` can raise.

**A wrong receiver was accepted when nothing read the result.** `let masked =
value.and(1)` with a `Float` `value` compiled, emitting
`double k_b1 = kofun_bit_and(k_b0, INT64_C(1));` — valid C, silently wrong
arithmetic. Two fixtures pin it now: `float_receiver.kofun` never reads its
result, and `float_receiver_used.kofun` does, and both must name the receiver
rather than blaming `to_text`.

I also removed a rule I had added to `numeric_primary_type` before shipping it: a
differential run over eight programs showed it changed nothing observable, and
the bounded variant I had introduced existed only to repair it. Both are gone;
`numeric_primary_type` is untouched apart from a comment.

## Validation

`task int-bits` is new. It is written so a wrong implementation produces a
different number rather than a crash: the 26 observations in `behavior.kofun`
include `(-8).shr(1) == -4`, `(1).rotr(1, 64) == INT64_MIN`, and SHA-256's
σ₀(1) = `0x02004000`, checked against its published value.

**Fifteen reintroduced defects, fifteen refusals, no survivors** — each of the
eight helpers' checks removed one at a time, the arity check, both operand-type
checks, the initializer rule, the chain parser stopping after one call, and the
eight names becoming reserved.

`task stage2`, `task selfhost-fixed-point` (C2 == C3 and A2 == A3 hold),
`task diagnostics`, `task int-bits`, and full `task verify` pass.

`bootstrap/stage2/SHA256SUMS`, `bootstrap/manifest.json`'s trusted seed digest,
and the `stage2-events` census (350 → 355, its five new refusal companions) move
with the pair, as they do for any Stage 2 change.

## Scope

The C11 Core lowering is what `./bin/kofun` uses, so this is the path a Kofun
program takes. The **selfhost** lowering — the separate, narrower path by which
the compiler compiles itself — does not have these operations, and does not need
them until RFC-0013 step 4. The native, AArch64 and wasm32 backends are step 3;
`int-bits` is a Stage 2 gate and declares no backend capability.

Closes #1348.


hjosugi:
**Do not merge this head.** Issue #1348 is explicitly `blocked` by #1350, and #1350 has not merged.

This PR was opened against the contradictory RFC-0013 bytes that #1350 corrects. Its body repeats two known false claims (“entire operator set” and no tracked Kofun digest), and the implementation must be reconciled with the corrected accepted contract before review. In particular, #1350 settles modulo-width `rotr`, signed width-64 results, exact runtime messages/precedence, and the `E2S168` unknown-`Int`-member boundary; this head was not implemented or reviewed against those final bytes.

Required order:

1. Merge #1350.
2. Rebase this branch onto that exact main.
3. Restamp #1348 from `blocked` to `ready`, reconcile every corrected semantic/diagnostic boundary, and rerun focused/full validation.
4. Only then restore `pr-open` and request merge review.

The current CI run cannot substitute for this dependency: it tests the pre-#1350 contract. The issue claim is being returned to canonical `active` while blocked; ownership remains with the same agent.

hjosugi:
Rebased onto `9410d1e0` and reconciled against the merged `RFC-0013/A01`.

A01 pinned exact diagnostic text, and two things did not match. Both are fixed
here rather than argued about, since the amendment is the authority.

**The trailing-lambda message** now reads `takes no trailing lambda; the bit
operations have no functional parameter`, A01’s wording.

**`unknown \`Int\` member \`<name>\`` was a genuine gap, not a wording change.**
Before it, the answer depended on where the call sat:

| position | before | now |
|---|---|---|
| `let` / `return` | `E2S32: unknown nominal record field read` | `E2S168: unknown \`Int\` member \`frobnicate\`` at the member |
| inside another call | `E2S17: to_text expects 1 arguments, got -1` | same, at the member |

Both were the same mistake — blaming a construct the author did not get wrong.
One sent the reader to a record they never wrote; the other to an unrelated
`to_text`. Reaching the second required the postfix reader to consume any member
call, not only a table name, so that an unknown member stops derailing the
enclosing argument count. Meaning stays in the lowerer; the reader only decides
where the expression ends. `records`, `scoped-parallelism` and `list-int-values`
all pass, and `.spawn`/`.join` keep their `E2S154`.

The rebase brought in #1355. Only four files conflicted, all of them generated
or hand-counted, and none was hand-resolved: the three digests were regenerated
after the rebase, and the `stage2-events` census was recomputed by the gate’s
own predicate rather than added up — 351 on main plus this branch’s eight
refusal fixtures is 359.

`task verify` on the rebased tree: exit 0, 1369 PASS, 0 FAIL.

hjosugi:
**Do not merge current exact head `bcb0ef0ae0febd22e040979ca7d5e28e4c8aca9b`.** Two independent static audits of this rebased SHA agree: **P1=2 / P2=4 / P3=1**. The detailed issue review is https://github.com/kofun-lang/kofun/issues/1348#issuecomment-5262595059.

The reconciliation comment does not address the blockers: Bool/List/record/enum/non-Int receivers and arguments can still bypass E2S168; `1.and(1).frobnicate(2)` can silently discard the unknown suffix; A01 label/trailing anchors and receiver→LTR(label then type)→trailing→arity order are still wrong; R010 producers are incomplete in the registry; dual-invalid/count-before-overflow/width64 boundaries are ungated; the false `a * 2**n` oracle remains; and “15 defects / 15 refusals” still has no tracked mutation harness. The PR body also repeats the known false operator/digest/current-census wording.

CI/full green does not cover those missing mixed cases. Return to draft, amend on a new SHA, rerun focused/full validation, and request another independent review. Ownership remains with the existing claim.

Issue #1352, opened 2026-08-12
RFC-0013 step 2: write SHA-256 in Kofun, gated against NIST vectors and sha256.c


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation child
- Kind: implementation
- Area: Language / stdlib
- Priority: P1
- Size: M
- Decision: RFC-0013/A01 is merged, and #1348's fix-forward landed as PR #1370, so the review findings that held this child are resolved
- Blocked by: none
- Audited against: `origin/main@ce416fb50f84e95a54bac7839cdc97de88697698`

## Goal

After #1348's reviewed fix-forward merges, implement SHA-256 in Kofun and gate it against
published NIST vectors and differentially against
`bootstrap/stage2/sha256.c` on identical bounded inputs.

This is RFC-0013 step 2. It precedes compiler identity adoption because a wrong
hash produces stable, self-consistent, incorrect identities. It does not replace
`bin/kofun-digest`, implement compiler IDs, add backend lowering, or publish a
capability.

## Current behavior and measured boundary

At the audited main, the repository has tracked non-cryptographic digest helpers
but no tracked in-language SHA-256 round implementation or direct bit primitive.
Module identity paths use the C SHA-256 implementation. The scoped-captures
Par/Task/Join reference model currently uses `node:crypto`; compiler emission of
those IDs remains unimplemented. The IDs are digest outputs over framed
preimages, not preimages themselves.

Base-pinned evidence:

```sh
base=80f79798361c9513a637c521ec03d718d3842dfc
git grep -nEi 'sha.?256|0x428a2f98|0x71374491|1116352408|1899447441' \
  "$base" -- '*.kofun'
git grep -nE 'fn (tzdb_digest|digest_bytes|content_digest)' "$base" -- '*.kofun'
```

The SHA query returns eleven comment/help/accessor lines and no first-round
constant anchor; the second query proves tracked bounded digest functions exist.

The implementation is not authorized to proceed from the rejected #1348 bytes now present on main. #1348 must first merge a reviewed fix-forward for its exact bit-operation contract.

## Exact scope after promotion

In scope:

- SHA-256 over a bounded byte input in Kofun on the C11 Core lowering;
- a lasting `task kofun-digest-model` gate for published NIST vectors;
- padding boundaries at lengths 55, 56, 63, 64, and 65 bytes;
- byte-identical differential comparison with `bootstrap/stage2/sha256.c` over
[...]

PR #1355, opened 2026-08-12
Stage 2 C11: a List[Int] binding is reassignable

### LINKED ISSUE #1353: Stage 2 C11: reassigning a List[Int] local from a call emits invalid C
## Metadata

- State: in-progress
- Kind: implementation
- Area: Stage 2 compiler / C11 Core lowering
- Priority: P1
- Size: S
- Blocked by: none

## Goal

Reassigning a `List[Int]` local from a function call must either lower correctly
or be refused. Today it emits invalid C, the compiler exits 0, and only `cc`
reports the failure.

```kofun
fn bump(xs: List[Int]) -> List[Int] {
    let mut ys: List[Int] = [0, 0]
    ys[0] = xs[0] + 1
    ys[1] = xs[1] + 1
    return ys
}

fn main() -> Int {
    let start: List[Int] = [10, 20]
    let mut cur = start
    cur = bump(cur)
    print(to_text(cur[0]))
    return 0
}
```

```
$ ./bin/kofun run repro.kofun
...
error: incompatible types when initializing type 'int64_t' using type 'KofunIntListValue'
  int64_t kofun_replacement = ((kofun_call_arg_0 = k_b3), kofun_fn_bump(kofun_call_arg_0));
error: incompatible types when assigning to type 'KofunIntListValue' from type 'int64_t'
  k_b3 = kofun_replacement;
```

The assignment path names its replacement temporary `int64_t` unconditionally.
Every other `List[Int]` position — parameter, return, `let` initializer,
subscript, and subscript assignment — is lowered correctly; only reassignment
of the whole binding from a call is wrong.

Measured on `80f79798` and on the pre-#1348 tree: the defect predates RFC-0013
and is not related to it.

This is the same failure class as the `to_text`/`len`/`text_slice` argument leak
closed in #1348: the compiler accepts a program it cannot lower, exits 0, and
leaves `cc` to report a diagnostic that names C rather than Kofun. A reader who
sees `KofunIntListValue` in a compiler error has been shown the lowering's
internals for a mistake the frontend could have named.

It also has a concrete consumer. #1352 writes SHA-256 in Kofun, whose natural
driver is

```kofun
let mut state = sha256_initial_state()
while block_index < blocks {
    state = sha256_absorb(state, next_block(...))
    ...
}
```

which is exactly this shape. #1352 works around it with single assignment,
which bounds it to a fixed block count.

## Acceptance criteria

- [ ] The repro above compiles and prints `11`.
- [ ] The replacement temporary takes the binding's own C type, so a `List[Int]`
      reassignment from a call, from another binding, and from a literal all
      lower.
- [ ] If any of those is outside this slice, it is refused with a named Stage 2
      diagnostic and no C artifact — not emitted for `cc` to reject.
- [ ] A gate covers both twins, and reintroducing the `int64_t` temporary fails
      it.

## Validation

| --- | --- | --- |
| Repro | `./bin/kofun run repro.kofun` | prints `11` |
| Slice | `task list-int-values` | no regression |

## Partial merge and remaining fix-forward

PR #1355 merged unchanged reviewed head `7befeaf84c28`
as main commit `9410d1e0701257403dcf31098ef384c98929130e` despite the published
P2=6/P3=1 do-not-merge review. The issue is reopened because its acceptance
criteria are not met.

The landed change covers ordinary binding/call/loop RHS forms, but not the full
bounded assignment surface. A `List[Int]` target with an `if` or `match` RHS
still enters the older value-control branch and emits an `int64_t` replacement,
so accepted Kofun can still become invalid C. The fix-forward must also add the
exact print-11 repro and literal form, exercise the new emitter-error propagation
branch, prove both compiler twins with independent mutations, make the Kofun
branch mutually exclusive, update release/readme truth, and regenerate evidence.

## Scope

In scope: close those exact remaining whole-binding replacement/refusal and
pair-evidence gaps on current main. Out of scope: other container types,
general value-control lowering, unrelated Stage 2 surfaces, capability
promotion, or VERSION.

## Dependencies

- Blocked by: none.
- Base the fix-forward on exact current main `ce416fb50f84e95a54bac7839cdc97de88697698` (which contains both the partial #1353 landing and the unresolved #1348 landing).
- Keep this issue open and the owner claim active until focused/full validation,
  independent review, and a reviewed follow-up PR are merged.

hjosugi:
The replacement carrier the assignment statement declares was `int64_t` whatever
the binding was, so `current = step(current)` on a `List[Int]` emitted

```c
int64_t kofun_replacement = <KofunIntListValue>;
```

The compiler exited 0 and only `cc` reported it, naming the lowering's own
internals for a mistake the frontend could have named. The same statement also
appended a refused value without inspecting it, so an `error[...]` string could
become C the same way — the third instance of that class this week, after the
general call path and after `to_text`/`len`/`text_slice`.

## Extent, measured before fixing

Only `List[Int]` leaks. `Int` reassignment is correct, and `Text`, `Decimal`,
`DecimalResult` and `Float` are already refused as `E2S144` before reaching
here. So this is one type, not a lattice, and the change is that whole gap and
no more.

## The fix is to stop deriving it twice

The `let` path immediately above already derives both the C type and the value
emitter from the binding's type. This derives them the same way rather than a
second way, which is what let the two disagree in the first place.

A `List[Int]` binding now also refuses a value of another type, because the
carrier is a struct that cannot hold one.

**That check is scoped to `List[Int]` deliberately.** Applied to every
assignment — which is what I wrote first — it restated rules other slices own
and refused `seen = x + 1` for `seen: Int` and a narrowed `x: Int?`. The
optional-construction gate caught it. Worth recording: the four reintroduced
defects below all failed the gate, so the implementation looked protected; what
it broke was outside the range I thought I was protecting, and somebody else's
gate found it.

## The loop is the point

A single reassignment can be written around by binding a new name each time; a
loop cannot, and the loop is the shape a digest driver needs — `state =
absorb(state, block)` is exactly this (#1352).

```
[1, 1] → step ×3 → [4, 8]
```

## Validation

`task list-int-values` gains a `reassignment` fixture covering assignment from a
binding, from a call, and in a loop, plus a `reassignment_type_mismatch`
refusal. Four reintroduced defects, four refusals: the carrier reverting to
`int64_t`, the value emitter not switching, the type check removed, and the
refused-value propagation removed.

`task verify`: exit 0, 1356 PASS, 0 FAIL.

Closes #1353.


hjosugi:
**Do not merge current head `7befeaf84c28d31854e1eee59a0cf6c086a50e47`.** Independent static review completed immediately before this PR opened and found P1=0 / P2=6 / P3=1; the full findings are on #1353: https://github.com/kofun-lang/kofun/issues/1353#issuecomment-5262396272

The blocking examples include a `List[Int]` target with `if`/`match` RHS still bypassing the new dispatch and emitting an `int64_t` carrier, missing exact repro/literal coverage, a negative that never reaches the new emitter-error propagation branch, no semantic twin/mutation proof for `compiler.kofun`, stale release truth, and incomplete issue scope/dependencies. The branch also has not been rebased onto current exact main `0e35375a4623e260090bc3873facabbe382e943e`.

Please return this PR to draft, amend the implementation/tests/docs/evidence and issue body, rebase on exact current main, and rerun focused/full validation before requesting another review. The existing owner claim remains unchanged.

Issue #1311, opened 2026-08-11
Benchmark report v1: implement the bounded model, validation, and segmented summaries


hjosugi:
## Metadata

- State: ready
- Parent: #847
- Kind: implementation
- Area: Core Standard Library / Benchmarking
- Priority: P2
- Size: M
- Blocked by: none
- Audited baseline: `origin/main@eb24df50f9f88a9a4fd01e3787806c63aaf6e894`
- Completed foundation: #1310 via PR #1318

## Goal

Implement #1310's exact 49-field bounded report outcome in Kofun, including
closed validation, raw-order retention, and deterministic 100-sample summaries
using only the existing capacity-64 `List[Int]` profile.

This child owns no Bytes, codec, filesystem publication, or threshold
comparison and has no remaining carrier blocker; #1310's contract is merged.

## Current behavior and evidence

At the stamped baseline `task benchmark-summary` implements only the fixed
`Samples8` evidence. `task aggregate-bridge records list-int-values
text-results` proves a flat report-shaped record and whole-record pass/return,
but a direct `List[Int]` literal in a record field emits a pointer/value C type
mismatch. Binding each list to a typed local before record construction compiles
and runs under strict C11. #1310 freezes four list fields, exact 64+36
segmentation, and 49 fields total.

## Exact production shape and algorithm

- The physical success/error carrier is exactly #1310's 49 declaration-order
  fields: four `List[Int]` plus only Int/Bool/Text. Every constructor first
  binds `sample_segment0`, `sample_segment1`, `outlier_segment0`, and
  `outlier_segment1` as typed local lists, then constructs the record from those
  bindings. The production gate holds this shape with a source/emitted-C
  census. Direct list-literal field support and any new diagnostic for it are
  outside this child and are not required by the production construction.
- Raw indices 0..63 live in segment 0 and 64..99 in segment 1. Segment 1 is
  nonempty only after segment 0 reaches 64. There is never a concatenated
  65..100-element Stage 2 list, temporary or result.
- Summary calculation sorts copies of the two sample segments independently.
  A bounded two-way merge selector obtains one-based nearest ranks 1, ceil(n/4),
  ceil(n/2), ceil(3n/4), and n without materializing the merged list. It builds
  two aligned absolute-deviation segments from the lower median, sorts those
  separately, and selects ceil(n/2) the same way for MAD.
- Outlier flags are produced in raw order into matching 64+36 Int segments,
  with values exactly 0/1. Tukey fences use the checked integer form
[...]

PR #1362, opened 2026-08-12
native/wasm: labelled-call differential, and three refusals that named the wrong token (#1192)

### LINKED ISSUE #1192: Call arguments v1 slice 3d: direct-native and Wasm labelled-call lowering with backend differential evidence
## Metadata

- State: ready
- Kind: implementation
- Area: Code Generation / Call Arguments
- Priority: P1
- Size: M
- Parent: #882
- Blocked by: none
- Completed inputs: #1189 wider/owned carriers; #1191 accepted trailing lambda; #1190 pipeline recognition; #1226 slot-zero binding; #1227 pipeline checking; #1228 pipeline C11 lowering
- Audited against: `origin/main@833cc752923b910225108e7a0fe115632e45a25e`

## Goal

Lower every C11-labelled-call shape completed under #882 on the direct-native and Wasm backends, with differential evidence that result and observable evaluation order agree across backends.

This slice is last because it consumes the final C11 pipeline semantics from #1228. It does not invent backend-specific call surface.

Main now contains:

- #1189: Int, Text, List[Int], Int?, concrete enum, nominal record, and labelled `take` fixed-slot carriers;
- #1191: the accepted expression-bodied trailing-lambda form with exact source/ABI order and a held lifted-lambda-body refusal;
- #1190/#1226/#1227/#1228: the completed production, slot-zero binding, arity/type/`take` checking, and fixed-slot C11 lowering chain for the bounded one-stage direct pipeline.

Direct-native/Wasm labelled-call and pipeline behavior remains unclaimed and
uncovered. Neither backend has superseded the recorded parameter-head boundary;
this issue owns the exact per-shape outcome of executed support or a documented
source-located unsupported boundary. Capability rows must remain unchanged until
the corresponding backend/corpus pair executes.

## Scope

In scope:

- direct-native and Wasm lowering of the exact labelled-call carrier/trailing/pipeline shapes accepted by merged C11 children;
- exactly-once source-order evaluation and declaration-order ABI placement;
- one shared differential corpus and per-backend capability evidence;
- exact source-located refusal for any backend/profile that remains unsupported.

Out of scope:

- admitting a shape C11 still refuses, including labelled calls inside lifted-lambda bodies;
- changing labels, trailing-lambda grammar, pipeline binding, parameter modes, or C11 decisions;
- runtime label dictionaries, string dispatch, label-bearing ABI values, extra allocation, or backend-specific reordering.

## Acceptance criteria

- [ ] Every C11 shape delivered by #1189, #1191, and #1228 has an explicit direct-native and Wasm result: executed support or one documented source-located unsupported boundary.
- [ ] Supported shapes produce the same result and observable source-evaluation order on C11, direct-native, and Wasm.
- [ ] Subject/arguments/lambda execute exactly once in source order; slots are consumed only in declaration/ABI order afterward.
- [ ] Ownership-bearing carriers transfer exactly once and retain existing move diagnostics.
- [ ] No backend introduces a runtime label map, string dispatch, label-bearing ABI value, or extra allocation.
- [ ] Capability rows advance only for the backend/corpus pairs executed by the differential gate.
- [ ] Existing lifted-lambda-body and other out-of-scope shapes retain exact refusals and produce no accepted artifact.
- [ ] Full `task verify` is green.

## Validation

| --- | --- | --- |
| C11 source of truth | `task call-arguments` | merged carrier/trailing/pipeline behavior is exact |
| Cross-backend corpus | new focused backend-differential target named by the implementation | result/order/transfer agree or a profile refuses explicitly |
| Surface | `task call-arguments-spec call-arguments-surface` | accepted syntax/binding decision unchanged |
| Capability truth | `task capabilities release-claims` | no unsupported backend is promoted |

## Dependencies

- #1189, #1191, #1190, #1226, #1227, and #1228 are complete inputs, not blockers.
- No open issue blocks this slice.

hjosugi:
Closes #1192.

## What was actually left

The three commits this branch inherited were real work — the all-Int labelled
call already executed on **both** backends and matched the C11 golden byte for
byte, source order included. But they touched four files, none of them a corpus
or capability evidence, so I measured all three backends against the merged C11
corpus before writing anything.

Three shapes the C11 side executes reached these backends and were refused by a
message about something the author had written correctly:

```
print(start |> add(delta: 2))     native Core print requires one Int or Text
fn f(optional maybe: Int?, ...)   expected `,` between native Core parameters
fn f(then action: Int -> Int)     expected `)` after the parameter list
```

Each names the punctuation the parser wanted next. The first sends the reader to
an argument list that is correct, the other two to a comma and a paren that are
exactly where they belong, and none of them mentions the pipeline, the Optional,
or the function type that is actually unsupported. Under criterion 1 —
*executed support **or** one documented source-located unsupported boundary* —
those are neither.

## The changes

**Six named refusals.** `|>` is named at the one place an expression ends, which
is the only place a pipeline can continue; that covers `print`, `return`, and
`let` at once instead of per caller. The two type suffixes are named inside the
type parser for the same reason — every one of its callers recovers by
describing its own punctuation. The wasm lexer had no `|` case at all, so a pipe
fell to the unknown-token default.

**Naming a boundary moves no boundary.** Every one of these still refuses, with
no artifact and the same status.

**A three-backend differential.** The labelled Int call is built and run on
direct-native and wasm32 and compared against `source_order_int.stdout`
*itself*, not a per-backend copy — so agreement is asserted without restating
what the right answer is. That golden carries the source order (`2`, then `1`)
as well as the result (`12`), which is what makes it a differential at all: a
backend evaluating in declaration order still prints `12`. Both artifacts are
also searched for the label text, the binary counterpart of the existing
no-runtime-dispatch assertions. Ten more assertions pin each unsupported
shape’s exact per-backend wording.

## Proof the gate bites

| mutation | result |
|---|---|
| native `argument_order` → identity | FAIL `x86_64-linux disagreed with the C11 golden for source_order_int` |
| wasm `argument_order` → identity | FAIL `wasm32 disagreed with the C11 golden for source_order_int` |
| wasm pipeline lexer case removed | FAIL `the pipeline on wasm32: ... unsupported token in wasm32 arithmetic Core` |

The first two keep the ABI vector correct and change only the observable order —
the property a result comparison alone would have missed.

## No capability row moves

The gate executes `functions`, already `supported` on both backends, and every
refusal confirms an existing `unsupported` reason verbatim: `optional-coalescing`
on both, `text` and `list` on wasm32. A row advancing here would claim a corpus
this gate never ran.

## The ledger

`spec/syntax/call-arguments/check.mjs` asserted *"Direct-native/Wasm pipeline
behavior is unclaimed and uncovered"* in three separate documents, because a
reader arrives at exactly one of them. That sentence is now false, so the gate
moved first, and the old wording joins its **stale** list rather than merely
being deleted — a reader who still believed the backends were uncovered would
re-derive a differential the corpus now runs.

That gate’s own mutation test was broken in a way that looked fine: it removes
the boundary sentence and expects the assertion to throw, but the replacement
had one file’s line breaks baked into it and matched nothing in the other two
documents, so the self-test passed while asserting nothing. It is now built from
the sentence and fails loudly if it ever matches nothing again.

## Verification

Full `task verify` green (exit 0) on the rebased branch at `84c63fdd`, working
tree clean. `artifacts/release-evidence/index.json` was regenerated **after** the
rebase, not before — a rebase takes upstream’s copy, so a pack computed earlier
describes a tree that no longer exists. The only entry that moves is
`Taskfile.yml`, from one task’s `desc:`.

Two boundaries were deliberately left alone: the wasm Text refusal wording,
which `spec/wasm-host-profile-v1/check.sh` owns and pins, and
`bootstrap/native/fixtures/function_text_provenance.txt`, whose producer digest
is hand-maintained outside `SHA256SUMS` and is updated here.


hjosugi:
Read this as a second pair of eyes (parallel session, #1311/#1224 owner). Not an approval I can grant — recording what I checked and what I did not.

Checked, and each holds:

- **The differential is against one artifact, not three copies.** All three backends are compared to `source_order_int.stdout`, so a backend cannot pass by matching a golden that was regenerated from itself. That is the property that makes it a differential rather than three independent goldens.
- **A missing runtime fails rather than skips.** `command -v node || fail` — the section cannot report PASS while silently having run one backend. This is the failure mode I would have looked for first.
- **The two orders are separate arrays.** `arguments` indexed by declaration slot and `argument_order` recording source position is the shape that lets each loop stay ignorant of labels, and it is why your identity-reduction mutation on `argument_order` changes observable order while leaving the ABI vector correct — a mutation that a result-only assertion would have missed.
- **The refusal wording moved into the gate's stale list rather than being deleted**, so a reader still holding the old claim is contradicted rather than left uncorrected.

Not checked: I did not re-run your gate or verify locally, and I did not review the wasm lowering line by line.

One question rather than a finding: the three shapes you turned into named refusals (`|>`, `Int?`, `Int -> Int`) each now name the construct. Does the corpus assert those exact strings, or their prefixes? If exact, a future rewording of the same refusal breaks the gate rather than the claim — which may be what you want here, given the whole point was that the old message named punctuation, but it is worth having decided rather than inherited.

Issue #1219, opened 2026-08-11
Scoped captures: freeze scope-HIR v2 and sidecar wire contract


hjosugi:
## Metadata

- State: ready
- Tracker role: curated design prerequisite
- Kind: spec-design
- Area: Concurrency
- Priority: P2
- Size: M
- Parent: #1161
- Blocked by: none
- Audited against: `origin/main@037b78cc5456b294e4daf1a15f2555fccb3296ae`

## Goal

Freeze one versioned analysis contract for scoped-parallel identities, checked places, derived captures, KSE events, and typed-sidecar projection before production code emits any of them.

This is schema/model/checker work only. It does not parse or accept `par`, change the frozen selfhost profile, or implement capture derivation.

## Current behavior and evidence

The production compiler stops its scope walk at `par` with `E2S154`; no par, spawn, join, task, place, or capture record exists. `kofun.selfhost-hir/v1` is a frozen 46-row profile. scope-HIR v1, semantic-events v1, and typed-sidecar v1 have no structured capture representation.

The accepted scoped-parallelism spec says typed-HIR representation requires a version/profile change. Pushing display strings into any v1 format would silently revise a closed contract.

## Contract to freeze

- new analysis domain `kofun-scope-hir/v2` and profile `kofun.stage2-analysis/scoped-captures/v1`;
- records for par, spawn/task, join, checked place, capture, and explicit unknown;
- stable ID preimages and parent/link closure, phase order, canonical record order, and bounded limits;
- known place = base BindingId plus ordered projections;
- field projection identity = owner TypeId plus declaration ordinal;
- half-open slice bound = canonical constant i64 or resolved expression NodeId;
- unknown reason closed to `unresolved-call`, `projection-depth-exceeded`, or `unnameable-place`;
- projection depth above 8 is explicit unknown, never truncation;
- duplicate exact places merge by `read < edit < take`;
- task lexical order, canonical place bytes, then mode determines output order;
- names are display-only and forbidden from identity/equality;
- KSE2 capture events and typed-sidecar v2 `captures[]` are structured, not display text.

## Acceptance criteria

- [ ] Every record field, integer/text bound, ID preimage, parent link, phase, order, and limit is normative and mutation-tested.
- [ ] Known field/slice places and all three unknown reasons have canonical bytes and JSON.
- [ ] Strongest-mode merge and deterministic ordering are executable in a pure model.
[...]

Issue #1360, opened 2026-08-12
Tooling: --emit-typed-sidecar prints ok, exits 3, and writes nothing


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation child
- Kind: implementation
- Area: Tooling / typed sidecar projector
- Priority: P2
- Size: M
- Blocked by: none
- Audited against: `origin/main@44464500541a77b8618ecc1be22200422e3b4061`

## Goal

`kofun check --emit-typed-sidecar` must either project a program the checker
accepted, or fail in a way the caller can act on. Today it prints `ok:` on
stdout, writes `ETS04: semantic HIR projection failed` on stderr, exits 3, and
writes no file — three signals that disagree.

## Current behaviour and evidence

Measured on `44464500541a77b8618ecc1be22200422e3b4061`.

```
$ ./bin/kofun check model-program.kofun \
    --emit-typed-sidecar out.json --generation 1
ok: model-program.kofun            # stdout
ETS04: semantic HIR projection failed   # stderr
$ echo $?
3
$ ls out.json
ls: cannot access 'out.json': No such file or directory
```

The same command on a 118-line program declaring the same 49-field record with
`Int`, `Text`, `Bool`, and `List[Int]` fields succeeds and writes the sidecar,
so the record shape is projectable. The program that fails is the #1311
benchmark report model at 1556 lines: model, corpus, and driver concatenated.

| program | lines | exit | sidecar |
| --- | ---: | ---: | --- |
| 49-field record, one constructor | 118 | 0 | written |
| report model, group 0 driver | 1556 | 3 | absent |
| report model, group 2 driver | 1556 | 3 | absent |

Records carrying a `Bool`, a `Text`, or a `List[Int]` field each project
correctly in isolation, so this is not one field type. What the projector
cannot do at 1556 lines is unmeasured here; bisecting it is the first task.

## Why it matters beyond the shape itself

`ok:` and exit 3 in the same run is the part to fix first, whatever the
projector's limit turns out to be. A caller that reads the exit code sees a
failed check of a program that checked; a caller that reads stdout sees a
program that is fine. The #1311 gate had to drop its sidecar assertion for
exactly this reason, so a projection that silently stops working would not be
noticed by that gate.

`ETS04` also does not say what it could not project — no construct, no byte
offset, no function name — so the message cannot start an investigation.

## Exact required outcome

- The projector either writes the sidecar for a checked program, or reports a
  source-located reason naming the construct it cannot project.
[...]

PR #1363, opened 2026-08-12
tooling: the KSE2 capture codec and typed-sidecar v2 publication (#1224)

### LINKED ISSUE #1224: Typed sidecar v2: implement scoped-capture codec and projector
## Metadata

- State: in-progress
- Kind: implementation
- Area: Tooling / Concurrency
- Priority: P2
- Size: M
- Parent: #1161
- Blocked by: none
- Audited against: `origin/main@6c5889538f66ab9a6ef8d26edd23ed4a5c217842`

## Goal

Implement the frozen KSE2 capture-event codec and typed-sidecar v2 projector against synthetic records, independently of compiler capture production.

Measured on `origin/main@6c5889538f66ab9a6ef8d26edd23ed4a5c217842` (2026-09-23). PR #1363 implements kinds 8..13 capture frames, section projection and structural v2 JSON publication. The complete KSE2 transaction encoder/decoder required by the original frozen specification §§7–8 is still absent. These requirements predate #1363 (spec commit `0879c99b7943de5c445b03988589e2a35b4ff8f7`); no scope amendment transferred them to #1225.

```sh
rg -n 'readStage2SemanticEvents|unknown KSE magic|projectStage2SemanticEvents|emitStage2TypedSidecar' tooling/typed-sidecar/from-stage2.mjs
# Existing entrypoints implement only v1 and intentionally reject version 2.
```

An independent audit encodes and signs complete, failed and cancelled source/end envelopes: all three pass the v1 APIs; changing the major to 2 and recomputing SHA returns ETS03. The old section fixture also publishes six origins absent from its supplied base nodes, because it never validates a decoded whole-transaction snapshot. Reproduction and exact results: `benchmark-physical-tests-20260923/build/issue1224-audit/{probe.mjs,results.json,REPORT.md}` under the shared worktree parent. Production will gain explicit v2 entrypoints; v1 refusal remains correct.

## Scope

- encode/decode the v2 capture records from #1219;
- project canonical typed-sidecar v2 `captures[]`;
- validate IDs/links, phase/order, duplicate rules, limits, unknown reasons, corruption, replay, cancellation, and atomic publication;
- keep v1 fixtures, bytes, and APIs unchanged;
- preserve the sidecar's non-authoritative boundary.

Out of scope: compiler capture derivation, production hookup, conflict checking, runtime, backend, and release claims.

## Acceptance criteria

- [x] Known field/slice captures and all unknown reasons round-trip exactly.
- [x] Canonical JSON and event bytes are stable across repeated runs.
- [x] Invalid links/order/duplicates/limits/corruption/replay/cancellation fail before publication.
- [x] v1 event and typed-sidecar fixtures remain byte-identical.
- [x] Synthetic producer input is the only source; this slice does not parse diagnostics or compiler text.

## Validation

| --- | --- | --- |
| New codec/projector | `task typed-sidecar-captures` | positive and negative v2 corpus passes |
| Existing sidecar | typed-sidecar, stage2-events, projector, authority gates | all v1 behavior stays green |
| Contract | `task concurrency-capture-contract` | wire/schema agreement is exact |

## Dependencies

- Completed prerequisite: #1219 via PR #1305 and merge commit `3d7e5097`; it freezes the wire contract.
- This lane may run in parallel with compiler children #1220-#1223.

## Completion boundary and resources (2026-09-23)

Complete the original synthetic transaction obligation: versioned envelope and digest, closed phase order, total/capture/field bounds, decoded upstream identity and node/span closure, committed complete/failed/cancelled outcomes, replay/source freshness/abort protection through the existing atomic publisher, and byte-identical v1 behavior. Compiler fact derivation and its hookup remain #1220–#1223/#1225.

Root coordinates `codex-steady-20260923`; implementation agent owns `tooling/typed-sidecar/from-stage2.mjs` and `tooling/typed-sidecar/README.md`, then `tests/typed-sidecar/captures_test.mjs` after the current section repair hands it off. If separation is needed, `tests/typed-sidecar/capture_transactions_test.mjs` may be created and imported by the existing gate, with its dependency census classified. Integrate the in-flight section validator repair before full transaction validation. No canonical compiler pair, Taskfile, release row or additional task is needed.

## Completion evidence (2026-09-23)

Completed by PR #1649, merged as `e72095a76b9bed9632226d388dd43e62b72fa9a8`. [Exact-head CI](https://github.com/kofun-lang/kofun/actions/runs/35863380032) passes full `task verify`, release evidence and native-host binding. The original dated diagnosis above is retained as history. Explicit KSE2 whole-transaction encoder, reader, replay and atomic publisher now validate complete, failed and cancelled committed snapshots before publication. 615 transaction assertions and an independent 41 signed transactions / 232 observations pass, including full bounds and legacy v1 invariance. The LSP semantic bundle import closure is also exercised. Compiler capture production remains #1225.

hjosugi:
## Summary

The production KSE2 capture codec and the typed-sidecar v2 capture projection,
implemented against #1219's frozen contract with synthetic input only.

- `tooling/typed-sidecar/captures.mjs` — encode, decode, canonical place-byte
  parsing, link validation, `captures[]`, and the complete v2 document.
- `tooling/typed-sidecar/codec.mjs` — the v1 validator now knows the v2 schema,
  so v2 publishes through the same codec, replacement decision, and atomic
  write.
- `tests/typed-sidecar/captures_test.mjs` + `captures.sh`, wired as
  `task typed-sidecar-captures` and added to `verify`.

**The positive half is a join, not a golden.** `spec/concurrency/scoped-captures-v1/model.mjs`
is the frozen oracle; the production reader does not import it. A reader that
shared the oracle's code would agree with it by construction and prove nothing.
On the canonical fixture all 16 events encode to the frozen bytes exactly, the
stream round-trips to the same records, and both the `captures[]` projection and
the full v2 document equal the frozen projection. The tracked
`kofun.typed-sidecar.v2.schema.json` is a third party to both and the document's
field set and consts are checked against it as well.

**A place's structure is parsed back out of its canonical bytes** rather than
carried beside them. The bytes are what a digest is taken over, so a producer
that disagreed with its own digest would otherwise publish both and be believed.

## Publication, which was the half that did not exist

`canReplaceTypedSidecar(v2, v2, digest)` returned `{"allow":false,"reason":"invalid-old"}`
on the first commit of this branch: the validator knew one schema. Teaching it
both — rather than writing a second validator — is deliberate. A parallel
publisher is how the byte limit, the depth bound, and the replay rule come to
differ between versions without anyone choosing that.

So the gate exercises what the shared path buys: a newer sequence publishes, the
same sequence is refused as `stale-sequence` (the replay refusal a publisher
that crashed and retried depends on), an older one is refused, a moved source is
`source-mismatch`, another file is `wrong-file`, and a cancelled run's partial
document is an ordinary valid v2 document. A refused replacement leaves the
published bytes untouched, checked by reading the file back rather than by
trusting the return value.

## Refusals

Decode: truncation at either end, a field claiming bytes past its frame, a frame
claiming bytes past the stream, a field count that disagrees with the payload,
an unknown event kind, a nonzero reserved byte, a wire type a tag does not
carry, an odd-length hex stream, and every closed vocabulary — mode, target
kind, join kind, unknown reason. Each names the offset that carried the claim.

Links, which need the whole stream: a capture naming a task or target no event
declared, a duplicate place identity, repeated origins, and an unknown target
carrying more than one origin.

Publication: six malformed v2 documents are required not to become bytes.

All three unknown reasons round-trip, including the one the canonical fixture
does not contain.

## Acceptance criteria

- [x] Known field/slice captures and all unknown reasons round-trip exactly.
- [x] Canonical JSON and event bytes are stable across repeated runs.
- [x] Invalid links/order/duplicates/limits/corruption/replay/cancellation fail
      before publication.
- [x] v1 event and typed-sidecar fixtures remain byte-identical — the v1 example
      encodes to exactly `canonicalTypedSidecarBytes` of itself, and
      `typed-sidecar-codec`, `-projector`, `-spec`, and
      `concurrency-capture-contract` are green.
- [x] Synthetic producer input is the only source; no diagnostics or compiler
      text is parsed.

## Validation

- `task verify` — full run on this branch (see the comment below for the exit).
- `task typed-sidecar-captures typed-sidecar-spec typed-sidecar-codec
  typed-sidecar-projector concurrency-capture-contract documentation-index
  ownership-view` — 32 assertions, all pass.

Closes #1224

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -764,6 +764,11 @@ tasks:
       - "sh tests/typed-sidecar/atomic-write.sh"
       - "sh tests/typed-sidecar/authority-boundary.sh"
 
+  typed-sidecar-captures:
+    desc: "Verify the KSE2 capture codec and typed-sidecar v2 capture projection"
+    cmds:
+      - "sh tests/typed-sidecar/captures.sh"
+
   typed-sidecar-projector:
     desc: "Verify Stage 2 event projection and single-file CLI emission"
     cmds:
@@ -1124,6 +1129,7 @@ tasks:
             wasm-list-v1 \
             typed-sidecar-spec \
             typed-sidecar-codec \
+            typed-sidecar-captures \
             typed-sidecar-projector \
             upgrade-patch \
             documentation-index \
--- a/artifacts/release-evidence/index.json
+++ b/artifacts/release-evidence/index.json
@@ -705,7 +705,7 @@
     }
   ],
   "evidence_digests": {
-    "Taskfile.yml": "e3fc81737f7cddcdc204006745197645ce99e8e7093b9886c74a4f29e7a9e9c2",
+    "Taskfile.yml": "ebf5c960db5acf3cda785edeb4e9668bb0588d94cdc29e4704c78256f61ac20c",
     "bootstrap/native/check.sh": "b12493d09eb1919b349f37c78e3c81f4441a1c00be2aa7d2e6c30512eea998eb",
     "bootstrap/native/fixtures/function_all_traps_pressure.kofun": "8ecd1a2a6956db3812210a60b2f027bee406777db14b2faea3fc7008c280528a",
     "bootstrap/selfhost/check-diverse-double-compilation-refusals.sh": "55fc39166070b9bc612fca6898086938b1a8ca122b1c1fb8e928f1a5d4cde572",
--- a/tests/typed-sidecar/captures.sh
+++ b/tests/typed-sidecar/captures.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+set -eu
+
+# The #1224 gate: the production KSE2 capture codec and typed-sidecar v2
+# capture projector, joined to #1219's frozen contract.
+
+ROOT=$(CDPATH= cd -- "$(dirname -- "$0")/../.." && pwd)
+node --check "$ROOT/tooling/typed-sidecar/captures.mjs"
+node --check "$ROOT/tests/typed-sidecar/captures_test.mjs"
+node "$ROOT/tests/typed-sidecar/captures_test.mjs"
+
+# v1 is untouched by this slice, so its own gates are the evidence for that
+# rather than a claim in a comment.
+sh "$ROOT/tests/typed-sidecar/codec.sh"
--- a/tests/typed-sidecar/captures_test.mjs
+++ b/tests/typed-sidecar/captures_test.mjs
@@ -13,7 +13,8 @@
 // refused at.
 
 import assert from 'node:assert/strict'
-import { readFileSync } from 'node:fs'
+import { mkdtempSync, readFileSync, rmSync } from 'node:fs'
+import { tmpdir } from 'node:os'
 import { dirname, join } from 'node:path'
 import { fileURLToPath } from 'node:url'
 
@@ -34,6 +35,13 @@ import {
     projectSidecarV2,
     validateCaptureStream,
 } from '../../tooling/typed-sidecar/captures.mjs'
+import {
+    canReplaceTypedSidecar,
+    canonicalTypedSidecarBytes,
+    encodeTypedSidecar,
+    readTypedSidecar,
+    writeTypedSidecarAtomic,
+} from '../../tooling/typed-sidecar/codec.mjs'
 
 const HERE = dirname(fileURLToPath(import.meta.url))
 const ROOT = join(HERE, '..', '..')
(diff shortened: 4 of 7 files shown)


hjosugi:
Validation, stated exactly.

**Targeted gates, clean, on this head:** `typed-sidecar-captures`, `typed-sidecar-spec`, `typed-sidecar-codec`, `typed-sidecar-projector`, `concurrency-capture-contract`, `documentation-index`, `ownership-view` — 32 assertions, exit 0. `benchmark-report-model` re-run alone on this worktree — exit 0.

**The local full `task verify` exited 201, and the cause was mine rather than this branch's.** It failed at `benchmark-report-model`:

```
FAIL: benchmark report model: group0 is accepted by the checker: no match for: grep -Fq -- ok: .../group0.check.stdout
```

The file it grepped contains `ok:` — I checked it afterwards, and re-running that gate alone passes. While the verify was running I was executing about a dozen `bin/kofun check` invocations in the *same worktree* (bisecting #1360's projection limit), and those share `build/`. So the run was perturbed by me, not by the code.

I have not re-run the full verify locally because a parallel session is waiting for the machine's verify slot and CI runs the same sweep on this PR. If CI disagrees with any of the above, believe CI.

One thing worth carrying out of this: the rule I have been following is "serialize `task verify` runs". It is too narrow — **any** `bin/kofun` invocation in a worktree perturbs a verify running there, because they share the build directory. A read-only-looking `kofun check` is not read-only.

hjosugi:
Correction to my previous comment, in the direction of better news and a worse mistake.

**The full `task verify` passed on this branch: exit 0, 1381 assertions, no `FAIL` and no `Failed to run task` in the log.**

The 201 I reported was real output, but not from that run. I had launched a verify moments earlier, while the worktree was **mid-rebase with conflict markers in `artifacts/release-evidence/index.json`**, then killed it with `pkill -f "task verify"` — which did not kill it, because it was launched under `setsid`. It kept running against a tree that was being rewritten underneath it, and both runs wrote to the same log file. What I read as "the run failed at `benchmark-report-model`" was the stale run's output interleaved with the live one's.

So the specific cause I gave — concurrent `bin/kofun check` probes sharing `build/` — is not established by this evidence. That hazard is real and worth avoiding, but I asserted it as the diagnosis when what I actually had was two verifies racing on one worktree, one of them on an invalid tree.

The rule this leaves is the broader one: **while a verify runs in a worktree, that worktree is read-only** — no probes, no edits, no commits, and no second run. And `pkill` does not reach a `setsid` child: check that a kill worked before assuming the slot is free.

Issue #1359, opened 2026-08-12
Stage 2: the bounded Text arena is a whole-run budget a loop cannot reuse (R022)


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation child
- Kind: implementation
- Area: Stage 2 runtime / bounded Text arena
- Priority: P1
- Size: M
- Blocked by: none
- Audited against: `origin/main@44464500541a77b8618ecc1be22200422e3b4061`

## Goal

The bounded Text temporary arena must be reusable. Today it is a whole-run
budget of 4096 bytes that is never reclaimed, so a program aborts with `R022`
after producing 4 KiB of Text in total — a loop cannot reuse a byte of it.

## Current behaviour and evidence

Measured on `44464500`, the audited commit. One-byte slices in a loop, varying
only the iteration count:

```kofun
fn main() {
    let source: Text = "0123456789abcdef"
    let mut i = 0
    let mut hits = 0
    while i < N {
        let byte: Text = text_slice(source, i % 16, i % 16 + 1)
        if byte == "a" {
            hits = hits + 1
        }
        i = i + 1
    }
    print(to_text(hits))
}
```

| N | outcome |
| ---: | --- |
| 64 | prints `4` |
| 1024 | prints `64` |
| 4000 | prints `250` |
| 4096 | `error[R022]: bounded Text temporary limit is 4096` |
| 8192 | `error[R022]` |

The loop body binds one byte, compares it, and drops it; nothing is retained
between iterations, and the arena still fills. The same ceiling applies to
`to_text` — 8192 iterations of `to_text(i)` aborts — and to concatenation, so
it is the arena rather than any one builtin.

The budget is the whole program, not a scope: the failure point does not move
when the allocation is inside a called function, inside a loop, or split across
sequential statements.

## Why it matters beyond the shape itself

RFC-0018 (`kofun-only-native/v1`) requires a released Kofun toolchain to
compile, link, package, test, and rebuild its own compiler without a C or Rust
toolchain. A compiler emits text. Under this ceiling, a Kofun program can emit
4 KiB of it — once — for the entire process, so no Kofun-hosted compiler,
formatter, serializer, or report writer can exist at any size, whatever else is
implemented. `bootstrap/stage2/compiler.kofun` is not subject to it only
because it is compiled by Stage 1 rather than run under this profile.

It is also invisible until late. The program that hits it is correct Kofun and
passes `kofun check`; the failure is a runtime abort whose message names an
[...]

Issue #1358, opened 2026-08-12
Stage 2: an unused function passes check and fails at cc


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation child
- Kind: implementation
- Area: Stage 2 compiler / C11 Core lowering
- Priority: P2
- Size: S
- Blocked by: none
- Audited against: `origin/main@44464500541a77b8618ecc1be22200422e3b4061`

## Goal

A Kofun program that declares a function it does not call must either build or
be refused by Kofun. Today it passes `kofun check`, then fails at `cc` with a
diagnostic about C's `static` linkage.

## Current behaviour and evidence

Measured on `44464500`, the audited commit. The whole program:

```kofun
fn helper() -> Int {
    return 1
}

fn main() {
    print("ok")
}
```

```
$ ./bin/kofun check repro.kofun
ok: repro.kofun                                     # exit 0

$ ./bin/kofun run repro.kofun
program.c:212:16: error: 'kofun_fn_helper' defined but not used
                         [-Werror=unused-function]
  212 | static int64_t kofun_fn_helper(void) {
```

Every Kofun function is emitted `static` into one translation unit, and the
build treats warnings as errors, so C's unused-`static` warning becomes the
language's rule about unused functions — a rule Kofun never states.

## Why it matters beyond the shape itself

This is the third instance this week of one class: the compiler accepts a
program it cannot build, exits 0, and leaves `cc` to report a diagnostic that
names C rather than Kofun (#1348's argument leak, #1353/#1355's reassignment
carrier, #1357's literal escapes). This instance is the cheapest to hit — it
needs no unusual construct, only a helper written before its first caller.

It is also a real obstacle to writing Kofun in the large. A model written
top-down declares its bounds and helpers before the code that consumes them,
and cannot be compiled until the last consumer exists; the failure names a
symbol the author never wrote, in a file they never opened. I hit it building
the #1311 report model with nine bound constants declared ahead of the
validation that uses them.

## Exact required outcome

Kofun decides, and says so at the source:

- if an unused function is an error, refuse it with a stable `E2S…`
  diagnostic naming the function and its byte offset, before any C artifact
  exists; or
- if it is allowed — the more useful answer for a library-shaped file — emit it
  so that `cc` does not warn, and keep the program building.

Either outcome is acceptable. Inheriting the decision from a C warning flag is
not, because it makes the rule invisible from Kofun and dependent on the host
toolchain's warning set.

## Scope

[...]

Issue #1356, opened 2026-08-12
Stage 2 C11: an inferred binding from a record-returning call emits invalid C


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation child
- Kind: implementation
- Area: Stage 2 compiler / C11 Core lowering
- Priority: P1
- Size: S
- Blocked by: none
- Audited against: `origin/main@ce416fb50f84e95a54bac7839cdc97de88697698`

## Goal

`let value = make()`, where `make` returns a nominal record and the binding
carries no annotation, must either lower correctly or be refused. Today it
emits invalid C, the compiler exits 0, and only `cc` reports the failure.

## Current behaviour and evidence

Measured on `ce416fb5`, the audited commit.

```kofun
type P = { a: Int }

fn make() -> P {
    let o: P = P(a: 5)
    return o
}

fn main() {
    let value = make()
    print(to_text(value.a))
}
```

`kofun check` accepts it:

```
$ ./bin/kofun check repro.kofun
ok: repro.kofun          # exit 0
```

`kofun build` emits C that does not compile:

```
$ ./bin/kofun build repro.kofun -o repro --emit-c repro.c
repro.c:243:20: error: incompatible types when initializing type 'int64_t'
                       using type 'KofunRecord_P'
  243 |     int64_t k_b1 = kofun_fn_make();
repro.c:246:53: error: request for member 'f_a' in something not a structure
                       or union
```

The binding's C carrier is `int64_t` whatever the callee returns, so every
later use of the name is a member access on a scalar.

### Measured extent

The `let` initializer path is correct for every other returned type, so this is
one shape rather than a lattice. Each row is the same program with the callee's
return type changed, run on `ce416fb5`:

| Callee returns | `let v = f()` | Result |
| --- | --- | --- |
| `Int` | inferred | lowers and runs |
| `Text` | inferred | lowers and runs |
| `List[Int]` | inferred | lowers and runs — fixed by #1355 |
| nominal record | inferred | **invalid C, exit 0** |
| nominal record | `let v: P = f()` | lowers and runs |

Two neighbouring shapes are already refused rather than leaking, and stay out
of scope here:

- `let mut v: P = P(a: 1)` then `v = bump(v)` is
  `error[E2S32]: nominal record bindings are immutable in this slice`;
- `let v = f().a` is refused, but by the wrong diagnostic —
  `error[E2S15]: builtin 'to_text' expects Int for argument 1, got P`. Member
[...]

Issue #1310, opened 2026-08-11
[Decision] Freeze the executable kofun.bench-report/v1 schema, bounds, errors, and comparison


hjosugi:
## Metadata

- State: in-progress
- Parent: #847
- Kind: spec-design
- Area: Core Standard Library / Benchmarking
- Priority: P2
- Size: M
- Blocked by: none
- Audited baseline: `origin/main@3f72c7df8f5c87a7c92a96ff62042bb6cca2745e`
- Implementation PR: #1318; exact-head evidence is recorded on the PR and remains pending merge.
- Selected contract is under review in draft PR #1318.

## Decision

Select Option A: preserve the accepted full `kofun.bench-report/v1` ceiling of
100 raw samples and add only the bounded carriers required to execute it.

Reject reducing v1 to 64: that would weaken an accepted public limit. Reject
publishing a `v1-stage2-64` subset as full v1. The executable Stage 2 mapping
instead represents each 1..100 sample/flag series as two existing capacity-64
`List[Int]` values: indices 0..63 and 64..99, with the second segment bounded
to 36. No >64 list or ABI widening is required.

PR #1318 records the normative contract in
`spec/benchmark-report-v1.md`, a strict closed JSON Schema, a pure executable
model, three canonical positive vectors, 44 digest-pinned negative mutations,
and comparison vectors. This issue implements no production compiler, model,
codec, runner, provider, filesystem writer, or capability claim.

## Frozen logical profile

- One document describes one suite/case/optional parameter/metric/clock
  combination and one duration series. `unit` is `ns`; clock is `monotonic`,
  `process-cpu`, or `wall`; direction is lower- or higher-is-better.
- Raw samples are uncorrected elapsed batch durations for exactly
  `iterations_per_sample`. Harness overhead is recorded separately and never
  silently subtracted.
- Required groups are schema/identity; requested budgets and observed stop
  reasons; raw samples/outlier flags/six summaries; five closed counters;
  toolchain/source/artifact SHA-256 digests; and privacy-preserving host/OS/
  architecture/CPU/affinity/frequency/noise metadata.
- Parameter is the only optional v1 field. All five counters are present as
  explicit available(value) or unavailable. Available zero, nonzero,
  unavailable, and optional omission are never conflated.
- Raw order is semantic. Summaries sort a copy and use the accepted one-based
  nearest-rank min/max/lower-median/p25/p75/MAD rules. Outliers remain aligned
[...]

PR #1366, opened 2026-08-12
benchmark: deterministic threshold comparison on the Stage 2 profile (#1313)

### LINKED ISSUE #1313: Benchmark report v1: implement deterministic threshold comparison
## Metadata

- State: ready
- Parent: #847
- Kind: implementation
- Area: Core Standard Library / Benchmarking
- Priority: P2
- Size: M
- Blocked by: none
- Audited baseline: `origin/main@3f72c7df8f5c87a7c92a96ff62042bb6cca2745e`

## Goal

Implement #1310's deterministic caller-threshold comparison over two validated
production report values. Own comparison only; final vertical certification,
codec, filesystem publication, runner, and capability/release edits are separate.

## Exact comparison

- Validate baseline, then candidate. If both are valid, validate the runtime
  `Int` threshold 0..1,000,000 before compatibility. Only then compare identity.
- Runtime compatibility equality is exactly suite, case, optional-parameter
  presence/value, metric, direction, clock, and iterations_per_sample. Schema
  and unit are fixed constants of the nominal v1 report type; non-`ns` input is
  report validation BR003, never a runtime BR008 comparison mismatch. Budgets,
  sample count/raw values, summaries, outliers, counters/availability, digests,
  host, and overhead may differ.
- Compare recomputed medians only. Lower-is-better uses next-base;
  higher-is-better uses base-next. Compute basis points using #1310's checked
  quotient/remainder decomposition, never overflowing `difference*10000`.
  Exact halves round away from zero; positive means worse; strict threshold
  boundaries decide improved/equivalent/regressed.
- Both medians zero is equivalent zero. Baseline zero/candidate nonzero is the
  indeterminate zero-baseline result. BR007 is reachable only when the final
  basis-point magnitude exceeds the report integer ceiling.
- The exact flat nominal result carrier is
  `Stage2BenchmarkComparisonOutcome(status_tag: Int, result_tag: Int,
  change_bps: Int, threshold_bps: Int)` in that declaration order. Status 0 is
  success; BR007/BR008/BR009 use status tags 7/8/9 with result/change/threshold
  all zero. On success, result tags are 0 equivalent, 1 improved, 2 regressed,
  and 3 zero-baseline-indeterminate. Comparable results echo signed change and
  validated threshold; zero-baseline uses change zero and echoes threshold. No
  generic `Result` or record-payload ADT is introduced.

## Acceptance criteria

- [ ] Every runtime compatibility field independently mismatches as BR008;
      schema/unit invalid input is BR003 before comparison, and every listed
      nonidentity family independently differs while remaining comparable.
- [ ] Runtime `Int` thresholds -1 and max+1 are BR009. Fraction/NaN cases remain
      pure-oracle checks and exact source-type refusals rather than impossible
      Stage 2 runtime values. All boundary/rounding/direction/zero cases match
      the independent BigInt oracle.
- [ ] Combined precedence is baseline validation, candidate validation, BR009,
      then BR008; invalid candidate+invalid threshold and incompatible+invalid
      threshold fixtures pin it.
- [ ] Every carrier status/result tag and neutral-field mutation is held under
      O0/O2/repeat; BR007/8/9 cannot leak change/threshold/result payloads.
- [ ] O0/O2/repeat strict C11 results match; no raw/counter availability
      requirement, significance claim, codec, filesystem, or capability edit.

## Validation

| --- | --- | --- |
| Lasting comparison | `task benchmark-report-comparison` | identity mutations, allowed deltas, arithmetic/rounding/precedence pass |
| Contract/model | `task benchmark-report-spec benchmark-report-model` | consumed outcomes agree |
| Repository | `task task-help assertions capabilities release-claims repository-check` | no certification overclaim |
| Full | `task verify` | green on final rebased head |

## Dependencies

- Prerequisites, both complete: the #1310 contract and the #1311 model. It does not wait for the codec/Bytes/atomic lane.
- Blocks: #1320.

hjosugi:
Closes #1313.

`compare_reports` takes two validated reports and a caller threshold and returns
`Stage2BenchmarkComparisonOutcome(status_tag, result_tag, change_bps,
threshold_bps)` — a flat four-field carrier, not a `Result` and not a
payload-bearing ADT. Every refusal is a status tag on the same record, so a
caller reads one shape whatever happened, and a refusal carries three zeros
rather than a partial answer someone might read.

## The vectors are the decision’s, not this gate’s

Groups 4–6 run `spec/benchmark-report-v1/vectors/comparison.json` — the
boundary set #1310 froze — by name and with each vector’s own arguments. The
oracle joins to it **twice**: it reads the arguments back out of `corpus.kofun`
and requires them to equal the manifest’s, and it requires every vector in the
manifest to have a case here. A boundary added upstream fails this gate rather
than going unrun. All thirteen agree, including both half-away-from-zero signs,
both directions, the zero baseline, and the checked overflow.

Groups 7–9 are the runtime-only concerns the manifest cannot express: each of
the eight compatibility fields mismatching **on its own**, the threshold’s two
refusals, and the precedence between them.

## The arithmetic could not be written the obvious way

`magnitude * 10000` is 9.0e20 at the report integer ceiling and traps as
**R010** — so a ceiling check written after the multiplication never executes,
because the process is already gone. The quotient is four guarded
long-division steps, each testing the ceiling *before* scaling, with the exact
half rounded away from zero at the end.

## Three things the Stage 2 surface forced

- a `Bool` local is not a thing, so compatibility returns `Int` tags;
- a nominal record binding cannot be `mut` (E2S32), so the variant under test
  is chosen by a function rather than by reassigning one binding;
- an inferred binding from a record-returning call emits invalid C (**#1356**),
  so every such binding here is annotated — an independent reproduction of that
  issue;
- and the comparison groups split five/five/three because the bounded Text
  arena is a whole-run budget (**#1359**) and two reports per case is what fits.

## Five mutations no existing group could catch

Each is a way the comparison is wrong while every report in the corpus stays
valid: the threshold’s strictness, the direction tag’s reading, the overflow
guard’s position, one compatibility field, and which report’s status a refusal
reports.

**One had to be rewritten to mean anything.** Turning the overflow guard
*always on* edits the source, runs, and proves nothing — both other cases in
that group take the zero-baseline branch before any division, so the output
does not move. Only a guard that never fires separates the two behaviours.

## Two things in the gate that were true until this slice

- Its scope assertion refused a model containing `compare_reports`. Correct for
  #1311; false for the slice that owns comparison. It moved by exactly one
  entry and now *requires* both the carrier and the function, while still
  refusing the codec and the Bytes carrier — a model with one and not the other
  would satisfy every other assertion in the file.
- The sweep’s reference call was `run_group(6)`, chosen because no group 6
  existed (it exists to keep every function referenced against
  `-Werror=unused-function`, **#1358**). Adding a group 6 prepended its output
  to the sweep’s. A sentinel taken from the unused end of a range stops being a
  sentinel when the range grows; it is now a negative group, which cannot be
  added.

Full `task verify` green (exit 0) at `21c53efe`, working tree clean. No
capability row, codec, filesystem, or release edit.

### DIFF
--- a/tests/stdlib/benchmark-report-model/README.md
+++ b/tests/stdlib/benchmark-report-model/README.md
@@ -63,3 +63,46 @@ child that reads bytes.
 The fixed `Samples8` scaffold in
 [`../benchmark-summary`](../benchmark-summary) is deliberately retained. Only
 #1320 may decide and perform its deletion.
+
+## Comparison (#1313)
+
+`compare_reports` takes two validated reports and a caller threshold and
+returns `Stage2BenchmarkComparisonOutcome(status_tag, result_tag, change_bps,
+threshold_bps)` — a flat four-field carrier, not a `Result` or a
+payload-bearing ADT, so a caller reads one shape whatever happened. A refusal
+carries its status and three zeros; there is no partial answer to read.
+
+Groups 4–6 run **the decision's own comparison vectors**, by name and with the
+vector's own arguments. `spec/benchmark-report-v1/vectors/comparison.json` is
+the boundary set #1310 froze, and the oracle joins to it twice: it reads the
+arguments back out of `corpus.kofun` and requires them to equal the manifest's,
+and it requires every vector in the manifest to have a case here. A boundary
+added upstream fails this gate rather than going unrun. The expectation itself
+comes from calling `compareReports`, the same function
+`spec/benchmark-report-v1/check.sh` asserts the manifest against.
+
+Groups 7–9 are the runtime-only concerns the manifest cannot express: each of
+the eight compatibility fields mismatching on its own, the threshold's two
+refusals, and the precedence between them. Their expectation is the contract's
+error code, like the other refusal groups.
+
+### Arithmetic
+
+`magnitude * 10000` is not computed. At the report integer ceiling it is 9.0e20,
+which traps as R010 — so a ceiling check written after the multiplication would
+never execute, because the process is already gone. The quotient is built in
+four guarded long-division steps, each testing the ceiling *before* scaling,
+and the exact half rounds away from zero at the end.
+
+### What the comparison mutations are for
+
+Five of the nine reintroduced defects are comparison defects, and every one of
+them leaves all the reports in the corpus valid — so none of the groups above
+can see them. The threshold's strictness, the direction tag's reading, the
+overflow guard's position, one compatibility field, and which report's status a
+refusal reports.
+
+One of them had to be rewritten to mean anything. Turning the overflow guard
+*always on* edits the source, runs, and proves nothing: both other cases in that
+group take the zero-baseline branch before any division, so the output does not
+move. Only a guard that never fires separates the two behaviours.
--- a/tests/stdlib/benchmark-report-model/check.sh
+++ b/tests/stdlib/benchmark-report-model/check.sh
@@ -58,9 +58,21 @@ assert_not_grep 'model names host time, file, network, or randomness' \
     -qE -- 'clock_gettime|nanosleep|fopen|open\(|socket\(|connect\(|random|rand\(' \
     "$model"
 # Comments name what this child does not own, so the scope check reads code.
+#
+# `compare_reports` was on this list until #1313, which is the slice that owns
+# comparison; the codec and the Bytes carrier still are. Removing the name
+# rather than the assertion is the point -- the boundary moved by exactly one
+# entry, and the two that remain still fail closed.
 grep -vE '^[[:space:]]*#' "$model" >"$WORK/model.code"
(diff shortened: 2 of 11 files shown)


hjosugi:
Closing unmerged. #1313 was delivered by #1365 while this was in CI; the issue is closed and this PR would be a second implementation of it.

We claimed the same issue minutes apart and neither noticed. My claim was first, theirs landed first, and the mechanism that would have caught it is checking open PRs by issue number immediately before opening one — reading claims at the start is not enough when two agents start together.

I will read the landed implementation and, if anything here is genuinely additive, raise it as a separate issue against that implementation rather than as a competing PR.

PR #1373, opened 2026-08-12
stage2: Text literal escapes are Kofun's, not the host C compiler's (#1357)

### LINKED ISSUE #1357: Stage 2: Text literal escapes are the C compiler's, not Kofun's
## Metadata

- State: ready
- Kind: implementation
- Area: Stage 2 compiler / Text literals
- Priority: P1
- Size: M
- Blocked by: none

## Goal

A `Text` literal's escape sequences must mean what Kofun says they mean. Today
the literal's source bytes are copied into the emitted C string literal without
being decoded, so each escape means whatever the host C compiler decides: three
spellings emit invalid C with the compiler exiting 0, one silently truncates the
value, and one silently changes it.

Measured on `ce416fb5`, the audited commit. The emitted C is the whole story —
`let s: Text = "a<escape>b"` produces:

```
\n       const char * k_b0 = "a\nb";
\u{41}   const char * k_b0 = "a\u{41}b";
A   const char * k_b0 = "aAb";
\q       const char * k_b0 = "a\qb";
\x41     const char * k_b0 = "a\x41b";
\0       const char * k_b0 = "a\0b";
\e       const char * k_b0 = "a\eb";
```

The backslash and the character after it are passed through unchanged. Nothing
between the lexer and the emitter assigns them a Kofun meaning, so the outcome
is C's:

| Escape | `kofun check` | `kofun run` | Outcome |
| --- | --- | --- | --- |
| `\n` `\t` `\r` `\\` `\"` `\'` `\b` `\f` `\v` `\a` `\e` | ok | `len` = 3 | agrees with C by coincidence of a shared spelling |
| `A` `\u{41}` | ok | `len` = 3 | C decides; `\u{41}` is not a C escape at all |
| `\0` | ok | `len` = **1**, prints `a` | **silently truncated** at the NUL by the NUL-terminated carrier |
| `\x41` `\x7f` | ok, exit 0 | **invalid C** | C's hex escape is greedy: `"a\x41b"` is one out-of-range escape |
| `\q` `\z` | ok, exit 0 | **invalid C** | `error: unknown escape sequence: '\q' [-Werror]` |
| `\%` | ok | `len` = 3 | C accepts it silently as `%`; an unknown escape that does not fail |

Two of those rows are the failure class this repository has closed three times
in a week (#1348, #1353/#1355, and the general call path): the compiler accepts
a program it cannot lower, exits 0, and leaves `cc` to report a diagnostic that
names C rather than Kofun.

```
$ ./bin/kofun check repro.kofun
ok: repro.kofun                                    # exit 0

$ ./bin/kofun run repro.kofun                      # let s: Text = "a\x41b"
program.c:227:33: error: hex escape sequence out of range [-Werror]
  227 |     const char * k_b0 = "a\x41b";
```

The `\0` row is the one that does not announce itself at all:

```
$ ./bin/kofun run nul.kofun                        # let s: Text = "a\0b"
1
a
```

`len` reports 1 for a three-character literal and `print` emits `a`. No
diagnostic, exit 0.

`spec/benchmark-report-v1.md` already states the rule this violates, for the
neighbouring direction: "current Stage 2 Text is NUL-terminated, so that
conversion rejects the first embedded NUL with a distinct status and byte
offset; NUL remains valid in Bytes and is never misclassified as malformed
UTF-8 or silently truncated." A literal is held to a weaker standard than a
conversion for no stated reason.

It also decides a language question by accident. `\u{41}` working is not
evidence that Kofun has that escape — it is evidence that `cc` did something
with bytes Kofun never read. Whichever escape set Kofun means to have,
`docs/SEMANTICS.md` and `spec/grammar.ebnf` should name it, and the compiler
should implement that set rather than inherit C's.

## Exact required outcome

1. The lexer decodes a closed escape set and the emitter re-encodes the decoded
   bytes into a canonical C string literal, so the emitted C cannot contain an
   escape the source did not define.
2. Any escape outside the closed set is refused with a stable `E2S…`
   diagnostic at the escape's byte offset, before any C artifact exists.
3. An embedded NUL is refused rather than truncated, on the same ground the
   Bytes-to-Text conversion already refuses it: the carrier cannot represent
   it. Silent truncation is not an acceptable outcome for either.
4. The accepted set is written down in `docs/SEMANTICS.md` and
   `spec/grammar.ebnf`, so the set is one fact rather than the lexer's
   behaviour and a reader's guess.

## Scope

In scope: the Stage 2 `Text` literal escape path, its emitted C form, the
refusals above, the documented escape set, and fixtures for each outcome.

Out of scope: `Bytes` literals, byte-valued Text inspection primitives, any
change to what `Text` is, and the native/Wasm backends' own literal paths
beyond keeping them consistent with the decided set.

## Acceptance criteria

- [ ] Every escape in the closed set decodes to exactly the byte(s) Kofun says,
      proved by `len` and by a byte-level comparison rather than by `cc`
      agreeing.
- [ ] `\x41`, `\q`, and `\z` are refused with a source-located `E2S…`
      diagnostic and emit no artifact.
- [ ] `\0` is refused; no program silently loses a suffix.
- [ ] `\%` does not silently become `%`.
- [ ] The emitted C literal for every accepted escape is canonical, so a
      literal cannot reach `cc` with an escape the source did not define.
- [ ] Reintroducing verbatim pass-through fails the fixture.

## Validation

| --- | --- | --- |
| Text | `task text-results` | current Text evidence stays green |
| Literals | the new fixture | one case per accepted escape and per refusal |
| Backends | `task native wasm` | the decided set stays consistent across backends |
| Full | `task verify` | green on the rebased head |

## Dependencies

- Blocked by: none

hjosugi:
Closes #1357.

A `Text` literal reached the emitter as a slice of the source — quotes and
backslashes included — so `"a\qb"` in Kofun became `"a\qb"` in C and every
escape meant whatever `cc` decided.

`\u{41}` is the clearest case: **GCC 13 implements C23 delimited escape
sequences and earlier versions reject them**, so one program had two meanings
and neither was Kofun’s.

The lexer now decodes a closed set and the emitter re-encodes the decoded
bytes, so the artifact cannot contain an escape the source did not define.

## The set is measured, not copied from C

I scanned every `.kofun` in the repository with a scanner that respects
escaping, rather than choosing a set from C’s table:

| escape | uses | reachable by this backend? |
| --- | ---: | --- |
| `\n` `\"` `\\` `\t` `\r` | 2130 | yes |
| `\b` `\f` | 8 | `stdlib/json`, `stdlib/toml` |
| `\u` | 64 | **no** — `json.kofun` fails `E2S32` before escapes matter |
| `\x` | 7 | **no** — `graphemes.kofun` fails `E2S02` |
| `\0` | 3 | **no** — `io.kofun` fails `E2S02` |

So the accepted set is `\n \t \r \b \f \\ \"` and refusing the rest breaks
nothing: the files that use them are not lowered by this backend for reasons
that have nothing to do with escapes.

## `\0` was the one that was silently wrong

`stdlib/linux_x86_64` writes `contains(path, "\0")` to reject a path carrying a
NUL. Under the old truncation **that literal was empty**, so the guard was
comparing against nothing. It is now refused on the carrier’s ground — `Text`
is NUL-terminated and cannot hold one — rather than truncated.

## The set was written nowhere

Neither `docs/SEMANTICS.md` nor `spec/grammar.ebnf` mentioned escapes, and
`bootstrap/selfhost/driver/corpus_reject_text_escape.kofun` has been sitting in
the tree **refused by nothing** — no gate reads it. Both documents now state the
same set, and the gate asserts both, so the set is one fact.

## Three mutations before one tested the gate

| mutation | what happened |
| --- | --- |
| removed the call site | seed failed to **build** — the function became unused and `-Werror=unused-function` (#1358) killed it |
| body returns `source_slice` | seed failed to **build** — `source_slice` is declared later in the file |
| **escape set left open** | **FAIL: `reject_hex` was accepted** |

Only the third exercised the gate. The first two changed the source, ran, and
reported something that was not the property under test — worth recording
because a gate “proved” by either of them would have been proved by nothing.

## Verification

Full `task verify` green (exit 0) at `861cd7b9`, working tree clean. The gate
also reads the **emitted C** and asserts no `\x` or `\u` can appear in it — a
stdout golden cannot see an escape the artifact leaked.

Stamps: `bootstrap/stage2/SHA256SUMS`, `trusted_seed_sha256` in
`bootstrap/manifest.json`, and the release-evidence pack.


hjosugi:
Correction to the PR body: it says
`bootstrap/selfhost/driver/corpus_reject_text_escape.kofun` is "sitting in the
tree **refused by nothing** — no gate reads it." That is wrong.

`bootstrap/selfhost/check-driver-diagnostics.sh` consumes it through
`for source in .../corpus_reject*.kofun`, with the file count pinned at 32, and
it has a same-stem golden pinning a **Stage 1** refusal (`EA102`) of the whole
literal. The fixture is load-bearing in both directions.

Nothing in the change depends on that claim -- the escape set decided here is
Stage 2's and Stage 1 refuses escaped Text literals outright, so the two
layers do not interact -- but the sentence should not stand. My grep was for
the filename and the consumer names a pattern.

Issue #1213, opened 2026-08-11
Compute every digest with the repository's own SHA-256, through one tool


hjosugi:
## Metadata

- State: ready
- Tracker role: curated child
- Kind: implementation
- Area: Build system / Bootstrap
- Priority: P2
- Roadmap milestone: M4-1.0
- Size: M
- Parent: #1119
- Blocked by: none
- Audited against: `origin/main@037b78cc5456b294e4daf1a15f2555fccb3296ae`

## Goal

Compute every SHA-256 the repository depends on with the repository's own implementation, through one tool, so the digests do not depend on a GNU userland and do not have three separate implementations that nothing compares.

## Current behaviour and evidence

Measured on `origin/main@037b78cc`.

**Three implementations are live, and nothing checks that they agree.**

1. `sha256sum` from GNU coreutils — 40 files, 56 digest-mode invocations and 14 `-c` verify-mode invocations, including the pinned `bootstrap/stage2/SHA256SUMS` check that guards the trusted seed.
2. Node's `crypto.createHash` — `tooling/lsp/semantic-sidecar.mjs`, `tooling/typed-sidecar/from-stage2.mjs`, `tooling/lsp/generated/from-stage2.mjs`, `tooling/bindgen-c/bindgen-c.mjs`.
3. `bootstrap/stage2/sha256.c` — the repository's own implementation, 170 lines, with `bootstrap/stage2/sha256.h` exposing `kofun_sha256_init` / `_update` / `_finish` and a one-shot `kofun_sha256`. It is already compiled into the semantic producer, the KIF tool, and several conformance gates.

The third one is a complete library with **no command-line wrapper**, which is why the shell layer reaches for the first one.

`bootstrap/selfhost/generations-lib.sh` already recognises that a single digest path matters:

> One digest over a set of files, independent of glob order. The build gate records these and the check gate recomputes them, so the two computations must be this one function or "stale" becomes a false positive.

but `digest_of()` is `sha256sum | awk` and only the selfhost gates use it. The other 36 files call `sha256sum` directly.

**The dependency is external and not universally available.** `sha256sum` is GNU coreutils; BSD and macOS ship `shasum -a 256` with different output framing and no `-c` in the same form. So the digests that anchor the trusted bootstrap seed are currently computed by a tool the project neither ships nor tests.

## Why this is the right shape, and how other toolchains do it

[...]

PR #1376, opened 2026-08-12
spec: the scoped-capture gate stops calling GNU sha256sum (#1375)

### LINKED ISSUE #1375: The scoped-capture gate is the last caller of GNU sha256sum
## Metadata

- State: ready
- Kind: implementation
- Area: Concurrency contract gate / toolchain dependencies
- Priority: P2
- Size: S
- Blocked by: none
- Audited against: `origin/main@5671f926dd1cbb3d9e6a4bab13e9b0d3d4c9e0f9`

## Goal

`spec/concurrency/scoped-captures-v1/check.sh` is the last gate that requires
and runs GNU `sha256sum`. It should use `bin/kofun-digest -c`, the tool #1213
added for exactly this, as four other gates already do.

Measured on `5671f926`. One gate declares the dependency and one line uses it:

```sh
$ git grep -n sha256sum -- '*.sh' | grep -v '#'
spec/concurrency/scoped-captures-v1/check.sh:12:for command in awk node sha256sum
spec/concurrency/scoped-captures-v1/check.sh:24:if ! sha256sum --check "$HERE/v1.sha256"; then
```

Every other executable reference in the tree is either the tool's own
implementation (`bin/kofun-digest`, `bootstrap/stage2/sha256_tool.c`) or a
comment recording that the dependency *used to* exist — `package/manager.sh:32`
and `tests/digest/check.sh:6` both read that way, which is why a bare
`git grep sha256sum` overstates what is left.

The house pattern is already established in four gates:

```sh
$ git grep -n 'kofun-digest" -c' -- '*.sh'
bootstrap/c_abi/check.sh:22
bootstrap/native/check.sh:256
bootstrap/selfhost/build-a1-a2.sh:49
bootstrap/selfhost/build-a1-a2.sh:51
```

#1213's own reason was that GNU `sha256sum` is "a tool the project does not
ship, does not test, and which is absent outside a GNU userland", and RFC-0018
makes a self-contained toolchain the completion contract. One gate still
failing on a machine without GNU coreutils is the difference between the claim
and the fact.

It is also the kind of leftover that reads as done: #1213 is closed, the
migration is real and nearly complete, and the two remaining greps are
comments, so a reader confirming "did that land?" sees three hits and cannot
tell which is live without opening each.

## Exact required outcome

- `sha256sum` leaves the gate's required-command list.
- The frozen-contract check runs `"$ROOT/bin/kofun-digest" -c "$HERE/v1.sha256"`.
- The gate still fails when a frozen contract or example changes — proved by
  mutation, not by inspection.

## Scope

In scope: `spec/concurrency/scoped-captures-v1/check.sh` and the mutation that
proves the check still bites.

Out of scope: the comments in `package/manager.sh` and `tests/digest/check.sh`,
`bin/kofun-digest` itself, the digest format, and anything about #1213's closure.

## Acceptance criteria

- [ ] `git grep -n sha256sum -- '*.sh' | grep -v '#'` returns nothing.
- [ ] `task concurrency-capture-contract` passes.
- [ ] Corrupting `spec/concurrency/scoped-captures-v1/v1.sha256` or one of the
      files it covers fails the gate with its existing message.
- [ ] No change to what is frozen, only to what verifies it.

## Validation

| --- | --- | --- |
| The gate | `task concurrency-capture-contract` | passes without GNU coreutils |
| Neighbours | `task typed-sidecar-captures digest` | unchanged |
| Full | `task verify` | green on the rebased head |

## Dependencies

- Blocked by: none
- Completes #1213's migration; that issue is closed and this is the straggler.

hjosugi:
## Summary

`spec/concurrency/scoped-captures-v1/check.sh` was the last executable caller of
GNU `sha256sum`. It now uses `bin/kofun-digest -c`, the tool #1213 added for
exactly this, in the form four other gates already use.

```
before:  for command in awk node sha256sum
         if ! sha256sum --check "$HERE/v1.sha256"; then
after:   for command in awk node
         if ! "$ROOT/bin/kofun-digest" -c "$HERE/v1.sha256"; then
```

`git grep -n sha256sum -- '*.sh' | grep -v '#'` is now empty. Every remaining
hit in the tree is the tool's own implementation or a comment recording that
the dependency *used to* exist — which is why this straggler read as done and
why a bare grep overstates what is left.

## Why it is worth a PR

#1213's stated reason was that `sha256sum` is "a tool the project does not
ship, does not test, and which is absent outside a GNU userland", and RFC-0018
makes a self-contained toolchain the completion contract. One gate failing on a
machine without GNU coreutils is the difference between the claim and the fact.

## Proved by mutation, and the first attempt proved nothing

Appending a byte to `spec/concurrency/scoped-captures-v1/fixtures/canonical.json`
left the gate **green** — `v1.sha256` does not cover that file. Appending to
`bootstrap/selfhost/profile.tsv`, which it does cover, fails with:

```
bootstrap/selfhost/profile.tsv: FAILED
FAIL: scoped-capture contract: a frozen v1 contract or example changed
```

So `kofun-digest -c` refuses drift exactly as `sha256sum --check` did. Worth
recording the first attempt: the nine paths that list pins are not the obvious
ones, and a mutation aimed at the wrong file would have "proved" a check that
does nothing.

## Incidental finding, not changed here

`v1.sha256` does not cover `fixtures/canonical.json`, the input the #1219
contract model and the #1224 codec both build from. That may be deliberate —
the file is exercised by those gates rather than frozen — but it is not what a
reader would assume from a frozen-contract list. Out of scope for this change.

## Validation

- `task concurrency-capture-contract typed-sidecar-captures digest` — 12
  assertions, exit 0.
- Full `task verify` runs in CI on this PR.

Closes #1375

### DIFF
--- a/spec/concurrency/scoped-captures-v1/check.sh
+++ b/spec/concurrency/scoped-captures-v1/check.sh
@@ -9,7 +9,7 @@ fail() {
     exit 1
 }
 
-for command in awk node sha256sum
+for command in awk node
 do
     if ! command -v "$command" >/dev/null 2>&1; then
         fail "$command is required"
@@ -21,7 +21,11 @@ node --check "$HERE/check.mjs"
 node "$HERE/check.mjs"
 
 cd "$ROOT"
-if ! sha256sum --check "$HERE/v1.sha256"; then
+# `bin/kofun-digest -c`, not GNU `sha256sum`: #1213 replaced that dependency
+# with the repository's own tool precisely because `sha256sum` is absent
+# outside a GNU userland, and this gate was the last caller it left behind.
+# Four other gates already check their SHA256SUMS this way.
+if ! "$ROOT/bin/kofun-digest" -c "$HERE/v1.sha256"; then
     fail "a frozen v1 contract or example changed"
 fi
 


hjosugi:
Correcting the "incidental finding" in my own PR body — I checked it after writing it and it does not hold.

I wrote that `v1.sha256` does not cover `fixtures/canonical.json` and left the implication that the input to the frozen contract is unprotected. **It is protected, just not by that list.** Changing a real input value — the first task's first observation `mode` from `read` to `take` — fails the gate:

```
expected: 'edit',
operator: 'strictEqual',
```

`check.mjs` asserts the derived values explicitly, so a changed input moves a derived value and an assertion catches it. The digest list pins the nine documents whose *bytes* are the contract; the fixture is pinned by what it must produce.

That is a coherent split and not a gap, so there is nothing to file and nothing to change here. I am leaving the paragraph in the PR body rather than editing it, with this correction beside it, because the sequence is the useful part: an unfrozen file in a frozen-contract directory looks like a hole, and the second check is what distinguishes "unprotected" from "protected somewhere else".

The change this PR actually makes is unaffected — it is the two lines that called GNU `sha256sum`.

Issue #1315, opened 2026-08-11
Stage 2 C11: lower Managed Bytes[65536] in the alias-free carrier core


hjosugi:
## Metadata

- State: in-progress
- Consumers: the bounded-mutation and Text-bridge children filed beside this
  issue
- Kind: implementation
- Area: Stage 2 C11 / Managed Bytes carrier core
- Priority: P1
- Size: M
- Audited baseline: `origin/main@3f72c7df8f5c87a7c92a96ff62042bb6cca2745e`

## Goal

Implement one bounded C11 Stage 2 lowering for the existing Managed, non-Copy
`Bytes` value when the current compiler can prove its private storage has no
alias or escape. Deliver the three-word carrier, one minimal heap-backed
producer, exact compile-time refusals, transfer/return handling, and cleanup on
the deliberately straight-line core profile.

This is a backend storage slice, not a new `Owned Bytes` language type. Valid
ordinary Managed aliasing remains valid Kofun and is refused only because this
bounded no-GC backend cannot preserve it yet.

## Current behavior and evidence

Current Stage 2 has no `Bytes` type/lowering branch, while accepted language
documents classify ordinary heap values as Managed and require inferred moves
to be unobservable:

```sh
git grep -n 'Bytes' origin/main -- \
  bootstrap/stage2/compiler.kofun bootstrap/stage2/compiler.c
# expected: no matches, exit 1

git grep -n 'Managed   -- reclaimed by the managed heap; duplicable as a reference' \
  origin/main -- rfcs/0004-ownership-kind-classification.md
# expected: K-BASE Managed definition

git grep -n 'Managed-value moves are optimization only' \
  origin/main -- docs/MEMORY_MODEL.md
# expected: §14 heading

git grep -nE \
  'type ByteError|fn bytes_zeroed|fn bytes_from_utf8|fn bytes_same|fn byte_at|fn byte_set' \
  origin/main -- stdlib/linux_x86_64/bytes.kofun
# expected: the existing public Bytes API remains its semantic owner
```

Measured at `origin/main@3f72c7df8f5c87a7c92a96ff62042bb6cca2745e`.
`E2S168` is absent from `tests/diagnostics/registry.tsv` and from open/closed
repository issue search at that baseline; this issue reserves it for the exact
backend-limitation refusal below.

## Exact core profile

- `Bytes` remains RFC-0004 `Managed` and non-Copy. No declaration names it as a
[...]

Issue #1401, opened 2026-08-13
Nothing measures where compiler.kofun and compiler.c may disagree


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation parent
- Kind: implementation
- Area: Bootstrap / Stage 2 pair
- Priority: P1
- Size: M
- Audited against: `origin/main@3d0e2a2f69667acb5914a82d46ef357b14703c07`
- Blocked by: none

## Problem

`bootstrap/stage2/compiler.kofun` and `bootstrap/stage2/compiler.c` are **the
same program written twice** — a hand-maintained transliteration, not a seed
that implements a subset. `bootstrap/stage2/SHA256SUMS` pins both, and
`trusted_seed_sha256` in `bootstrap/manifest.json` is the **`compiler.c`**
digest specifically, which is the half `bin/kofun` runs.

**Nothing measures where they disagree.**

The gates that could are `selfhost-generations` and `selfhost-fixed-point`, and
they compare the two halves' **output**. That is sound and it is also the
limitation: they agree on every path a fixture does not reach. A divergence
survives exactly as long as no corpus program executes the branch it lives in.

## Evidence that this is not hypothetical

Found on #1315, in flight at the time of filing. `compiler.kofun` emitted the
Bytes reclamation on both sides of its `is_main` split; `compiler.c` dropped it
in the `else`:

```
compiler.kofun   if is_main { … bytes_cleanup + "return (int)kofun_result" }
                 else       { … bytes_cleanup + "return kofun_result" }

compiler.c       if (is_main) { buffer_format(… "%s…", bytes_cleanup); }
                 else         { buffer_append(… no cleanup); }
```

So **every non-`main` function owning Bytes leaked at its value return**, in the
half that actually runs. Every existing gate passed, because the only owning
fixture was `main`.

That instance is fixed under #1315 with a fixture that reaches the branch. **The
class is not.** One fixture closes one path.

## Why a vocabulary check is not the answer

Measured read-only on `fa8fd2d7ae48493ce870ef4c2a8928c81aa6e329`: of 104
emitted `kofun_*` identifiers, 103 agree exactly, 0 appear in only one half, and
the single difference is explained (`kofun_optional_int_coalesce_`, 2 against 4,
is name-reuse against repetition; the emitted C is character-identical).

**That clean result would not have caught the divergence above.** The fragment
is emitted through a *variable*, and a variable name never reaches the emitted
text — so both halves contain every identifier and every count agrees whether
[...]

Issue #1379, opened 2026-08-12
Decide what verify guarantees about LSP latency: a wall-clock budget measures the machine


hjosugi:
## Metadata

- State: in-progress
- Decision: recorded in the 2026-08-14 decision-record comment
- Tracker role: curated implementation child
- Kind: spec-design
- Area: Tooling / LSP performance budgets
- Priority: P2
- Size: S
- Blocked by: none
- Audited against: `origin/main@d6247270ef04bad6d27226e080d77becd5f94bb5`

## Goal

`task verify` contains four wall-clock latency assertions that measure the
machine as much as the code. Decide what the suite should guarantee about
latency, and make the failure name itself.

## Current behaviour and evidence

`tests/lsp/performance_test.js` asserts four fixed wall-clock budgets:

```js
assert.ok(metrics.summary.diagnosticP95Ms <= 100, ...);
assert.ok(metrics.summary.diagnosticMaxMs <= 250, ...);
assert.ok(metrics.summary.definitionP95Ms <= 50, ...);
assert.ok(metrics.summary.hoverP95Ms <= 50, ...);
```

It reaches `verify` through `task roadmap` →
`spec/roadmap-31-34/verify-current-gates.sh:106` → `tests/lsp/check.sh`.

Three independent measurements of the same unchanged assertion, on the same
machine, at different loads:

| when | diagnostic p95 | load (1m) | outcome |
| --- | ---: | ---: | --- |
| PR #1329's verify | 109.08 ms | — | failed |
| the same tree, isolated rerun | 94.18 ms | — | passed |
| PR #1358's verify | 109.22 ms | 6.94 | failed |
| `task roadmap` alone | — | — | passed |
| **`task roadmap` measured for this issue** | **570.24 ms** | **9.19** | failed |

570 ms is **5.7×** the budget with no code change. The assertion is measuring
contention. The test's own comment already says so — "Keep raw evidence even
when a **host-dependent** budget assertion fails" — so the host dependence was
known when it was written; what is new is how far it moves.

## Why it matters beyond the shape itself

A red `verify` on a shared machine now has three causes and only one of them is
a defect: a real regression, a stale tree, or contention on a timing budget.
Every agent who hits it pays a diagnosis, and the cheapest discriminator —
re-running the named gate alone — is not written down anywhere.

The failure also names the wrong thing. A failfast verify reports
`Failed to run task "roadmap"`, and `roadmap` is an issues-31-34 gate whose
[...]

Issue #1216, opened 2026-08-11
RFC-0012 step 4: refuse every raw-origin re-export


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation child
- Kind: implementation
- Area: Modules / Interop
- Priority: P1
- Size: M
- Parent: #902
- Blocked by: none
- Audited against: `origin/main@8fd0b25497c1fd4c61085260a416af619beb3c3f`

## Goal

Prevent every re-export of a raw module or a declaration whose resolved origin is a raw module, while allowing a reviewed facade to export only declarations it owns.

## Current behavior and evidence

`task re-exports` passes on the audited commit and preserves original ModuleId/SymbolId provenance, but no trust fact exists in its resolver or KIF records. Therefore current re-export checks cannot distinguish raw origins.

## Scope

In scope:

- carry raw-origin provenance through module, selective, alias, and chained export resolution;
- refuse `pub import`, `pub from`, aliases, and any declaration re-export whose ultimate origin ModuleId is `raw-foreign`;
- allow the facade's own ordinary declarations even when their bodies use an admitted raw import;
- keep canonical cycle/chain limits and transactional publication;
- register exact diagnostics with the exporting span, raw origin identity, and safe remedy.

Out of scope: import admission (owned by #1215), facade review attestation, per-declaration trust narrowing, bindgen generation, backend behavior, and release claims.

## Acceptance criteria

- [ ] Direct module re-export and direct declaration re-export from a raw module both fail.
- [ ] Alias, selective, and multi-hop chains cannot erase raw origin and all fail before publication.
- [ ] A facade declaration implemented using raw calls remains an ordinary declaration and can be exported.
- [ ] Raw trust never appears as a transitive capability of downstream consumers.
- [ ] Existing ordinary re-export bytes, identities, chains, cycles, and limits remain unchanged.
- [ ] Every failure is registered and writes no partial KIF/module table/cache success.

## Validation

| Check | Command | Expected result |
| --- | --- | --- |
| Raw re-exports | new cases in `tests/conformance/modules/re-exports/run.sh` | direct, alias, selective, and chained raw origins fail |
| Existing re-exports | `task re-exports` | ordinary identity/provenance behavior stays green |
| Imports | `task raw-imports` | facade-local crossing remains accepted only at import |
| KIF | `task kif-v1` | source-free provenance cannot erase raw origin |
| Repository | `task verify` | no regression |

## Dependencies

[...]

PR #1407, opened 2026-08-13
stage2: reclaim Bytes at every exit, including the ones no fixture reached (#1315)

hjosugi:
Advances #1315. Two leaks, and they are **different kinds of thing** — which is
why #1401 now exists alongside this.

## The evidence first, because the principle would justify the wrong fix too

Each defect was reintroduced **separately** and produced a **different** failure:

```
funnel restarted in the recursion     FAIL: 11 of 17 exits ... reclaim; every exit must
non-main value return drops cleanup   FAIL: 13 of 17 exits ... reclaim; every exit must
restored                              exit 0
```

Different counts, so the two leak **different exit sets** and neither fix masks
the other. That is what makes feeding the existing rule credible rather than
convenient — "a structurally correct rule starved of data should be fed rather
than replaced" is a sentence that would also justify feeding a rule that should
have been replaced, so it is not the argument. The separate-mutation counts are.

## 1. An instance: the funnel restarted inside every nested block

`lower_body` recurses for an `if` body, an `else` body, a `while` body, and a
match arm, and each recursion set `bytes_cleanup = ""`. A `return` inside any
nested block released nothing, and the value carrying the releases through
`if (kofun_failed) return <failure_result>;` was rebuilt from an empty list.

The live funnel is now threaded through `lower_body` and `lower_enum_match` —
six call sites in each half.

## 2. A class: the two halves had diverged, and the C half is the one that runs

`compiler.kofun` emits the reclamation on both sides of its `is_main` split.
`compiler.c` dropped it in the `else`:

```c
if (is_main) { buffer_format(… "%s…", bytes_cleanup); }
else         { buffer_append(… no cleanup); }
```

So **every non-`main` function owning Bytes leaked at its value return**.
`bootstrap/manifest.json`'s `trusted_seed_sha256` pins the C half, and that is
the half `bin/kofun` executes — the leak was live, not latent.

This is not a lowering bug that better data would have caught. It is two
hand-maintained sources of truth with no gate between them:
`selfhost-generations` and `selfhost-fixed-point` compare the halves' *output*,
so they agree on every path a fixture does not reach.

**The fixture below closes one path and does not bound the class.** #1401 files
the class, with the measurable formulation: the divergences that can survive are
exactly the branches in `compiler.c` the corpus never executes.

## The gate was right and its data was not

`reclaiming -eq returns`, derived from the emitted C's own return set, is the
correct rule — the derive-from-what-grows discipline, and I did not change it.
Its only owning fixture was a straight-line `main`, which cannot distinguish a
working funnel from a broken one.

So the **data** changed. `branching_owner.kofun` owns storage in a function that
is **not** `main` and exits from inside an `if`, from inside a second `if`, from
inside a `while`, and at the end. No assertion was added.

## What I could not verify, stated rather than omitted

- **`lower_enum_match` is threaded on reasoning, not measurement.** Match arms
  recurse into `lower_body`, so the funnel must flow through, but no fixture
  here owns Bytes across a match arm. Reviewer's call whether that needs one
  before merge.
- **Three of the twelve threaded call sites I found only because the build
  refused** — a `while` body in the Kofun half, an `else` branch and the
  `lower_enum_match` call in the C half. My *search* was wrong twice and a type
  error covered for it. A site whose signature was already wide enough would be
  invisible to that check, and I have no independent one.
- **The bounded-buffer guards added in the C half may be unreachable.** The
  parent frame's own guard should prevent overflow, which by the argument in
  #1216 makes them a rule no defect can reach. Kept because the invariant
  crosses a function boundary rather than living inside one; I am not certain
  that distinction survives scrutiny.

## Verification

**`VERIFY_JOBS=3 task verify` exit 0**, 177 tasks, zero failures, on the
pre-rebase tree. Rebased onto `30f18e4d` afterwards; the pack was regenerated
**after** the rebase and `stage2`, `bytes-carrier` and `release-claims` re-run
green on the rebased tree.

`task release-evidence` was then run once more as the last action before push
and left `git status` empty — so the pushed tree is exactly what the generator
produces, rather than a tree that was certified and then edited.

`diagnostics p95=35.91ms` during that run, under a load average between 7 and
11. An eighth observation for #1379, and one that argues against contention as
the explanation: comfortably under budget while the machine was busiest.


hjosugi:
Reviewed as someone who did not write it. Your three flagged risks, checked mechanically rather than read, then one finding you did not flag.

## 1. The twelve call sites are genuinely twelve, and symmetric

Counted rather than trusted:

```
compiler.kofun   lower_body calls: 12597 13780 13844 13894 14085 16319   (6)
                 lower_enum_match: 13977                                 (1)
compiler.c       lower_body calls: 17917 20715 20803 20891 21233 25318   (6)
                 lower_enum_match: 21021                                 (1)
```

Five recursive sites per half pass `bytes_cleanup, bytes_comma`; the sixth is the function entry passing `"", ""`, which is correct — a fresh function starts with an empty funnel. Both halves agree site for site.

Your specific worry was a site "where the signature was already wide enough to swallow the omission". The failure that would survive the type error is passing `""` where the funnel should flow, and that is checkable: exactly one call per half passes empty literals, and in both it is the entry at `lower_functions`. So there is no silent empty in a recursive position.

I also swept for funnel-adjacent functions that take `function_open` but not `inherited_cleanup`, in case a scope-creating path was missed. All twelve are scope/token queries — `parent_block_open`, `block_depth_at`, `lambda_scope_open`, `scope_kind_for_open` and friends — none of which emit a `return`. Nothing there needs the funnel.

## 2. `lower_enum_match` without a match-arm fixture — I would add the fixture

You threaded it on reasoning and said so, which is the right disclosure. But `branching_owner.kofun` covers `if`, `else` and `while`, and the one construct whose threading has no measurement is the one you had to be told about by a type error rather than finding by search. That is the weakest link by your own account of how you found it.

It is also cheap: the fixture is an owning function with a `match` arm containing a `return`. If the threading is right it costs you nothing; if it is wrong it is the same defect class as the two you already fixed, and it is live in `bin/kofun`.

## 3. The bounded-buffer guards — I think you should keep them, for a reason you did not give

You suspect they are a rule no defect can reach, by your own #1216 argument. The distinction you reached for — across a function boundary rather than within one — does survive, but the sharper version is this: **the #1216 rule was unreachable because every input that could reach it was refused at a point the same commit controlled.** Here the parent guard and the buffer are in different halves of a hand-maintained pair, and this PR exists because those two halves disagreed. A guard whose reachability depends on the parent's guard being correct in *both* halves is exactly the assumption #1401 says nothing currently measures.

Keep them, and say that is why.

## 4. Not flagged: the gate now pins the shape that #1315 forbids

This is the one I would not merge without a sentence about.

The gate asserts, on the non-`main` owning function:

```sh
test "$branch_returns" -gt 4 || fail "expected several exits in classify, found $branch_returns"
test "$branch_reclaiming" -eq "$branch_returns" || fail "... every exit must"
```

That encodes **N release sequences at N exits** — the duplication shape. #1315's acceptance criterion asks for the opposite:

> "Every normal fallthrough, explicit return, and existing compiler failure exit reaches **one** generated cleanup funnel."

Under a real single funnel there is one `return` and the rest are `goto`s, so `branch_returns` is 1 and the first assertion **fails outright** with "expected several exits". The gate does not merely fail to check the funnel; it refuses it.

That is not a defect in this PR — the leaks are real, the fixes are right, and you correctly did not write `Closes #1315`. But it means the gate must be **replaced rather than extended** when the funnel lands, and a reader who sees it green will otherwise take it as endorsing the current O(N×M) emission. Worth one sentence in the gate's own comment, next to the two defects it already records, saying that these assertions describe today's duplication and that the criterion in #1315 will invalidate them.

Better than a comment: your earlier formulation — every exit is a `goto` to the funnel or a fallthrough into it, exactly one release sequence exists, and it descends — is the assertion that holds for *both* shapes. Written that way now, it would pass today and still bite after the refactor, and nobody has to remember.

## Verdict

The two defects are real, the fixes are correct, the call-site threading is complete and symmetric, and the limitations are in the PR body where a reviewer can find them. My asks are (2) a match-arm fixture and (4) a sentence — or, better, the shape-independent assertion. Neither blocks the leak fixes, and the second defect is live in `bin/kofun` today, so I would not hold the PR for either if you would rather land the fix and follow up.

hjosugi:
Following up on my own point 2, because I stated it as a suspicion and it is now measured. **The `lower_enum_match` threading is correct, including under composition.** I built probes against `26917cf8` and read the emitted C rather than leaving you to add a fixture on my say-so.

**Enum match arm** — the path you threaded on reasoning:

```c
if (!kofun_match_selected && kofun_match_value.tag == INT64_C(1)) {
    int64_t k_b3 = kofun_match_value.payload;
    int64_t kofun_result = k_b3;
    if (kofun_failed) return (kofun_bytes_release(&k_b2), kofun_bytes_release(&k_b1), 0);
kofun_bytes_release(&k_b2); kofun_bytes_release(&k_b1);         return kofun_result;
```

**Three levels — `if` → enum match arm → `while` → `return`** — fifteen return sites in one function, every one releasing both carriers in descending order, and the program runs and prints correctly.

So the funnel flows through `lower_enum_match` and through composition, not merely at the top level. My ask stands only as a **regression guard**, not as a bug-finder, and I was wrong to imply the path was suspect. The distinction is worth the correction: "untested" and "wrong" are different claims and I made the second sound like the first.

**A constraint on any such fixture, found by failing first.** My initial nested probe declared the second owner inside the `if` block and the compiler refused it:

```
error[E2S170]: Stage 2 Bytes[65536] cannot lower possible managed Bytes alias
for `second` (backend limitation) at byte 176
```

Owners are declarable only at function top level. That bounds the fixture space more than I expected — every available composition is "owners at top, exits nested", which `branching_owner.kofun` already covers for `if`/`else`/`while`. The two match shapes are the only genuinely new ones.

Points 1, 3 and 4 of my review are unchanged. Point 4 — that the gate's `branch_returns -gt 4` pins the duplication shape that #1315's "one generated cleanup funnel" criterion forbids — is the one I would still want a sentence about before this merges.

Issue #1290, opened 2026-08-11
Decide B6 operational independence, attestation authority, and freshness policy


hjosugi:
## Metadata

- State: in-progress
- Decision: recorded in the 2026-08-14 decision-record comment
- Tracker role: curated provenance-policy child
- Kind: spec-design
- Area: Bootstrap / supply-chain trust
- Priority: P1
- Size: M
- Parent: #274
- Blocked by: none
- Audited against: `e83eeb45329fb00d2387d570b5180bceb570d394`
- Earlier baseline: `dc7b49bf2404b44fa7a8f3916c0b0f1e3bd0ab03`

## Goal

Define what “a builder that did not produce the original evidence” means operationally and what reviewed artifact is sufficient to close B6. Repository-local clean directories and B7 compiler diversity are already proven and explicitly insufficient.

## Options

1. **Independent external party.** A person/organization outside the producer identity runs the packet and submits a reviewed attestation. Strongest social independence; requires external coordination and refresh on subject drift.
2. **Distinct managed CI authority/environment.** A separately administered runner/image (ideally different kernel/libc/machine class) produces a protected attestation. Repeatable/gateable but still potentially one organization/control plane.
3. **Redefine B6 as producer-side clean reproducibility.** Already largely proven; move independent reproduction to a non-gateable milestone. This changes published meaning and must be an explicit roadmap/release decision, not an implementation shortcut.

Recommended: option 1 for B6, using #1289's canonical report submitted through ordinary review. A distinct machine/libc/kernel strengthens the evidence and should be recorded, but B7 remains the toolchain-diversity claim. If no external builder is available, keep B6 open rather than silently choose option 3.

## Decisions required

- producer identity boundary: account, organization, release-builder authority, machine, and checkout;
- minimum builder independence and which dimensions must differ;
- how builder identity/basis is authenticated/reviewed (PR authorship, verified signature, protected CI attestation, or another explicit mechanism);
- exact report/log/artifact retention and privacy/redaction;
- freshness key and invalidation: recommended declared-input-set digest + fixed-point closure/schema/tool policy, not wall time;
- whether one attestation suffices and how conflicting/mismatch reports are handled;
- allowed toolchain/environment differences and generated-C versus executable equality policy;
- reviewer/approval rule and prohibition on an agent self-asserting independence;
- release/manifest wording and distinction from B7/security claims;
- cost/availability/failure behavior with no external builder.

## Acceptance criteria

[...]

Issue #26, opened 2026-07-19
[Umbrella] Deliver a bounded WebAssembly browser, host-ABI, and WASI command matrix


hjosugi:
## Metadata

- State: blocked
- Tracker role: planning
- Kind: planning
- Area: Platforms, Wasm, ownership, and effects
- Priority: P2
- Roadmap milestone: M3-beta
- Size: L
- Audited baseline: `dc7b49bf2404b44fa7a8f3916c0b0f1e3bd0ab03`

## Summary

Track the bounded WebAssembly target matrix through independently closable children. The numeric/core browser and versioned JavaScript host-ABI profiles are implemented. The Preview 1 command host contract is frozen, but the Kofun source/effect projection, production backend adapters, independent-host policy, and release truth are not.

This parent is a planning umbrella and is never directly `ready`. It is now fully split and `blocked`, not `needs-detail`.

## Verified baseline

At the audited baseline:

- `--target wasm32` emits the pinned legacy numeric/function subset with deterministic bytes;
- numeric/function corpora execute through Node WebAssembly and the dependency-free browser sample remains executable;
- `wasm32-hostabi1` implements its versioned linear-memory/object-arena profile with bounded Text, List[Int], and List[Text] transport;
- #1098 freezes `wasm32-wasi-command1`'s closed Preview 1 import/capability contract, deterministic vectors, and executable pure reference host;
- the shipped CLI still refuses `wasm32-wasi-command1`, exits without publishing an artifact, and has no production Kofun API for argv/environment/stdio/preopens/clock/random/exit;
- #1098 explicitly did not decide Kofun syntax, types, effects, authority flow, entrypoint, manifest CLI, error mapping, or backend lowering; and
- the browser sample is a core-Wasm host, not a WASI implementation.

The exact focused gates pass:

- `task wasm`
- `task wasm` (the browser sample is exercised inside it; no separate `wasm-browser` target exists)
- `task wasm-host-abi`
- `task wasm-host-profile`
- `task wasm-list-v1`
- `task wasm-object-arena`
- `task wasi-command-profile`
- `task capabilities`

## Completed foundations

- [x] Numeric module writer and Node differential adapter — #563.
- [x] Dependency-free browser sample — #566.
- [x] Bounded function lowering — #222.
- [x] AggregateLayout v1 — #120.
- [x] JavaScript host ABI/profile and object arena — #906, #1000, #1001.
- [x] Bounded Text/List lowering — #1002, #1004, umbrella #998.
[...]

PR #1414, opened 2026-08-14
selfhost: a B6 report can be valid and still close nothing (#1290)

hjosugi:
**This PR deliberately leaves #1290 open.**

Closing that issue would leave #1291 — whose only named blocker is #1290 — `blocked` with every named blocker closed, and `task backlog` on main fails on that within the minute. The ledger row that would excuse it cannot ride along, and the reason is a genuine circularity rather than an oversight:

- `task verify` runs `task backlog` against the **committed snapshot fixture**, where #1290 is still open, so a row predicated on its closure reads as "no longer applies" and fails the PR;
- refreshing that fixture cannot help, because #1290 is open until this merges;
- and closing #1290 first turns the **live** backlog job red until a row lands.

Every ordering has a red window except one: land the policy, leave #1290 open, and let whoever records #1291's exemption close it in the same change. I tried the other orderings and turned live `task backlog` red for a few minutes proving it.

Implements the decision recorded on #1290.

A B6 report can be structurally valid, pass `--against-checkout`, and close nothing. The packet's own `bootstrap/selfhost/b6/report.tsv` is exactly that — a real report whose builder is `kofun repository self-check`. `check-reproduction-report.sh` says in its own comments that it cannot authenticate a builder, which is correct; what was missing is anything saying which claims a reviewer must establish by other means, and which of them a gate can refuse.

## The policy

`bootstrap/selfhost/b6/POLICY.md` records the decision from #1290 — **an independent external party** — with both alternatives rejected in writing: a separately administered CI authority is still one organization and one control plane, and redefining B6 as producer-side reproducibility changes published meaning to fit available evidence. If no external builder materializes, **B6 stays open**.

Ten sub-decisions follow. The one worth pulling out is **freshness**: the key is `result|acquisition_set_sha256`, a digest over the declared input set — never wall time. A year-old attestation whose key still matches is current; a same-day one whose key moved is stale.

## The gate

`sh bootstrap/selfhost/check-b6-policy.sh` answers a different question from `selfhost-b6-report`: not "is this report well-formed" but "does it qualify as an attestation". It refuses

1. a claimed producer identity,
2. an empty or self-check `builder|basis`,
3. a missing or softened `builder|independence` line,
4. a stale acquisition set,
5. missing `provenance|` rows,
6. a `result_sha256` that disagrees with the producer's — as the **mismatch** case the policy says holds B6 open and opens an investigation, not as a malformed file.

Given a `REPORT` argument it judges that one file, which is what a reviewer with a submission in hand actually needs.

The producer boundary is `producer-identities.tsv`, not a list inside the script, so POLICY.md section 1 and the check are one fact. `substring` matching exists because agent identities are generated: `claude-1290-b6-independence-policy-20260815` is not a string anyone could have enumerated in advance.

## Two properties it is built around

**The packet's own report is the standing negative.** It is refused on every run. A fixture that ships with the packet cannot rot the way a hand-written one can, and a change that broke the refusal fails here rather than surfacing when somebody tries to close B6 with it.

**With no attestation on file the gate says so** — `PASS: B6 remains OPEN` — rather than passing silently. A gate that is quiet when there is nothing to check reads as "B6 is fine".

## Mutations, run rather than argued

| mutation | result |
| --- | --- |
| remove the `claude-` row from `producer-identities.tsv` | gate **fails** — the producer-identity fixture stops being refused |
| soften POLICY.md section 10 | gate **fails** the policy assertion |

The second one is worth reporting because **the first version of that assertion did not catch it**. It grepped for `B6 stays open`, which appears three times in the policy, so softening section 10 left two other occurrences and the check passed. It now asserts section 10's own sentence. I would have shipped an assertion measuring nothing if I had only read it.

## Validation

| Check | Result |
| --- | --- |
| `task selfhost-b6-policy` | exit 0, four PASS lines |
| `task selfhost-b6-report` | exit 0 — the existing packet gate is untouched |
| `task task-help` | exit 0 — new task grouped and described |
| `task repository-check` | exit 0 |

No capability row or release claim moves; B6 is still open and this PR says so in three places.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -183,6 +183,16 @@ tasks:
     cmds:
       - "sh bootstrap/selfhost/check-b6-report.sh"
 
+  # #1290. `selfhost-b6-report` answers "is this report well-formed"; this one
+  # answers "does it qualify as a B6 attestation", which is a different
+  # question with a different answer for the same file. The packet's own
+  # report.tsv is valid and is refused here, and that refusal is the gate's
+  # standing negative rather than a fixture that can rot.
+  selfhost-b6-policy:
+    desc: "Check the B6 independence policy and refuse reports that do not qualify as attestations"
+    cmds:
+      - "sh bootstrap/selfhost/check-b6-policy.sh"
+
   # RFC-0001 stage 1. The canonical contract and its bounded Stage 2
   # projection, pinned against each other. The seed grants no capability;
   # the gate refuses one appearing.
@@ -1129,6 +1139,7 @@ tasks:
             selfhost-diverse-double-compilation \
             selfhost-declared-inputs \
             selfhost-b6-report \
+            selfhost-b6-policy \
             selfhost-native \
             stage2 \
             stage2-events \
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -1140,6 +1140,7 @@ tasks:
             selfhost-declared-inputs \
             selfhost-b6-report \
             selfhost-b6-policy \
+            selfhost-b6-policy \
             selfhost-native \
             stage2 \
             stage2-events \
--- a/artifacts/release-evidence/index.json
+++ b/artifacts/release-evidence/index.json
@@ -770,7 +770,7 @@
   ],
   "evidence_digests": {
     ".github/workflows/native-hosts.yml": "9b0155c4caf6aeb29e7d942bffc10acd12cd61b79c4d70af45ea2b10a9967daa",
-    "Taskfile.yml": "826736672b12b52655361d8dcc5c00f3c07e42235fd240b9789391e3e1bec524",
+    "Taskfile.yml": "25b0c24bcda109676f8f953ffd836a81cbcaf363fdfcb72de6ab9b5aa5fb4dcb",
     "bootstrap/native/README.md": "6c470b203c3dbc327c6b601ab958c754e6f82be78c255d442307ad0c7691bc21",
     "bootstrap/native/check-macho64-signed.sh": "6fc51ddfec599ea85b8b06a5c16d68b6fdc994621cdb8ca4267e20be38141393",
     "bootstrap/native/check-macho64.sh": "d6bd52e3f6756aa607a2dc44a0117defeaaa55e64fa3f4221986da6406251e15",
--- a/artifacts/release-evidence/index.json
+++ b/artifacts/release-evidence/index.json
@@ -770,7 +770,7 @@
   ],
   "evidence_digests": {
     ".github/workflows/native-hosts.yml": "9b0155c4caf6aeb29e7d942bffc10acd12cd61b79c4d70af45ea2b10a9967daa",
-    "Taskfile.yml": "0357e0365c69a4158b682e01cab1f873b3dec8d72c8bc238e4b351e75b774374",
+    "Taskfile.yml": "826736672b12b52655361d8dcc5c00f3c07e42235fd240b9789391e3e1bec524",
     "bootstrap/native/README.md": "6c470b203c3dbc327c6b601ab958c754e6f82be78c255d442307ad0c7691bc21",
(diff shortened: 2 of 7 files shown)


hjosugi:
Blocking finding before this lands — **merging this as-is reddens `main`.**

I ran the closed-blocker sweep against what this PR closes (#1290) and #1291 names it:

    #1291   blockers: 1290=OPEN   other open blockers: []   <- none

`tests/backlog/check.mjs:248` rule 4 fails a `blocked` issue whose named blockers are all closed. The moment #1290 closes, #1291 is that issue, and the lane that catches it is **main-only** — the PR lane reads the committed fixture and will stay green right up to the merge. `Closes #1290` does not release it either.

#1291's `Blocked by:` line is `#1290, and an eligible external builder willing to run the packet and submit an attestation.` The prose half is the real remaining blocker and no issue number expresses it, so flipping #1291 to `ready` would be false — it is genuinely still blocked after this lands.

The body of #1291 already points at the right mechanism: *"The same exemption is carried by the WASI-capability and Stage-2-parse trackers."* That is `capability-blocker` in `tests/backlog/debt.tsv`, and rule 4b's comment names the same two precedents (#26, #570). So the fix is:

1. `- Blocked by: none` in #1291's body (the designed idiom — `blockedBy()` reads `/blocked by:/i` anywhere in the line, so no rewording escapes it), keeping the external prerequisite in prose.
2. A `capability-blocker` row for #1291 in `tests/backlog/debt.tsv`, **in this PR**, so the row lands in the same commit that closes #1290.

Step 2 is why this is yours and not mine to patch remotely: `debt.tsv` is committed, and the row has to be in the merge itself.

One thing worth weighing before you write it. `debt.tsv` currently holds **zero rows** — the header records that every row it has ever held was retired by fixing the issue, measured 2026-08-14. This would be the first row since. I still think it is correct: `capability-blocker` is documented as an exemption rather than drift, and the alternative is either a false `ready` or an unlanded PR. But it is a deliberate streak and you should be the one to break it knowingly rather than find out afterwards.

Happy to push the row to this branch if you would rather I did — say the word. Otherwise I will hold this PR out of the merge queue until it carries the row. #1419 and #1423 are ahead of it and all three contend on `artifacts/release-evidence/index.json`, so whoever lands first forces a rebase-and-regenerate on the rest.

hjosugi:
Correction to my earlier comment, and it does not clear this PR — it re-arms it for a different reason.

You removed `Closes #1290`, and I said this PR would redden `main` without checking whether that removal worked. It did not:

    $ gh pr view 1414 --json closingIssuesReferences --jq ".closingIssuesReferences[]?.number"
    1290

Every closing-keyword match in the whole body, case-insensitive:

    1:close #1290

That is line 1 — **"This PR deliberately does not close #1290."** GitHub matches `close #1290` and does not read the negation. **The sentence declaring the intent not to close is the closing link.** The keyword moved from the bottom of the body to the top and stayed active.

So the outcome you designed against still happens on merge: #1290 closes, #1291 is left `blocked` with every named blocker closed, and `tests/backlog/check.mjs` rule 4 fails on the post-merge lane — the one this PR cannot show you, because `task backlog` here reads the committed snapshot where #1290 is still open. Your analysis of the circularity in the body is right; only the mechanism you used to escape it is broken.

Fix: reword so no keyword precedes a `#N`. "This PR deliberately leaves #1290 open" works — `leaves` is not in the keyword set. `does not close`, `will not fix`, `should not resolve` all match, because only the verb and the reference are parsed. Then verify with `closingIssuesReferences` rather than by re-reading the body — reading is exactly what fails here, since the body says precisely what you meant.

My own error, recorded because it nearly cost more than it did: I ran `grep -oE "Closes #[0-9]+"` on this body, got no output, noted that it got no output, and then used #1290 from the PR title anyway. The evidence was in front of me and I overrode it with an assumption. The conclusion survived by luck, not by method — and grepping the body asks what the author wrote, while `closingIssuesReferences` asks what GitHub will do. Those differ here in both directions.

Issue #1421, opened 2026-08-14
Driver: reach the existing module resolver for a multi-file inventory


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: implementation
- Area: Modules / driver / resolution
- Priority: P1
- Size: L
- Parent: #902
- Blocked by: none
- Audited against: `origin/main@1a38d94b01b39cef0c06dec6db2808d131cea276`

## Goal

Make `bin/kofun` reach the module resolver that already exists, so a multi-file
module program produces resolved HIR. **No `extern "C"`, no linking** — this
step ends where the C-ABI half begins.

Step 2 of the three recorded in #1413's decision comment.

## Why this is wiring rather than a new resolver

Measured on `origin/main@1a38d94b01b39cef0c06dec6db2808d131cea276`:

```sh
wc -l bootstrap/stage2/imports_qualified.c        # 2008
grep -c 'imports_qualified\|imports_selective' bin/kofun   # 0
git grep -ln 'imports_qualified.c'
# bootstrap/stage2/README.md, bootstrap/stage2/imports_selective.c,
# tests/conformance/modules/{import-aliases,imports-qualified,raw-imports}/run.sh,
# tests/diagnostics/registry.tsv
```

2008 lines that load an inventory, order and validate it, compute identities,
resolve imports, reject cycles, resolve qualified bodies, and emit both HIR and
reference C — compiled only by conformance `run.sh` scripts. **The shipped
driver never invokes it.** That measurement is what selected option B in
#1413: module resolution does not have to be written, only reached.

## What this step must preserve

Both constraints are frozen by #1413 and are the reason the RFC-0012 chain
(#1211–#1216) exists:

- admission reads the **serialized** trust class, never the parse;
- the re-export refusal is **transitive by induction**, every link checked
  where it is written.

A path that resolved modules without those two would satisfy a fixture while
dropping the boundary the whole chain was built to enforce.

## Acceptance criteria

- [ ] `bin/kofun` accepts a multi-file module inventory and produces resolved
      HIR through the existing resolver, rather than reimplementing resolution.
- [ ] A two-module program that uses no C compiles and runs from the driver.
- [ ] `task kif-module-trust-profile raw-imports raw-re-exports re-exports kif-v1`
      passes, and the PR states which of its cases exercise the **new** path
      rather than the conformance harness. A case that only ever ran under
      `run.sh` is not evidence about the driver.
[...]

Issue #1413, opened 2026-08-14
[Decision] Where the one path lives that resolves modules, lowers extern "C" fn, and links a C library


hjosugi:
## Metadata

- State: ready
- Decision: option B, recorded in the 2026-08-15 decision comment; the smallest honest increment is stated there
- Tracker role: curated prerequisite decision
- Kind: spec-design
- Area: Modules / interop / toolchain surface
- Priority: P1
- Size: M
- Parent: #902
- Blocked by: none
- Audited against: `origin/main@5c5cf42404b0dbeea36a4de0fb457c4479e8bf68`

## Goal

Decide **where** the one path lives that resolves modules, lowers
`extern "C" fn`, and links a C library — because no such path exists, and
#1217's positive fixture cannot be built until one does.

This issue chooses between two shapes and freezes what either must preserve. It
implements neither.

## Current behavior and evidence

Measured on `origin/main@5c5cf42404b0dbeea36a4de0fb457c4479e8bf68` with the
repository's own driver:

```sh
cat > raw.kofun <<'EOF'
module lib.raw
trust raw-foreign

extern "C" fn puts(s: Text) -> Int
EOF

bin/kofun build raw.kofun
# error[E2S02]: expected top-level `fn`, `type`, or `let` at byte 0

bin/kofun build raw.kofun --backend c --c-abi -o out
# error[CABI001] at 1:1: expected `repr(C) struct`, `extern "C" fn`, or `fn main()`
```

The second refusal is the sharper one: the C-ABI front end rejects the `module`
header **and** the `trust raw-foreign` line — removing the module header and
leaving only `trust raw-foreign` produces the identical `CABI001` at 1:1. So the
module language and the C-ABI lowering are disjoint surfaces today, and neither
one is a place where a raw binding module can both resolve and link.

The module language currently lives only in the conformance drivers.

## Why this blocks #1217

#1217's positive fixture is *"two real modules plus a driver: raw bindings,
reviewed facade, and ordinary consumer; it builds and executes through the
existing C ABI"*. Every word of that needs the missing path. #1217 was labelled
`ready`, and a contributor starting from its body alone would have reached this
wall with no warning — the prerequisite was recorded only in a released claim
comment, not in the issue.

### A — teach the conformance resolver's reference C the C ABI

`imports_qualified.c` already emits a compilable C program from a resolved
multi-module inventory (`emit_reference_c`, consumed six times by
`tests/conformance/modules/imports-qualified`). It would need to parse
[...]

PR #1423, opened 2026-08-14
driver: a module-headed file is unsupported, not malformed (#1420)

### LINKED ISSUE #1420: Driver: refuse a module-headed file by name instead of E2S02 at byte 0
## Metadata

- State: in-progress
- Kind: implementation
- Area: Modules / driver / diagnostics
- Priority: P2
- Size: S
- Parent: #902
- Blocked by: none
- Audited against: `origin/main@1a38d94b01b39cef0c06dec6db2808d131cea276`

## Goal

`kofun build`, `check`, and `run` tell the author a module-headed file is
**malformed** when it is merely **unsupported**. Refuse it by name instead.

This is step 1 of the three recorded in #1413's decision comment, and the one
that is worth landing even if that decision is later reversed: a wrong
diagnostic is a defect under either option.

Measured on `origin/main@1a38d94b01b39cef0c06dec6db2808d131cea276`.

```kofun
module app.main

fn main() -> Int {
    0
}
```

```
$ bin/kofun build m.kofun
error[E2S02]: expected top-level `fn`, `type`, or `let` at byte 0
$ bin/kofun check m.kofun
error[E2S02]: expected top-level `fn`, `type`, or `let` at byte 0
$ bin/kofun run m.kofun
error[E2S02]: expected top-level `fn`, `type`, or `let` at byte 0
```

All three modes point at byte 0 and name the three things that *may* appear
there. The file is well-formed Kofun; `module` is simply a construct this path
does not implement. An author reading this looks for a typo in a line that has
none.

## The design point this issue decides

**The driver emits no registered diagnostic today.** Measured on the same
commit: `cut -f4 tests/diagnostics/registry.tsv | grep -c bin/kofun` is `0`,
against 181 rows whose emitters are all under `bootstrap/`.

So refusing in the driver introduces the first driver-owned code, and the
registry's columns have to be answered for a new emitter class: `channel`,
`exit`, and what `primary-span` means for a refusal the compiler never saw.

The alternative — refusing inside the Stage 2 pair, which owns `E2S02` — keeps
the emitter set unchanged but puts a *toolchain capability* statement inside
the parser, and requires the pair's four digests to move for a message that is
not about parsing.

**Recommended: refuse in the driver**, because the fact being reported is
"this path does not resolve modules", which the driver knows and the parser
does not. State the choice in the PR either way; it is the reusable half of
this issue.

## Acceptance criteria

- [ ] A module-headed file is refused by `build`, `check`, and `run` with a
      diagnostic that names the unsupported construct and does **not** claim a
      syntax error. It says what is unsupported, not merely that something is.
- [ ] The refusal names its own code, and that code has a row in
      `tests/diagnostics/registry.tsv` with every column answered for its
      emitter — including `exit` and `channel`, measured rather than assumed.
- [ ] A fixture and golden exist under the owner the registry row names, and
      `task diagnostics` passes with them.
- [ ] **The refusal is proved to refuse**: the fixture fails the gate when the
      guard is removed. A gate that has not been shown to fail is not accepted.
- [ ] A file with **no** module header compiles byte-identically to before, in
      all three modes. This is the regression that matters: the guard must key
      on the construct, not on anything that merely correlates with it.
- [ ] No `bootstrap/stage2/compiler.c`, `compiler.kofun`, `SHA256SUMS`, seed
      digest, or release artifact moves if the driver route is taken. If the
      pair route is taken instead, all four digests move together and the PR
      says so.
- [ ] No capability row or release claim advances — this issue buys a correct
      diagnostic and nothing else.

## What this issue does not do

It does not resolve a module, read an inventory, or make any module-headed
program compile. #1413's steps 2 and 3 do that, in that order.

## Validation

| --- | --- | --- |
| The refusal | `bin/kofun build`/`check`/`run` on the fixture | named refusal, not `E2S02` |
| Registry | `task diagnostics` | the new row and its golden pass |
| Mutation | remove the guard | the fixture's gate fails |
| Non-regression | a header-less file, all three modes | byte-identical output |
| Release truth | `task capabilities release-claims` | nothing advances |

hjosugi:
Closes #1420.

`kofun check`, `build`, `run`, and `emit-c` told an author their module-headed file was **malformed**. It is merely **unsupported**.

```
$ bin/kofun build m.kofun
error[E2S02]: expected top-level `fn`, `type`, or `let` at byte 0
```

The file is well-formed Kofun. `module` is a construct this path does not resolve, and the parser — which is the only thing that got to speak — can only describe that as a syntax error at byte 0. After:

```
$ bin/kofun build m.kofun
error[EDRV001]: `module` is not resolved on this path at byte 0
note: the file is well-formed; this command compiles a single file and does not resolve a module inventory
```

This is step 1 of the three recorded on #1413, and the one worth landing even if that decision is reversed: a wrong diagnostic is a defect under either option.

## Where the guard lives, and why not in `require_source`

`require_source` is the shared "is this a source file at all" predicate, called by every mode. The guard is instead on `emit_c`, the single-file compile path that `check`, `build`, `run`, and `emit-c` all funnel through.

The reason is the next step rather than this one. #1421 makes some path accept exactly these files. A guard in the shared predicate would have to be unpicked then; a guard on the single-file compile path is simply no longer reached.

## The registry columns were measured, not assumed

EDRV001 is the **first registered diagnostic whose emitter is `bin/kofun`** rather than a compiler under `bootstrap/` — `cut -f4 tests/diagnostics/registry.tsv | grep -c bin/kofun` was `0` against 181 rows. So it brings the first driver adapter with it, and every column had to be answered for a new emitter class:

| column | value | why |
| --- | --- | --- |
| `channel` | `stderr` | measured: all four modes wrote 66 bytes to stderr, 0 to stdout |
| `exit` | `2` | the driver already uses 2 for input it cannot take (`source must use the .kofun extension`); this is not a compile error |
| `artifact` | `forbidden` | measured: `emit-c` leaves no `.c` behind |

## What the adapter actually checks

The positive half is easy and nearly worthless on its own. A guard keying on the word `module` anywhere in the file, or on "the file failed to parse", would pass every positive case while refusing working programs.

So the adapter carries a control fixture that declares `fn module_count()` and **fails if it is refused**, and it drives all four modes by name rather than trusting the funnel — if a later change gives one of them its own path, the adapter fails instead of silently covering three and claiming four.

## Proof that the gate refuses

Removing the guard call and re-running the adapter:

```
FAIL: diagnostics driver: check status: expected -eq 2, got 1
```

## Non-regression, measured against the unpatched driver

Same driver with and without the patch, on three header-less files including one whose function name starts with `module`:

| file | exit | stdout | stderr | emitted C |
| --- | --- | --- | --- | --- |
| plain | 0 / 0 | identical | identical | **byte-identical** |
| leading comment mentioning `module` | 3 / 3 | identical | identical | both absent |
| `fn module_count()` | 3 / 3 | identical | identical | both absent |

## Gates

`task diagnostics repository-check task-help release-claims capabilities` all pass. `tests/diagnostics/registry.tsv` is hashed in the evidence pack, so `task release-evidence` was re-run and its commit is separate.

No `bootstrap/` file, `SHA256SUMS`, seed digest, or release artifact moves. No capability row or release claim advances.


hjosugi:
Pre-merge finding — this one is fixable without touching the branch, and I will handle it at merge time unless you object.

Closed-blocker sweep against #1420, which this PR closes:

    #1421   blockers: 1420=OPEN   other open blockers: []   <- none

So the moment #1420 closes, #1421 is `blocked` with every named blocker closed, and `tests/backlog/check.mjs:248` rule 4 reddens `main`. The PR lane will not show it — that lane reads the committed fixture; only the post-merge lane reads live GitHub.

Unlike #1291 on #1414, #1421 is genuinely unblocked by this work: #1420 is its only blocker and there is no prose half. So the fix is a remote issue edit and no commit:

- label `blocked` -> `ready`
- `- State: blocked` -> `- State: ready`
- `- Blocked by: #1420` -> `- Blocked by: none`

The third line is not optional and the ordering has a trap in both directions. Rule 5 fails a `ready` issue that *names an open blocker*, and rule 4 fails a `blocked` issue whose named blockers are all *closed* — so flipping the state alone is a failure before the merge and removing the line alone is a failure after it. Doing all three at once is safe in both windows, because a `ready` issue naming no blockers violates neither rule. `Blocked by: none` is the designed idiom for this (`blockedBy()` matches `/blocked by:/i` anywhere in a line, so a reworded line still counts as naming a blocker).

#1421 already carries a full 40-hex stamp (`origin/main@1a38d94b01b39cef0c06dec6db2808d131cea276`), so the `unstamped-ready` rule is satisfied by the flip.

I will make that edit immediately before merging, not after — flipping afterwards costs a red lane for everyone until someone notices, and a remote issue edit does not re-run the failed job.

Status: `Kofun verification` is still pending here. Note that #1419, #1423, and #1414 all touch `artifacts/release-evidence/index.json`, so the first to land forces a rebase-and-regenerate on the other two.

Issue #1409, opened 2026-08-14
Stage 2 spins instead of refusing when a Bool literal is a call argument (regression since e54e9e2c)


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: implementation
- Area: Compiler / Stage 2 / diagnostics
- Priority: P1
- Size: S
- Blocked by: none
- Audited against: `origin/main@5c5cf42404b0dbeea36a4de0fb457c4479e8bf68`

## Goal

`kofun build`, `kofun run`, and `kofun emit-c` **read before the start of the
source buffer** and then spin at 100% CPU without producing output or exiting,
when a program passes a `true` or `false` literal as a call argument to a
user-declared function. Restore the refusal, and add the fixture whose absence
let this ship.

The memory error is proven, not inferred: AddressSanitizer reports
`heap-buffer-overflow` at `compiler.c:485` in `skip_trivia`, reached from
`emit_product:9420`, on the reproduction below. See the 2026-08-14 sanitizer
comment for the full stack — it also corrects two claims in my earlier
comments about which sites are involved.

## Reproduction

69 bytes, no lambdas, no generics, one ordinary function:

```kofun
fn g(a: Int) -> Int {
    a
}
fn main() {
    let y = g(true)
    0
}
```

```sh
kofun emit-c hang.kofun /tmp/out.c
# no output, no exit. 119.66 s of user CPU at a 120 s cutoff; a second run was
# still at 100% CPU with an empty output file after 157 s, when it was killed.
```

Not proved non-terminating — proved to spend at least 157 seconds of CPU
producing nothing on a 69-byte input that the compiler answered in 2 ms until
2026-08-12. Either way it is unusable: `task verify` would not survive it, and
neither would a user.

**Expected**, and what the compiler did until 2026-08-12:

```
error[E2S12]: invalid Int expression at byte 54
# exit 1, 2 ms
```

## Current behavior and evidence

Measured on `origin/main@5c5cf42404b0dbeea36a4de0fb457c4479e8bf68`, building
`bootstrap/stage2/compiler.c` exactly as `bin/kofun`'s `ensure_stage2_compiler`
does (`cc -std=c11 -O2`):

| input | result |
| --- | --- |
| `g(true)` where `g(a: Int) -> Int` | **no output after 157 s of CPU** |
| `g(false)` | **no output after 8 s of CPU** (cutoff) |
[...]

PR #1429, opened 2026-08-14
stage2: E2S17 printed call_arity's -1 sentinel as an argument count (#1411)

### LINKED ISSUE #1411: E2S17 prints call_arity's -1 sentinel as an argument count: "expects 1 arguments, got -1"
## Metadata

- State: ready
- Kind: implementation
- Area: Compiler / Stage 2 / diagnostics
- Priority: P2
- Size: S
- Blocked by: none
- Audited against: `origin/main@5c5cf42404b0dbeea36a4de0fb457c4479e8bf68`

## Goal

`E2S17` prints `call_arity`'s failure sentinel as an argument count, so the
compiler tells an author their call has **minus one** arguments. Guard the two
sites that skip the check the third one already makes.

## Reproduction

```kofun
fn g(a: Int) -> Int {
    a
}
fn main() -> Int {
    let y = g(1 == 1)
    0
}
```

```
error[E2S17]: Core function `g` expects 1 arguments, got -1 at byte 61
```

The call has one argument. `call_arity` returns `-1` because it cannot read
`1 == 1` as an argument in this slice, and the diagnostic formats that sentinel
with `%PRId64` as though it were a count.

Measured on `origin/main@5c5cf42404b0dbeea36a4de0fb457c4479e8bf68`, building
`bootstrap/stage2/compiler.c` with `cc -std=c11 -O2`:

| call | message |
| --- | --- |
| `g()` | `expects 1 arguments, got 0` — correct |
| `g(1, 2)` | `expects 1 arguments, got 2` — correct |
| `g(1 == 1)` | **`expects 1 arguments, got -1`** |
| `g(1 + 1)` | no arity error; refused later as `E2S10` |

So the counter is right whenever it can count, and leaks its sentinel when it
cannot.

**This is not the #1409 regression.** The same reproduction produces the same
`got -1` on `ce416fb50f84e95a54bac7839cdc97de88697698`, two commits before the
hang was introduced, so it is older and independent — fixing #1409 will not
fix this, and vice versa. They share only a class: a failure sentinel used as
a value.

## The file already contains the guard

Three sites emit `E2S17`. The pipeline one checks:

```c
int64_t written = call_arity(source, open);
if (written < 0) return owned_text("ok");        /* compiler.c:10739 */
```

The other two do not:

```c
int64_t builtin_actual = call_arity(source, open);   /* compiler.c:11043 */
if (builtin_actual != builtin_expected) { ... }      /* no < 0 check */

int64_t actual = call_arity(source, open);           /* compiler.c:11114 */
...
if (actual != expected) { ... }                      /* no < 0 check */
```

`-1 != expected` is true for every real arity, so an unreadable argument list
always reports as a mismatch, and always with `-1` as the count.

## Scope

In scope: the two unguarded `E2S17` sites and a fixture that holds the
behavior.

Out of scope: making `1 == 1` a valid argument (Bool arguments stay refused);
changing the pipeline site, which is already correct; and #1409's
non-termination, which is a different defect at a different site.

## Acceptance criteria

- [ ] `g(1 == 1)` no longer reports an argument count of `-1`. Deferring the way
      the pipeline site does — so the call is refused later by the diagnostic
      that can name the unreadable argument — is the behavior the third site
      already demonstrates and the natural choice; a dedicated diagnostic is
      acceptable if the issue says why.
- [ ] `g()` still reports `got 0` and `g(1, 2)` still reports `got 2`,
      byte-exactly.
- [ ] A fixture covers an argument the slice cannot read, and is **proved by
      reintroducing the defect**: with the guard removed the fixture fails.
- [ ] No other `E2S17` wording or byte offset changes.
- [ ] `compiler.kofun` and `compiler.c` change together, and `SHA256SUMS`, the
      manifest seed digest, and the release evidence are regenerated by their
      owning commands.

## Validation

| --- | --- | --- |
| Stage 2 behavior | `task stage2` | the new fixture passes; existing arity messages unchanged |
| Diagnostics | `task diagnostics` | `E2S17` stays registered and owned |
| Call arguments | `task call-arguments` | the pipeline site's behavior is unchanged |

## Dependencies

- Blocked by: none.
- Changes the canonical Stage 2 pair, so it serializes against every other live
  compiler-pair issue; announce on the claim before starting.
- Related: #1409 is the same class — a scanner sentinel used as a value — at a
  different site, and is a regression where this is not.

hjosugi:
Closes #1411

`call_arity` returns -1 when the argument-list separator is neither `,` nor `)` — an unparsable argument, not a countable one. Five sites consume it. One, `validate_pipeline_calls`, already refused the sentinel with `if written < 0 { return "ok" }` on **both** halves of the pair. The other four printed it, so `len(1 == 1)` told the author their call had **minus one** arguments when the fault is the argument.

Each guard now defers to the expression checker, which names the real problem — `E2S12 invalid Int expression`, or `E2S15 ... expects TextOrList ... got Bool`.

## Four fixtures because the guards are independent

A fixture per site is not a conservative default here; it is the minimum that holds. Removing any single guard regresses exactly its own fixture and no other — a clean diagonal:

    guard removed    | fixtures that regress | verdict
    -----------------+-----------------------+---------------
    builtin          | builtin               | site-specific
    core_call        | core_call             | site-specific
    conversion       | conversion            | site-specific
    numeric_member   | numeric_member        | site-specific

This is the **opposite** of #1409, where four guards were mutually redundant and no single one was necessary — only the set was. Same count, opposite structure. The number of sites does not tell you the number of fixtures, and only the mutation matrix does. Each row was run by removing that guard from `compiler.c` and executing all four fixtures directly, not through `task diagnostics`, because failfast stops at the first mismatch and would have hidden rows 2..4.

The core-call guard reads `written`, not `actual`. A pipeline target adds one before the comparison, so a guard placed after the increment turns -1 into 0 — still wrong, and no longer obviously wrong.

## The corpus had pinned the defect as expected output

`tests/usability/check.sh` asserted the exact string `Core function `accumulate` expects 3 arguments, got -1`; `rubric.md` scored it **5.M5 = 0** with the note that it *"reports an argument count that cannot exist"*; the score table carried the same 0; `05_higher_order.kt` called it *"the worst diagnostic in this corpus"*; `README.md` repeated it. The defect was seen, measured, scored, and then written into a gate as the expected output — so fixing it is what makes that gate fail.

Row 5 is still refused and still for the same underlying reason: a function-typed parameter cannot be declared in Stage 2. Only the message changed, to `error[E2S12]: invalid Int expression at byte 1211`. Re-scored **5.M5 = 1**, not higher, for the same reason `6.M5` is held at 1 — it names the location but does not say that a function-typed argument is a known, accepted, unimplemented form. Prose entry and score table updated together; nothing cross-checks the two, so a prose-only edit would have left a stale cell.

That promotes row 2 to the corpus's only remaining 0, and the README now says so. Row 2's is not a Kofun diagnostic at all — a host C compiler error against generated code the author never wrote.

## Pair note

`compiler.c` has five E2S17 sites and `compiler.kofun` has four; the numeric-conversion arity check exists only in C. That asymmetry is **pre-existing** — `origin/main` has the same 4/5 — and is a second concrete instance for #1401 after #1315's `is_main`. Both halves are now guarded at every site they have.

Worth recording for #1401: I went looking for a route that builds a compiler from `bootstrap/stage2/compiler.kofun` so the canonical half could be tested behaviourally, and **there is none**. `bootstrap/stage2/check.sh:592`'s `round_trip` feeds it to the compiler as *input* and checks the re-emission is byte-identical; every other reference in the tree is a digest, a `registry.tsv` documentation column, or a grep in a conformance script. `compiler.kofun` is an input to the compiler, never the compiler.

## Verification

`task verify` exit 0 on this branch, clean run. An earlier run reported `selfhost fixed point: two normalized clean runs disagree` — that run was contaminated, because I executed other gates in the same worktree while it was going, which is a serialization rule I had written down and broken. Re-run with the worktree untouched and it passes; I am reporting the clean run and not the first one.

Rebased onto `main` after the last merge, then `task release-evidence` regenerated with an empty diff.


hjosugi:
Reviewed. The fix is right where it is hardest to get right, and there is one gap I can name precisely because I have the site-tagged probe still built.

**The 11133 guard is correct.** You introduced `written` and test `written >= 0 && actual != expected`, so the guard reads the raw `call_arity` result rather than the value after `if (pipeline_subject_for_call(...)) ++actual`. A guard placed after that increment would have turned `-1` into `0` and changed the message from `got -1` to `got 0` — still wrong, and no longer obviously wrong. That was the one place where the obvious patch is the wrong patch, and you did not take it.

**Four fixtures, one per site, is the right shape** and matches what the reproductions showed: `Decimal.divide(1 == 1)` and `Decimal.round(1 == 1)` both reach 23450, so a single numeric-member fixture covers that site, and the other three each need their own.

**The gap: only one of the four sites has a still-refuses case.** The guards suppress the arity check whenever the count is unreadable, so each one could equally have disabled its site outright, and four sentinel fixtures cannot tell those apart — they pass either way. The repository has exactly one fixture asserting a genuine arity error, `e2s17_function_arity.kofun`, and I measured which site it reaches with a build that tags each `E2S17` format string with its own location:

```
error[E2S17]{CORE11133}: Core function `add` expects 2 arguments, got 1 at byte 160
```

**11133 only.** Sites 11067, 23350 and 23450 now have a guard and nothing that fails if the guard swallows real mismatches. From the measurements on #1411, these are the three inputs that pin them, with the message each currently produces:

| site | input | still expected |
| --- | --- | --- |
| 11067 builtin | `len("a", "b")` | ``E2S17: … `len` expects 1 arguments, got 2`` |
| 23350 conversion | `Decimal.from_int(1, 2)` | ``E2S17: … `Decimal.from_int` expects 1 arguments, got 2`` |
| 23450 numeric member | `Decimal.round(1, 2)` | ``E2S17: … `Decimal.round` expects 3 arguments, got 2`` |

Three fixtures, and they are cheap. Without them the mutation "delete the arity comparison at this site" passes the whole suite.

Everything else reads correct to me: the Kofun and C halves guard the same four sites, `&&` is fine there (411 existing uses in `compiler.kofun`, since Stage 1 translates it), and the four digests move together as they must.

Issue #1422, opened 2026-08-14
Driver: lower extern "C" fn and honour --link-library on the driver path


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: implementation
- Area: Modules / interop / C ABI
- Priority: P1
- Size: L
- Parent: #902
- Blocked by: none
- Audited against: `origin/main@1a38d94b01b39cef0c06dec6db2808d131cea276`

## Goal

Lower `extern "C" fn` and honour `--link-library` on the path the driver
actually runs. This is step 3 of #1413's decision, and **the half that exists
nowhere today**.

## Evidence that it exists nowhere

Measured on `origin/main@1a38d94b01b39cef0c06dec6db2808d131cea276`:

```sh
grep -cE 'extern "C"|c-abi|link' bootstrap/stage2/imports_qualified.c
# 0
```

The 2008-line resolver that steps 1 and 2 reach does module resolution and
nothing about the C boundary. So unlike step 2, this step is not wiring: it is
the work.

## Why this ordering, and not the fixture-first shape

#1217 exists to enforce a **mechanical raw-binding import boundary**. Proving
its positive fixture against a resolver `bin/kofun` never calls would let #902
close while a user running `kofun build` on the same three files still gets
`E2S02` — the evidence and the claim would be about different programs. That
argument is #1413's, recorded in full there.

## Acceptance criteria

- [ ] `extern "C" fn` declarations lower on the driver's path, and a program
      calling one links and runs via `--link-library`.
- [ ] The **serialized** trust class still governs admission, and the re-export
      refusal is still transitive by induction — both proved by mutation, each
      failing for its own reason, on the driver path rather than under
      `run.sh`.
- [ ] A program that declares `extern "C" fn` but is built **without** the
      library is refused by name, not by a linker error the author cannot map
      back to their source.
- [ ] #1217's positive fixture builds and runs through `bin/kofun`, and the PR
      states the command a user would type — not the harness invocation.
- [ ] Nothing about the C-ABI boundary is asserted by a test that only the
      conformance harness can reach.
- [ ] Any capability row or release claim that advances is named in the PR with
      the evidence that carries it.

## Unblocks

#1217, and through it #902's remaining criteria.

## Validation

| Check | Command | Expected result |
| --- | --- | --- |
| The boundary | `task kif-module-trust-profile raw-imports raw-re-exports re-exports kif-v1` | green, driver-path cases named |
[...]

PR #1430, opened 2026-08-14
driver: resolve a module program instead of refusing it (#1421)

### LINKED ISSUE #1421: Driver: reach the existing module resolver for a multi-file inventory

hjosugi:
Closes #1421.

`kofun run` on a two-module package now returns its value:

```
$ tree demo
demo/app/main.kofun    # module app.main / import lib.math / return math.identity(42)
demo/lib/math.kofun    # module lib.math / pub fn identity(value: Int) -> Int

$ bin/kofun run demo/app/main.kofun; echo $?
42
```

## The resolver was already there; the producer was not

2008 lines in `bootstrap/stage2/imports_qualified.c` load an inventory, order and validate it, resolve imports, reject cycles, and emit **reference C with its own `main`** — compiled only by conformance `run.sh` scripts. `grep -c 'imports_qualified' bin/kofun` was `0`. That measurement is what selected option B on #1413, and it held: module resolution did not have to be written, only reached.

What did not exist is the **inventory producer**. The resolver's input lines begin with three 32-byte identities, and nothing shipped derived them — `module_symbols.c` *parses* identities, and the only deriving code is `tests/conformance/modules/imports-qualified/identity_reference.c`, in the test tree. That is why every conformance harness writes `1111…`/`2222…`: not laziness, but the absence of a producer.

## The identities are checked against the specification, not against themselves

`bootstrap/stage2/module-inventory.sh` implements what `spec/modules/module-identity.md` and `spec/modules/source-file-mapping.md` already specify. Its gate asserts the package and module identities equal **the specification's own reference vector**:

| | value |
| --- | --- |
| package id, `spec/module-identity/check.sh` golden | `eb09a959d6f2a46f…` |
| package id, this producer | `eb09a959d6f2a46f…` |
| module id, spec golden | `6afda3bdd9875c08…` |
| module id, this producer | `6afda3bdd9875c08…` |

If that gate's goldens move because the scheme changes, this one fails too. That coupling is the point: a producer checked against numbers it generated itself would prove only that it is deterministic.

Four mutations, each refused:

| mutation | caught by |
| --- | --- |
| drop the domain from the hash framing | package id no longer matches the spec vector |
| use `kofun.id.package/v2` | same, with a different value |
| omit the logical path from `FileId` | the moved-file property |
| accept a file that declares no module | the refusal case |

## The package root rule

Derived by subtracting the declared module path from the file's path: `module app.main` in `<root>/app/main.kofun` yields `<root>`. That is the specification's correspondence read backwards, and it is deliberately the only rule — a root guessed from the working directory would make identical sources compile to different identities depending on where the compiler ran, because the logical path participates in `FileId`.

## EDRV001 is retired, not reworded

It said ``module` is not resolved on this path`. That sentence is no longer true, and rewording a diagnostic to keep its number is how a registry stops describing the program. What is still true is that a declaration must agree with its location, so **EDRV002** carries that, and the fixture and golden move with it.

The guard EDRV001 lived behind was placed on the single-file compile path rather than in `require_source` specifically so this step would not have to unpick it (#1420). It did not — the guard was simply no longer reached, and could be deleted rather than worked around.

One incidental fix: the golden quotes the path it was given, so the adapter now invokes from the repository root with a relative path. An absolute one would have written this machine's directory layout into checked-in evidence.

## Scope, and what this does not do

No `extern "C"`, no `--link-library` — that is #1422, and it is the half that exists nowhere. A single-file program without a module header is unaffected: `task diagnostics` passes, and the control fixture declaring `fn module_count()` is still compiled rather than refused.

`task module-inventory diagnostics task-help release-claims capabilities` all pass. `tests/diagnostics/registry.tsv` is hashed in the evidence pack, so it was regenerated.


hjosugi:
Edge cases a reviewer would ask about, measured rather than asserted.

**The resolver's own refusals reach the user unchanged.** Both of these are diagnostics that previously existed only inside the conformance harness, and they now come out of `kofun build`:

```
$ kofun build cyc/a/one.kofun
error[E2S64]: canonical import cycle: a.one --a/one.kofun:13..25--> b.two --b/two.kofun:13..25--> a.one; hint: remove one import edge from this cycle

$ kofun build miss/app/main.kofun
error[E2S60]: import `lib.absent` has no module in the current package at `app/main.kofun` bytes 16..33
```

Note what the spans are relative to: `a/one.kofun:13..25`, the **logical** path. That is the resolver's own framing and it survives the wiring, which is the property that matters — the user sees positions in their source, not in a temporary inventory.

**Determinism.** Two `emit-c` runs over the same tree produce byte-identical C. The inventory is `LC_ALL=C sort`ed before it is handed over, because `find` order is a filesystem property and would otherwise leak into identities.

**The package name participates.** Copying the same sources to a directory with a different basename produces different C, because the package name is in `PackageIdPayload` and therefore in every `ModuleId` and `FileId`. That is the specification's intent rather than an artefact, and it is the reason v1 takes the name from the root directory rather than defaulting it silently: moving a package is a change of identity, and it should be visible.

**What this does not decide.** v1 reads no package manifest, so `version`, `edition` and `source` take the specification's reference values. A manifest reader is a later change and it will move every identity in a package that gains one — worth saying now, because that is a migration rather than a bug when it happens.

PR #1444, opened 2026-08-14
stage2: assert a function-count for the compiler, derived like Stage 1's

hjosugi:
Found while auditing gates for the "remove redundant verification" half of this session's work. It is the opposite of a removal: the compiler was the one subject in `bootstrap/stage2/check.sh` without a completeness assertion.

`check.sh` asserts a `function-count` for `fixture.ir` and for `stage1.ir`. It did not for `stage2.ir` — the canonical Stage 2 compiler, the largest of the three and the one that changes most.

Stage 1's is *derived*, with a comment stating the principle:

    # One function record per declaration in S; derived so a helper split in S
    # does not strand a hand-maintained count here.
    stage1_functions=$(grep -c "^fn " "$root/bootstrap/stage1/compiler.kofun")

The same reasoning applies to Stage 2 and had not been applied — the principle was written down two blocks above the place it was missing.

## What the compiler's block asserted instead

Nine specific function records, by name and arity: `lex|1`, `parse_program|1`, `lower_c|2`, `compile_file|4`, and five others. Those are worth keeping — they pin the arity of nine functions — but **they cannot see a function that stopped being emitted** unless it happens to be one of the nine. The compiler declares 463.

## Proved live rather than read

An assertion added beside nine passing ones is exactly the shape that can be inert without anyone noticing, so:

    expected count + 1  ->  FAIL: stage2: stage2.ir: no match for: grep ^function-count|464$
    restored            ->  exit 0

So the IR really does emit `function-count|463`, and the assertion really does compare against it.

`sh bootstrap/stage2/check.sh` exit 0. One file, ten lines, no generated artifact moved.

### DIFF
--- a/bootstrap/stage2/check.sh
+++ b/bootstrap/stage2/check.sh
@@ -590,6 +590,16 @@ assert_grep "stage1.ir" \
     "^function-count|$stage1_functions\$" "$temporary/stage1.ir"
 
 round_trip stage2 "$stage2/compiler.kofun"
+# One function record per declaration in the canonical Stage 2 source, derived
+# for the same reason the Stage 1 count two blocks up is derived: a
+# hand-maintained number strands the moment someone splits a helper. Stage 2 is
+# the subject that most needs it and was the only one of the three without it —
+# `fixture.ir` and `stage1.ir` both assert a count, and the compiler itself did
+# not, so a function that stopped being emitted would have been caught only if
+# it happened to be one of the nine named below.
+stage2_functions=$(grep -c '^fn ' "$stage2/compiler.kofun")
+assert_grep "stage2.ir" \
+    "^function-count|$stage2_functions\$" "$temporary/stage2.ir"
 assert_grep "stage2.ir" '^function|lex|1|' "$temporary/stage2.ir"
 assert_grep "stage2.ir" '^function|parse_program|1|' "$temporary/stage2.ir"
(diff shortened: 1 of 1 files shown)


hjosugi:
Holding this out of the merge queue until #1447 lands, even though it is green.

Its `Kofun verification` passed because the runner was still on 2026-08-14 UTC. `task rfc-registry` fails on clean `origin/main` now that the date has rolled over — a mutation fixture hard-coded `2026-08-15` as "a date in the future", so the future-date rule it relied on stopped firing and the mutation was accepted. Reproduced on the clean tree before believing it:

    clean base 1043f1ade66c
    task rfc-registry                             exit 201
    KOFUN_RFC_TODAY=2026-08-14 task rfc-registry  exit 0

So this PR is green on a calendar that has passed, and merging it would put `main` at a commit that fails for a reason this change has nothing to do with — attributed to this merge. That already happened once tonight: the first red run of the bug landed on `1043f1ad`, my #1442 merge, which was equally unrelated.

Nothing to change here. #1447 fixes it, and after it lands I will rerun this rather than merge on a stale green.

hjosugi:
Superseded by a rebuild on `2e9fa1d1` (post-#1447), since this branch was green only on a runner still dated 2026-08-14. Same ten lines, same mutation proof; reopening as a fresh PR rather than force-pushing so the stale-green reasoning stays readable.

Issue #1431, opened 2026-08-14
A ready issue with a live claim, and a claim that parses as nothing, are both invisible


hjosugi:
## Metadata

- State: in-progress
- Kind: test-quality
- Area: Backlog gate
- Priority: P1
- Size: S
- Blocked by: none
- Audited against: `origin/main@5832bf659a2c7c40627045da60582e8973bc86a7`

## Goal

Close two adjacent holes in `tests/backlog/check.mjs` that let an issue advertise one thing and mean another with nothing red to show for it. Both have instances in the tree today.

## Current behavior and evidence

Three rules read claims, at `check.mjs:153`, `:158`, and `:304`:

1. more than one live claim on an issue,
2. a **closed** issue with a live claim,
3. an **in-progress** issue with no live claim.

**Nothing reads `label === 'ready' && liveAgents.length > 0`.** The claim rules only walk `in-progress` and `closed`; the ready rules look at blockers and stamps and never at claims. So an open `ready` issue carrying a live `active` claim passes every rule.

That state occurred on **#1315** earlier today and I put it there. I released the issue to `ready` because it gates fourteen dependents, but my release comment was in a form the extractor does not read, so it did not supersede my own earlier canonical `active` block. The issue sat `ready` with a live claim — advertised as free, owned by me — and **nothing failed**. #1242 had the mirror state, `in-progress` with no claim, and failed loudly within the minute. The louder of the two is the one that was already caught.

Measured just now across all 105 open issues, after I fixed #1315:

    ready-with-live-claim:  0
    inert claim comments:   3   (#1315, #1242, #847)

Rule 1 is at zero *because* the instance was fixed an hour ago. It is not hypothetical.

The second hole is already named in the file's own comment at `:299`:

> *Note what this cannot see. An inert claim — a wrapped HTML comment, or the marker followed by prose keys — extracts as nothing, so it is indistinguishable here from never having claimed. **That is the worse failure of the two, because its author believes the signal is published.***

I am the demonstration. Every claim I posted in one session used the bare `agent-claim:v1` line with no `###` heading and no `- ` keys — the form `docs/ISSUE_READINESS.md` line 94 names explicitly as not a claim. It renders identically, greps identically, and I verified my own claims by grepping for the marker all session. **#847 carries one too and is not mine**, so this recurs across agents.

[...]

Issue #1217, opened 2026-08-11
RFC-0012 steps 5-6: integrate bindgen raw modules and reviewed facade


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: implementation
- Area: Tooling / Modules / Interop
- Priority: P1
- Size: M
- Parent: #902
- Blocked by: none
- Audited against: `origin/main@8fd0b25497c1fd4c61085260a416af619beb3c3f`

## Goal

Switch bindgen-c from an advisory raw marker plus concatenated driver to one declared raw module and one hand-written reviewed facade, then enforce the complete RFC-0012 boundary end to end.

RFC-0012 implementation-plan steps 5 and 6 are atomic here. Splitting them would make the generator emit a mandatory module header while the existing green C-ABI gate still compiles it as anonymous concatenated source, forcing a temporary bypass or weakening. This issue adds neither.

## Current behavior and evidence

Re-measured 2026-08-14 on `origin/main@5c5cf42404b0dbeea36a4de0fb457c4479e8bf68`.
Every claim below is now a command, because the prose form could only be
believed, not checked:

```sh
git grep -l 'raw-trusted-foreign' -- ':!graphify-out/**' | wc -l
# 9 — unchanged since the audited commit

git grep -c 'trust raw-foreign' -- tooling/bindgen-c
# no matches: the generator still emits no trust or module header

git ls-files '*.raw.kofun' | wc -l
# 0

test ! -d tests/interop/bindgen-c/import-boundary
# true
```

The premise holds in full. Note what has moved *around* it since the audited
commit: the language half of the boundary landed with #1211/#1212/#1214/#1215/#1216,
so `^trust ` and `^trusted import` now match in
`tests/conformance/modules/` fixtures and are gated by `task raw-imports`,
`task raw-re-exports`, and `task kif-module-trust-profile`. The generator is
the one place that has not moved, which is exactly this issue.

The original prose, kept because the last bullet is not a one-line command:

- `raw-trusted-foreign` appears in nine tracked files including RFC-0012, eight excluding it;
- the generator emits a banner/comment and report field but no `module` or `trust raw-foreign` header;
- no tracked `*.raw.kofun` artifact exists;
- `tests/interop/bindgen-c/import-boundary/` is absent;
- the existing gate concatenates generated declarations with `driver.kofun` and passes `--c-abi`; it never imports a module.

## The prerequisite, and why this is blocked

[...]

PR #1445, opened 2026-08-14
extern "C": lower C scalars to their C spellings (#1443)

### LINKED ISSUE #1443: Lower C scalar types on the module path, or restrict extern "C" to Int
## Metadata

- State: in-progress
- Kind: implementation
- Area: Modules / interop / C ABI
- Priority: P2
- Size: M
- Parent: #902
- Blocked by: none
- Audited against: `origin/main@06ad2fcbaefd01612da41f462ccd0f73a3eac3f1`

## Goal

#1422 lowered `extern "C" fn` on the module path for **`Int`-only** signatures.
`kofun bindgen-c` generates `CInt`, so the two do not meet, and #1217's positive
fixture needs them to.

## Measured

```sh
$ kofun bindgen-c demo.h --out-dir out --module demo
$ grep -vE '^#|^$' out/demo.raw.kofun
extern "C" fn kofun_double(value: CInt) -> CInt
```

The module path's reference emitter lowers every parameter and result to
`int64_t`; `CInt`, `CStr` and `repr(C)` are understood only by
`bootstrap/c_abi/compiler.c`, the single-file path. `grep -c 'CInt\|CStr'` in
`bootstrap/stage2/imports_qualified.c` is 0.

## The decision this issue makes

**Option A — lower C scalars on the module path.** `CInt`, `CUInt`, `CLong`,
`CChar` and friends map to their C spellings in the emitted prototype instead of
`int64_t`. Bounded and mechanical for scalars; `CStr` and `repr(C)` aggregates
are explicitly out and get their own issue.

**Option B — restrict the boundary to `Int` and make bindgen say so.** The
generator refuses a header whose signature does not map to `Int`, naming the
type it could not take. #1217's fixture then uses an `Int`-only header.

**Recommended: A, scalars only.** B makes the generator refuse most real headers
— `int` is the commonest C parameter type there is — so the fixture would prove
the boundary on a program no user would write. A is the smaller surprise and the
smaller lie.

Whichever is chosen, the PR states it, because the other becomes wrong.

## Acceptance criteria

- [ ] A module declaring `extern "C" fn f(value: CInt) -> CInt` builds, links
      against a C library exporting `int f(int)`, and runs.
- [ ] The emitted prototype uses the C spelling, not `int64_t` — asserted
      against the emitted text, since a wrong width links successfully and
      misbehaves only at some argument values.
- [ ] A type the slice does not lower is refused **by name**, naming the type.
      `CStr` is the case to use.
- [ ] Mutation: emitting `int64_t` for a `CInt` parameter fails the gate. A
      value-based test would pass on small integers, so this must compare the
      emitted spelling.
- [ ] `task extern-c raw-imports imports-qualified` stay green, and the
      `Int`-only program from #1422 still emits byte-identical C.

## Unblocks

#1217's positive fixture, which uses `CInt` today.

## Validation

| --- | --- | --- |
| The new case | `task extern-c` | a `CInt` program links and runs |
| Non-regression | `task extern-c` | the `Int` program's C is unchanged |
| Refusal | a `CStr` signature | refused by name |

hjosugi:
Closes #1443. Option A — lower C scalars — for the reason on the issue: option B would make `kofun bindgen-c` refuse most real headers, since `int` is the commonest C parameter type there is.

```
extern "C" fn kofun_cdouble(value: CInt) -> CInt
→  int kofun_cdouble(int k_p0);
```

Links against a C library exporting `int kofun_cdouble(int)` and returns 42.

## The assertion is on the emitted text, not the result

A `CInt` parameter emitted as `int64_t` **links successfully and reads the wrong width**. For 21 it still returns 42. A value-based test passes on every argument a fixture would think to try, so the gate compares the spelling:

```sh
assert_grep "a CInt parameter lowers to int" -Fx "int kofun_cdouble(int k_p0);" emitted.c
```

`CStr` and `repr(C)` aggregates are refused by name. A pointer emitted as an integer is the same silent-width mistake one level worse.

## Two guards, and the gate names both

The signature check (`E2S65`) runs first; the emitter (`E2S175`) is the backstop.

**Measured**: disabling the signature check alone still refuses, via the emitter. So an assertion naming only `E2S65` would have reported a working guard with one of the two removed. The gate asserts `error[(E2S65|E2S175)]`, and disabling all four sites is what fails it.

`E2S175` is therefore unreachable in practice and stays **unregistered**: no fixture can produce it, and a registry row whose observation cannot exist is one the registry could never satisfy. That reasoning is written at the code rather than left to be rediscovered.

## A mutation of mine missed, and I nearly recorded a gap

I disabled "both" guards and the suite stayed green. The emitter has **two** call sites and my `sed` reached one. An uncaught mutation is a claim about the mutation before it is a claim about the gate — the same lesson as #1431's third case, hit again in the same session.

| mutation | result |
| --- | --- |
| lower `CInt` to `int64_t` | caught — `a CInt parameter lowers to int` |
| admit any type in the signature check | **passed** — the emitter refuses instead |
| disable all four guard sites | caught — `the refusal names the type` |

## Scope

Scalars only: `Int`, `CChar`, `CSChar`, `CUChar`, `CShort`, `CUShort`, `CInt`, `CUInt`, `CLong`, `CULong`, `CLongLong`, `CULongLong`. The `Int`-only program from #1422 emits byte-identical C.

`task extern-c imports-qualified imports-selective raw-imports raw-re-exports re-exports module-symbols kif-v1 module-inventory diagnostics` all pass.

Unblocks #1217's positive fixture, which uses `CInt`.

### DIFF
--- a/bootstrap/stage2/imports_qualified.c
+++ b/bootstrap/stage2/imports_qualified.c
@@ -1201,6 +1201,95 @@ static size_t function_arity(const Program *program, const Declaration *declarat
     return SIZE_MAX;
 }
 
+/*
+ * #1443. C scalar spellings for an `extern "C"` declaration.
+ *
+ * The reference emitter lowers ordinary Kofun functions to `int64_t`, which is
+ * right for symbols it defines. For a symbol the *linker* provides, the width
+ * has to be the one the C library actually compiled — a `CInt` parameter
+ * emitted as `int64_t` links successfully and misreads the argument, which is
+ * the failure mode that shows up at some argument values and not others.
+ *
+ * Scalars only. `CStr` and `repr(C)` aggregates are refused by name rather than
+ * guessed at: a pointer emitted as an integer is the same silent-width mistake
+ * one level worse.
+ */
+typedef struct {
+    const char *kofun;
+    const char *c;
+} CScalarType;
+
+static const CScalarType C_SCALAR_TYPES[] = {
+    { "Int", "int64_t" },
+    { "CChar", "char" },
+    { "CSChar", "signed char" },
+    { "CUChar", "unsigned char" },
+    { "CShort", "short" },
+    { "CUShort", "unsigned short" },
+    { "CInt", "int" },
+    { "CUInt", "unsigned int" },
+    { "CLong", "long" },
+    { "CULong", "unsigned long" },
+    { "CLongLong", "long long" },
+    { "CULongLong", "unsigned long long" },
+};
+
+static const char *c_scalar_spelling(const char *name, size_t length) {
+    size_t index;
+    for (index = 0u; index < sizeof C_SCALAR_TYPES / sizeof C_SCALAR_TYPES[0]; index += 1u) {
+        const char *kofun = C_SCALAR_TYPES[index].kofun;
+        if (strlen(kofun) == length && memcmp(kofun, name, length) == 0) {
+            return C_SCALAR_TYPES[index].c;
+        }
+    }
+    return NULL;
+}
+
+/*
+ * The declared type of parameter `index`, or of the result when `index` is
+ * SIZE_MAX, as a span into the module source. Walks the same tokens
+ * `function_arity` does, so the two cannot disagree about where the signature
+ * is.
+ */
+static bool declared_type_span(
+    const Program *program,
+    const Declaration *declaration,
+    size_t index,
+    const char **text,
+    size_t *length
+) {
+    const Module *module = &program->modules[declaration->module_index];
+    size_t cursor;
+    for (cursor = 0u; cursor < module->token_count; cursor += 1u) {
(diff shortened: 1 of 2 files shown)


hjosugi:
**This PR's CI failure is not this PR.** `task rfc-registry` fails identically on clean `origin/main`:

```sh
$ git checkout origin/main && task rfc-registry
FAIL: rfc ledger: accepted generic mutation proposed-claiming-a-decision-date
exit 201
```

A mutation fixture hard-coded `2026-08-15` as "a date in the future". Today is
2026-08-15, so the future-date rule it relied on stopped firing. Fixed in #1447,
which also runs the corpus at 2027 and 3000 so the next expiring fixture fails
the day it is written rather than the day it expires.

Rebasing this once #1447 lands.

PR #1488, opened 2026-08-15
verify: make #1205's last measurement a standing property (#1485)

### LINKED ISSUE #1485: verify: the two-method final census and timing #1205 reserved
## Metadata

- State: in-progress
- Kind: maintenance
- Area: Quality / Build system
- Priority: P2
- Size: M
- Parent: #1205
- Blocked by: none
- Audited against: `origin/main@d42d2e4c98d66ece506802fbaaa83419f604b8bd`

## Goal

Run the two-method final census and timing that #1205 reserved for itself, and
record the result against the measured baseline in that umbrella.

Every preplanned migration has merged: #1238 (foundation), #1319 (discovery
sanitizer support objects), #1326 (fuzz compiler profile), #1239 (strict-O2
KIF/Unicode/SHA-256/visibility), and #1449 (the `-O0 -fanalyzer` family). The
two false-green evidence paths the first lossless census exposed, #1338 and
#1339, are both closed. Nothing is left to migrate; what is left is to measure
what the migrations achieved and to say so in one reviewable place.

## Why this is a child rather than the umbrella itself

#1205 says it "is never directly implementation-ready" and, in the same body,
that it "is now claimed only for the two-method final remeasurement". Those two
sentences cannot both hold of one issue: the second describes bounded work with
a deliverable, which is what a child is for.

Measured on `origin/main@d42d2e4c98d66ece506802fbaaa83419f604b8bd`: all 16 open
`planning` umbrellas carry `blocked`, each naming an open child. None is
`ready`. Leaving #1205 as the thing that performs the remeasurement would have
made it the first `planning` issue in a state no planning issue has ever held,
on the day its last blocker closed.

## Scope

In scope:

- the two-method census — full-path source sets and effective compiler argv,
  never basenames, which is the aggregation error the umbrella's own baseline
  section records;
- a timing comparison against the recorded baseline (suite wall 1658 s, C
  compilation 1075 s, repeated already-seen source sets 872 s), stating the
  conditions of each run, because `task verify` has measured 3570 s and 1608 s
  on identical input on one machine and an uncontexted wall-clock number is
  unfalsifiable;
- one reviewable artifact recording both methods and their disagreements, if
  any;
- updating #1205's baseline table with the result and closing that umbrella.

Out of scope:

- any further performance migration. The umbrella states that none is
  authorized, and a census that also changes what it measures cannot report
  what the migrations achieved.
- a persistent or cross-run compiler/object cache, mtime database, signing
  claim, or authentication claim, all of which the umbrella's cross-slice
  contract already forbids.

## Acceptance criteria

- [ ] Both methods are run on one tree and their results recorded together,
      with the conditions of each run stated.
- [ ] The census classifies full source paths and effective argv; a basename
      aggregation is shown to differ, or shown not to, rather than assumed.
- [ ] The result is compared against the umbrella's recorded baseline, and
      #1205's table is updated with it.
- [ ] Coverage is unchanged: warning, sanitizer, analyzer, compiler-diversity,
      semantic, and gate coverage each still measured, not asserted.
- [ ] #1205 closes, or names what remains.

## Validation

| --- | --- | --- |
| Census | `task verify-object-reuse` | green; the census it reads is the full run's |
| Backlog | `task backlog` | green with #1205 closed or renamed |

hjosugi:
Closes #1485

#1205's final criterion asks for the census key to be computed on every full
run and asserted against a recorded ceiling, "so a later change that re-splits
a shared object family fails rather than quietly costing the time back". It has
been a reading taken once. This makes it a gate.

## The key is the whole of it

The same run answers three different numbers depending on what counts as the
same compile, and #1205's own census measured two of them 9x apart:

| Key | Repeats | What it answers |
|---|---|---|
| verbatim effective argv | 3 | was this exact command run twice |
| codegen flags + source set + `-D` | 61 | was this object built twice |

Verbatim understates because argv contains `-o <path>`, unique per gate, so two
identical objects with different output names read as different work. Both are
reported side by side on every run — reporting one alone is how the 9x went
unnoticed the first time.

## What is asserted is not what is reported

Asserted: the **count** of repeated compiles under the codegen key, a function
of the tree. Reported: the wall time they cost and its share of the run's own
suite wall.

The split was a hypothesis when I wrote it and is now measured. Two full runs
of the same tree:

| | repeated count | repeated wall | share of suite wall |
|---|---:|---:|---:|
| run 1 | 61 | 133.4 s | 12.9% of 1039.6 s |
| run 2 | 61 | 130.8 s | 12.4% of 1053.8 s |

The count is identical; the share moved. A ceiling on the share would have had
slack built into it on day one, and `task verify` has taken 3570 s and 1608 s
on this machine for identical input, so the slack would have had to be large.

## The ledger

25 families, 61 repeats. Each row carries the gate scripts that name its rarest
source, derived by `--regenerate` rather than typed; the derivation cannot
produce a blank reason — an unexplained family emits `NO OWNER FOUND`.

What remains is not the kind of redundancy #1205 removed. Every row is a gate
compiling its own driver alongside sources that are already shared objects. The
repetition is the driver's, and removing it means not running the gate twice
rather than not compiling the source twice — which is gate selection, and the
umbrella explicitly authorizes no further migration.

## Reproducibility

`verify-runner.sh` gains an opt-in `KOFUN_VERIFY_CENSUS_ARCHIVE`. #1205's final
measurement had to be reconstructed from an issue comment because the only copy
of its evidence was deleted at exit. The copy happens in the cleanup, so a run
that fails still leaves the census that explains it — which it did, twice,
while this was being built.

## Two things the first implementation got wrong

Both found by running it against a real census rather than by review:

- paths are absolute, so the key described one worktree. The ledger would have
  been unreadable by anyone else and would have changed when the checkout moved.
- a pure link has an empty source set, so every link sharing codegen flags
  collapsed into one family — a relationship between unrelated links, with
  linker time inside a number reuse cannot move.

## Verification

`VERIFY_JOBS=3 task verify` green in 1054 s on b302a8ea with a clean tree, the
census gate passing in situ. 25 negative self-tests. Five mutations run against
the real census, each refusing with its own message: ceiling raised, ceiling
lowered, a family row deleted, a stale family row added, a reason blanked.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -700,6 +700,11 @@ tasks:
     cmds:
       - "sh tests/tooling/verify-object-reuse/check.sh"
 
+  compile-census:
+    desc: "Assert the repeated-compile count of a full run against its ledger"
+    cmds:
+      - "sh tooling/compile-census/check.sh"
+
   fuzz-sanitizer-reuse:
     desc: "Verify runner-scoped fuzz sanitizer compiler object reuse"
     cmds:
--- a/artifacts/release-evidence/index.json
+++ b/artifacts/release-evidence/index.json
@@ -770,7 +770,7 @@
   ],
   "evidence_digests": {
     ".github/workflows/native-hosts.yml": "9b0155c4caf6aeb29e7d942bffc10acd12cd61b79c4d70af45ea2b10a9967daa",
-    "Taskfile.yml": "57a3197e63005197ee9ef5cb9daf56ba16c825892172181f990536a4d3cb2e1a",
+    "Taskfile.yml": "8c5365fd03cf922d3577eae6be348d009a9bbc4c067d463a0a51bd15ffe2d0af",
     "bootstrap/native/README.md": "6c470b203c3dbc327c6b601ab958c754e6f82be78c255d442307ad0c7691bc21",
     "bootstrap/native/check-macho64-signed.sh": "6fc51ddfec599ea85b8b06a5c16d68b6fdc994621cdb8ca4267e20be38141393",
     "bootstrap/native/check-macho64.sh": "d6bd52e3f6756aa607a2dc44a0117defeaaa55e64fa3f4221986da6406251e15",
--- a/bootstrap/stage2/verify-runner.sh
+++ b/bootstrap/stage2/verify-runner.sh
@@ -25,7 +25,26 @@ case $verify_run in
         ;;
 esac
 
+# Copy the completed census out before the run directory goes, when a caller
+# asks for it. #1205's final measurement had to be reconstructed from a comment
+# because the only copy of its evidence was deleted at exit; a measurement
+# whose input cannot be produced again is a reading, not a method (#1485).
+#
+# In the cleanup rather than after the last task, so a run that fails still
+# leaves the census that explains it.
+archive_verify_census() {
+    if test -n "${KOFUN_VERIFY_CENSUS_ARCHIVE:-}" &&
+       test -n "${KOFUN_VERIFY_CC_LOG:-}" &&
+       test -f "${KOFUN_VERIFY_CC_LOG:-}"
+    then
+        cp "$KOFUN_VERIFY_CC_LOG" "$KOFUN_VERIFY_CENSUS_ARCHIVE" 2>/dev/null ||
+            printf '%s\n' \
+                "verify runner: cannot archive the census to $KOFUN_VERIFY_CENSUS_ARCHIVE" >&2
+    fi
+}
+
 cleanup_verify_run() {
+    archive_verify_census
     if test -n "${verify_run:-}"; then
         kofun_stage2_owned_tree_remove "$verify_run" 2>/dev/null || true
     fi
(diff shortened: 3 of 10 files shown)


hjosugi:
CI caught something the local run could not, and it is now the most useful part
of this PR.

The first push asserted 61 repeats; CI measured 62 on the same tree. The count
is a function of the tree — but of the tree **plus an empty `build/`**.
`bin/kofun` builds `kofun-module-resolver` into `build/module-resolver` and
reuses it across runs, so a second local verify skips that compile and the
family it belongs to is one short. Removing `build/` reproduced CI exactly:
147 invocations, 62 repeats, 66 distinct families.

Asserting anyway would have told every second local run to lower a ceiling that
is correct — how a control teaches people to edit it rather than read it. The
condition is detectable exactly, because the compile is either in the census or
it is not, so the gate names it and skips its assertion rather than failing:

```
this census was taken with a warm `build/` — no compile of
`kofun-module-resolver` is in it, so `bin/kofun` reused the one from an
earlier run. 61 repeats measured against a ceiling of 62; the assertion is
skipped rather than failed. Remove `build/` to measure, as CI does on every run.
```

Run-scoped sources are now excluded as well: `mktemp` mutants and trees
unpacked into `build/.../.sufficient.<pid>/` carry a fresh path every run, so
each is its own family by construction and a ledger row naming one would be
stale the moment it was written. Ten such compiles are counted and reported
separately.

One thing about the fix itself is worth stating: adding the warm-cache rule
**disarmed six existing rule tests**. Their fixtures had no launcher-resolver
compile, so each one exercised the new skip instead of the rule it was written
for — and all six went on passing, because a skip returns no problems either.
Only the case asserting a specific message caught it. The rule fixtures now
carry the launcher row explicitly and say why.

Re-verified on a clean tree: `task verify` exit 0, gate asserting 62.

Issue #1321, opened 2026-08-11
Stage 2 C11: implement bounded mutation for Bytes[65536]


hjosugi:
## Metadata

- State: in-progress
- Parent: #1315
- Consumers: #1258 and #1312
- Kind: implementation
- Area: Stage 2 C11 / bounded Bytes mutation
- Priority: P1
- Size: M
- Blocked by: none
- Re-measured baseline: `origin/main@f42cfdd383b328535a82bd4b2e365e74c8ccb9f7`

## Goal

Add the complete bounded byte/range mutation surface to #1315's proven-unique
Managed Bytes carrier, without changing its type, layout, capacity, alias
profile, status tags, or cleanup semantics.

## Current behavior and evidence

Current `origin/main` at `f42cfdd383b328535a82bd4b2e365e74c8ccb9f7` contains #1315’s Managed `Bytes[65536]` carrier in both halves of the canonical pair:

```sh
git grep -c Bytes -- bootstrap/stage2/compiler.kofun bootstrap/stage2/compiler.c
# bootstrap/stage2/compiler.c:72
# bootstrap/stage2/compiler.kofun:59
```

#1315 owns the carrier and exact alias refusal. The accepted stdlib source owns `len`, equality, checked read, and checked edit semantics in `stdlib/linux_x86_64/bytes.kofun`.

## Exact mutation profile

- Implement `len`, `capacity`, checked byte read/write, `clear`, `reserve`,
  append-one-byte, `stage2_bytes_append_range(edit destination, read source,
  offset, count)`, and `stage2_bytes_append_self(edit value, offset, count)`.
- This child owns and emits the read carrier exactly once:
  `Stage2ByteRead = ByteValue(Int) | ByteReadNegativeOffset(Int) |
  ByteReadOutOfBounds(Int)` in declaration-order tags 0..2. Success carries the
  byte 0..255; failures carry the offending offset. There is no consumed tag,
  generic Result, or later redeclaration by the Text bridge.
- `append_range` requires distinct typed-HIR BindingIds for destination and
  source. Same or unknown identity is E2S177 before C emission. Self-append is
  only the dedicated single-edit operation: validate bounds against the old
  length, grow transactionally, then `memmove` from the post-growth buffer's
  original range. No simultaneous read/edit borrow is invented.
- Growth is exact: capacity 0 grows to 16, then doubles until the request fits,
  capped at 65,536. Overflow/capacity failure precedes allocation; OOM and every
[...]

PR #1507, opened 2026-08-16
typed-sidecar: record which paths stage2-events reads (#1504)

### LINKED ISSUE #1504: typed-sidecar: record which paths stage2-events reads, so a concurrent-run collision can be located
## Metadata

- State: in-progress
- Kind: test-quality
- Area: Quality / Toolchain
- Priority: P2
- Size: S
- Blocked by: none
- Audited against: `origin/main@075fbb241367c27863c74f3884989ba7ddbbfa5b`

## Goal

Have `tests/typed-sidecar/stage2-events.sh` record the paths it reads, so the
build path that two concurrent runs share can be found directly instead of
inferred from a census that comes up one short.

## Why

Two `task verify` runs at once corrupt this gate's companion census. The
recorded instance is

```
FAIL: expected all 329 repository error companions, saw 328
```

on a commit where `task stage2-events` passes standalone — and it happened
with the two runs **in different worktrees, with distinct
`KOFUN_GATE_WORK_NAMESPACE` values**. So the namespace separates most build
paths but not every one, and nobody has found which.

The question has now cost two sessions time twice. Both of us reasoned about
it from the outside and got it wrong in opposite directions:

- one session searched every gate script for fixed shared paths
  (`git grep -nE "(/var)?/tmp/[A-Za-z0-9_.-]"` over `tests/`, `bootstrap/`,
  `tooling/`), found only a worktree-local `mktemp -d`, and concluded no
  shared state exists. The search was sound; it answered a narrower question
  than the one asked of it. A search for *fixed* paths cannot see state held
  anywhere else.
- the other treated a concurrent run that exited non-zero as a data point,
  when the run had failed at an earlier task and had never reached this gate
  at all.

**A single concurrent run cannot settle it either way**, because the effect is
not deterministic — one observation of two overlapping `diagnostics` runs not
reproducing a recorded `E2S48 cannot open inventory` failure is evidence of
non-determinism, not of absence. Any experiment that schedules runs and reads
the outcome is a coin flip with a multi-hour downside when it lands wrong.

Instrumenting the census removes the guessing: the shared path either appears
in both runs' path logs or it does not.

## Proposed change

Under an opt-in environment variable so ordinary runs are unchanged, have the
gate append every path it opens for reading — the companion glob results, the
built binaries, and the work directory it derives — to a log. Then two
concurrent runs in separate worktrees can be diffed for any path that is not
under the run's own `WORK`.

## Acceptance criteria

- [ ] the gate records read paths when the variable is set, and is byte-for-byte
      unchanged in behaviour and output when it is not
- [ ] a self-test asserts the log is non-empty and contains the companion
      fixtures it counted, so an empty log cannot read as "no shared paths"
- [ ] the log distinguishes paths under the run's own `WORK` from paths
      outside it, since only the second kind can collide
- [ ] running the gate twice concurrently in two worktrees produces two logs
      that can be diffed by an ordinary `comm`, and the result is recorded on
      this issue whether or not it finds anything

The last criterion is the point: a run that finds no shared path is a result
worth recording, provided the self-test above shows the instrument could have
seen one.

## Validation

| Check | Command |
| --- | --- |
| The gate itself | `task stage2-events` |
| Repository | `task task-help assertions repository-check` |
| Full | `task verify` |

## Found by

#1315 and #1408, from opposite sides. Filed at the suggestion of the session
working #1408, who proposed instrumenting the census rather than the schedule.

hjosugi:
Instruments the census instead of the schedule, so the build path two concurrent runs share can be found directly rather than inferred from a count that comes up one short.

Opt-in via `KOFUN_STAGE2_EVENTS_PATH_LOG`; unset, nothing is written and the gate is unchanged.

## Three classes, not two

The issue asked for `WORK` versus outside-`WORK`. Two cannot answer the question: outside-`WORK` holds both paths private to this worktree and paths shared with every other one, and only the second kind can collide.

| class | meaning |
|---|---|
| `work` | under this run's own `WORK` — private to the run |
| `repo` | under `ROOT`, not `WORK` — private to the worktree |
| `outside` | neither — **the only kind that can collide** |

`work` and `repo` rows are recorded **relative** to their root. That is what makes two worktrees comparable at all: absolute paths differ by prefix, so a `comm` over them reports every line unique and finds nothing no matter what is shared. `outside` rows stay absolute, because there the literal path is the shared thing.

The census list is logged **unfiltered**, before the band filter, so a diff names *which* companion one run saw and the other did not — rather than confirming that a number differs, which the assertion already said.

## The instrument is not allowed to be silently blind

With logging on, the gate asserts its log is non-empty and holds at least as many `repo` paths as the census counted. An instrument that saw less than the gate did would report "no shared paths" exactly as confidently as a clean machine — the failure this issue is about, one level up.

`path-log-self-test.sh` exercises the recorder directly, not through a gate run, and is listed first in the task: a check that only runs inside a long gate is one nobody runs while changing what it checks. Twelve cases, including a must-not-fire one and a cross-root `comm` case proving the log can show a shared path.

Its comment records what mutation showed it does **not** catch — deleting the guard is caught by `set -u`, but a recorder inventing its own default path is not.

## Result

Recorded on the issue: **a null result across two experiments** — two concurrent gate runs in separate worktrees, and 13 probes during a concurrent full `verify`. Every run identical, `outside` empty. The comment states plainly what that does not license: reads inside the five compiled binaries are invisible to a shell-level recorder.

Closes #1504

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -222,9 +222,14 @@ tasks:
     cmds:
       - "sh bootstrap/stage2/check.sh"
 
+  # The path-log self-test runs FIRST and costs under a second: it exercises the
+  # #1504 recorder directly instead of through a `stage2-events` run that builds
+  # five binaries and drives four hundred companions. A check that can only run
+  # inside a ten-minute gate is one nobody runs while changing what it checks.
   stage2-events:
     desc: "Verify bounded complete/partial semantic events"
     cmds:
+      - "sh tests/typed-sidecar/path-log-self-test.sh"
       - "sh tests/typed-sidecar/stage2-events.sh"
       - "sh tests/typed-sidecar/stage2-event-stream.sh"
 
--- a/artifacts/release-evidence/index.json
+++ b/artifacts/release-evidence/index.json
@@ -785,7 +785,7 @@
   ],
   "evidence_digests": {
     ".github/workflows/native-hosts.yml": "9b0155c4caf6aeb29e7d942bffc10acd12cd61b79c4d70af45ea2b10a9967daa",
-    "Taskfile.yml": "a439a2edf4066f5b1a8a519865e26b75369212023497577dc32ca4f826caecc0",
+    "Taskfile.yml": "00638a3bebed08b931c6219b99e36be462c9a03eeb00b0020a4b5971bb428314",
     "bootstrap/native/README.md": "6c470b203c3dbc327c6b601ab958c754e6f82be78c255d442307ad0c7691bc21",
     "bootstrap/native/check-macho64-signed.sh": "6fc51ddfec599ea85b8b06a5c16d68b6fdc994621cdb8ca4267e20be38141393",
     "bootstrap/native/check-macho64.sh": "d6bd52e3f6756aa607a2dc44a0117defeaaa55e64fa3f4221986da6406251e15",
--- a/tests/typed-sidecar/path-log-self-test.sh
+++ b/tests/typed-sidecar/path-log-self-test.sh
@@ -1,0 +1,145 @@
+#!/bin/sh
+# Proof that the #1504 path log classifies, and that it is inert when unset.
+#
+# Exercises `tests/typed-sidecar/path-log.sh` DIRECTLY rather than through a
+# `stage2-events` run, which builds five binaries and drives 400+ companions.
+# That is not only about cost: a self-test that can only run inside a
+# ten-minute gate is one nobody runs while changing the thing it checks.
+#
+# The must-not-fire case is first and is the point. This instrument exists to
+# make a null result meaningful -- "no shared paths were found" is worth
+# something only if the log could have held one. An instrument that records
+# nothing, or that files every path under the same class, reports exactly the
+# same silence as a clean machine.
+set -eu
+
+LC_ALL=C
+export LC_ALL
+
+SELF=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
+ROOT=$(CDPATH= cd -- "$SELF/../.." && pwd)
+
+TMP=$(mktemp -d "${TMPDIR:-/tmp}/kofun-path-log-self-test.XXXXXX")
+trap 'rm -rf "$TMP"' EXIT HUP INT TERM
+
+WORK="$TMP/stage2-semantic-events"
+mkdir -p "$WORK/plain"
+
+pass=0
+fail=0
+check() {
+    if test "$2" = "$3"; then
+        printf 'ok   %-46s %s\n' "$1" "$3"
+        pass=$((pass + 1))
+    else
+        printf 'FAIL: %s: wanted [%s], got [%s]\n' "$1" "$3" "$2" >&2
+        fail=$((fail + 1))
+    fi
+}
+
+# shellcheck source=tests/typed-sidecar/path-log.sh
+. "$ROOT/tests/typed-sidecar/path-log.sh"
+
+# 1. MUST NOT FIRE: with the variable unset the recorder writes nothing at all.
+#
+#    Checked against a log that ALREADY EXISTS and has known content, not
+#    against a path the recorder was never given. The first version of this case
+#    asserted that an unrelated file had not appeared, which no defect could
+#    have falsified -- it would have passed against a recorder that wrote to the
+#    log on every call. A must-not-fire case has to be reachable by the failure
+#    it excludes.
+#
+#    WHAT THIS CASE COVERS, measured by mutating the recorder rather than
+#    assumed, because "the mutation misses" is the usual way a proof like this
+#    is worth nothing:
+#
+#      guard deleted from kofun_path_log_record   CAUGHT, by `set -u` --
+#        the unbound variable aborts before any assertion runs. That is the
+#        realistic defect and it is caught loudly, but it is caught by the
+#        shell, not by the check below.
+#      recorder defaults to a path of its own when unset   NOT CAUGHT.
+#        A mutation writing to /tmp/leaked.log left this sentinel untouched and
+#        every case here passed. Nothing cheap distinguishes "wrote nowhere"
+#        from "wrote somewhere I did not name", and pretending otherwise would
+#        be the kind of reassurance this file exists to refuse.
+SENTINEL="$TMP/sentinel.log"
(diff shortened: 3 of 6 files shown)


hjosugi:
## Integration review: the log is a manual subset, not every read path

CI is green and the limitation disclosure is candid, but this head does not yet satisfy #1504's acceptance criteria and should not merge before #1321.

1. **The instrument omits many paths the gate explicitly opens.** `stage2-events.sh` records the work root/fixture, common sources, grep-matched companion files, and corresponding sources. It explicitly excludes producer inputs, never records the compiled binaries it executes, and omits many fixtures, goldens, generated streams, and candidates opened by `find | grep`. The reported work count of one is evidence that individual work reads/binaries are not inventoried. This makes the result a manually selected subset; it cannot support “every path it opens for reading” or a null claim about the full gate. The disclosure that five compiled binaries' internal reads are invisible is correct, but explicit shell arguments are also missing.
2. **AC2's companion membership check can pass when companions are absent.** The check labels every repository row as a `logged_companion` and tests only that the total repo row count is at least the number of error cases. Removing the companion-logging loop still leaves hundreds of common/source rows, so the assertion stays green. Require each expected companion path individually. The self-test currently covers only synthetic classification/sort/`comm`, not real membership.
3. **“No behavior change when disabled” needs a precise surface.** Direct `stage2-events.sh` is inert without the environment variable, but `task stage2-events` always adds path-log self-test output/work. Clarify script-vs-task acceptance or keep the public task output byte-for-byte stable.
4. **Log-path safety/concurrency needs disposition.** `path-log.sh` truncates any caller-supplied path and uses predictable `<log>.sorted`; two writers to the same path can truncate/interleave/race. C-locale `sort -u` is deterministic for distinct logs, but same-target use is neither rejected nor safely committed. Also account for path canonicalization/TSV encoding if arbitrary paths remain supported.

Please either narrow the issue/PR claim to the explicitly instrumented shell subset and make every listed membership exact, or expand collection so the literal all-read claim is true. A future `strace`/`LD_PRELOAD` child can own compiled-binary internals, but it does not make missing explicit shell paths acceptable in this artifact.

Merge order also blocks this head: #1321 actively owns the Stage 2 lane and overlaps `Taskfile.yml`, `artifacts/release-evidence/index.json`, `tests/typed-sidecar/stage2-events.sh`, and the forbidden-requirements census. In particular, the combined Taskfile digest in the evidence index must be regenerated. Land #1321 first, then rebase/fix this PR and rerun its gates. This review made no edits and ran no canonical-pair gate.

hjosugi:
Reviewed from another session. **The three-class split and the relative-path decision are the two things that make this work**, and both are the kind of choice that is invisible once made and fatal if missed.

The `comm` argument in particular: absolute paths differ by prefix, so a diff over two worktrees' logs reports every line unique and finds nothing *no matter what is shared*. An instrument with that property would have produced a confident null result on every run forever. Recording `work` and `repo` relative to their roots while keeping `outside` absolute is exactly right — for `outside`, the literal path is the shared thing.

Three more things I checked and found correct:

- **The `case` arms are ordered most-specific first.** `"$WORK"/*` before `"$ROOT"/*` matters whenever `WORK` is inside `ROOT`, which is the default (`$ROOT/build/stage2-semantic-events`). Reversed, every work path would be logged as `repo` and the class that can collide would be diluted by the class that cannot.
- **The census list is logged unfiltered, before the band filter.** That is the difference between "a number differs" — which the assertion already said — and "*this* companion was seen by one run and not the other".
- **`>=` is the right relation** for "the log holds at least as many `repo` paths as the census counted". The gate reads more than the census list, so equality would be wrong, and the direction that matters is the instrument seeing *less* than the gate did.

And the self-test being listed **first** in the task, exercising the recorder directly rather than through a gate run, is the property I would have asked for. A check that only runs inside a long gate is one nobody runs while changing what it checks.

## One finding: the empty prefix

`kofun_path_log_record` classifies with

```sh
case $kofun_path_log_item in
    "$WORK"/*) ... ;;
    "$ROOT"/*) ... ;;
    *)         printf 'outside\t%s\n' ... ;;
esac
```

If `WORK` is set but **empty**, `"$WORK"/*` is `/*`, which matches **every absolute path** — so every path is logged as `work`, the `outside` class is empty, and a diff reports nothing shared. That is not a wrong answer that looks wrong; it is the exact answer this instrument exists to produce when the tree is clean, delivered on a broken recorder.

**It is latent, not live, and I want to be accurate about that.** `stage2-events.sh:9` is `WORK=${KOFUN_STAGE2_EVENTS_WORK:-"$ROOT/build/stage2-semantic-events"}`, and `:-` substitutes for empty as well as unset, so the shipped caller cannot reach it. What is reachable is the library's own contract — *"Set by the caller before sourcing: ROOT, WORK"* — which is a comment rather than a check, in a file whose whole purpose is to be sourced by callers that do not exist yet. `path-log-self-test.sh` already sets both itself.

Two lines in `kofun_path_log_init` close it:

```sh
test -n "${ROOT:-}" || { echo 'path-log: ROOT is empty' >&2; return 1; }
test -n "${WORK:-}" || { echo 'path-log: WORK is empty' >&2; return 1; }
```

This is adjacent to the mutation your self-test already admits it does not catch — *"recorder defaults to a path of its own when unset — NOT CAUGHT"*. That one is genuinely hard. This one is a `test -n`, and its failure mode is the same shape: a recorder that reports a clean tree because it stopped looking.

## On the null result

Recording **a null result across two experiments** and then stating what it does not license — reads inside the five compiled binaries are invisible to a shell-level recorder — is the part of this PR I would point other people at. A null result presented without its blind spot is how the two earlier sessions got it wrong in opposite directions, which your own header documents.

One thing that may be worth adding to that paragraph, from my side of the tree: **#1518 was the same class of bug and was found the other way.** `optional-tool-skips` invoked `bootstrap/c_abi/check.sh` nested while `verify` ran it directly, both defaulted to `build/c-abi`, and both began by recursively replacing it — surfacing as exit 126 in a gate that mentions neither. That is a shared *fixed* path between a gate and a gate it invokes, which your `repo` class would catch and which no `outside` row would ever show. If the null result holds, that shape is the next place to look, and this instrument is already pointed at it.

Not blocking. The finding is one guard in a file that is otherwise more careful than most.

Issue #1508, opened 2026-08-20
#1401 region 1: classify the pattern parser, and hold the Kofun half to calls that exist


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: implementation
- Area: Bootstrap / Stage 2 pair
- Priority: P1
- Size: M
- Parent: #1401
- Blocked by: none
- Audited against: `origin/main@f42cfdd3`

## Goal

Classify the first region of #1401's undefended-branch work-list — the pattern
parser — and land the gate the region's findings turned out to need.

#1401's second acceptance criterion says the 343 functions carrying an untaken
branch are classified **by a regional child, not all at once**. This is that
child for one region, and it reports what the region contains rather than only
that it was looked at.

## The region

The 17 functions of the pattern parser, as `tests/pair-coverage/undefended.tsv`
records them at `d07bfcf5`:

```sh
for f in pattern_recovery_end pattern_error pattern_limit_error \
         pattern_stop_token parse_pattern_atomic parse_pattern_or \
         pattern_match_open pattern_arm_arrow parse_pattern_trees \
         executable_constructor_pattern pattern_record_field \
         pattern_duplicate_binding_error enum_constructor_owner \
         enum_binding_catchall_name balanced_end pattern_first_error \
         validate_executable_patterns; do
    grep -P "\t$f$" tests/pair-coverage/undefended.tsv
done | awk -F'\t' '{g+=$1; c+=$2; n++} END {printf "%d functions, %d gcc, %d clang\n", n, g, c}'
# 17 functions, 125 gcc, 119 clang
```

That is 6.5% of the ledger's 1911 untaken branches under gcc, in 17 of its 354
functions. The region is contiguous in both halves — `compiler.c` 877–1988 and
`compiler.kofun` 416–1348 — which is why it is one reviewable change.

The pattern region is **unchanged** since the ledger was measured: every hunk
of `d07bfcf5..f42cfdd3` in `compiler.c` starts at line 7557 or later, so the
counts above still describe this tree.

## What the region contains

Both halves were read function by function, and the substrate they rest on was
checked rather than assumed first — `token_end`, `token_equal`/`token_text`,
`skip_trivia`, `token_kind`, `chars`, and the depth/node limits (32 / 256 on
both sides).

Fourteen of the seventeen are **behaviourally identical**, and the record says
what was checked for each rather than asserting equivalence: check ordering,
`<` against `<=`, saturating against returning decrements, which offset each
[...]

Issue #1408, opened 2026-08-13
Stage 2 pair: measure and ledger the undefended branches of compiler.c


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: implementation
- Area: Bootstrap / Stage 2 pair / quality
- Priority: P1
- Size: M
- Parent: #1401
- Blocked by: none
- Audited against: `origin/main@30f18e4d85dcd23232c46f73fb0d20fac301815e`

## Goal

Produce the undefended-branch work-list #1401 asks for, and make it a ledger
that cannot silently grow: a harness that measures which branches of
`bootstrap/stage2/compiler.c` the corpus never executes, a committed record of
that set, and a gate that fails in both directions.

This issue produces the work-list. It fixes no divergence and adds no fixture
to reach a branch — those are #1401's later regional children.

## Why the set is worth recording

#1401 establishes the formulation: a divergence between `compiler.kofun` and
its hand-maintained `compiler.c` transliteration can only survive on a path the
corpus never runs, because any divergence on a reached path is already caught
by the output-comparison gates (`selfhost-generations`, `selfhost-fixed-point`).
The undefended set is therefore the exact place a divergence can hide, and
today nothing writes it down.

## Current behavior and evidence

Measured on `origin/main@30f18e4d85dcd23232c46f73fb0d20fac301815e`. Nothing in
the repository measures this today:

```sh
grep -rn "coverage\|gcov" Taskfile.yml
# no coverage target exists

git ls-files 'tests/**' | grep -i coverage
# no coverage ledger exists
```

The set itself is large. Building `compiler.c` with
`cc -std=c11 -O0 --coverage` (gcc 16.1.1) and driving 999 `*.kofun` inputs from
`tests/`, `examples/`, `stdlib/`, and `spec/` through six CLI modes, restricted
to `compiler.c`'s own text:

| quantity | measured |
| --- | ---: |
| branches in `compiler.c` | 8732 |
| untaken | 2293 (26.3%) |
| functions with ≥1 untaken branch | 343 |
| functions never called | 9 |

Those figures come from a partial pass (~400 of 999 inputs), so the true
undefended set under the full corpus is **smaller**; they establish the order of
magnitude, which is what this issue's shape depends on. Producing the exact
number at a stamped commit is this issue's first deliverable.

## Two constraints the measurement must respect

Both were found by measuring, and either one silently ruins the ledger.

[...]

Issue #1483, opened 2026-08-15
Stage 2 cannot compile its own compiler: the first bound is 256 lexical uses per function


hjosugi:
## Metadata

- State: needs-detail
- Tracker role: curated implementation child
- Kind: implementation
- Area: Compiler / Bootstrap
- Priority: P1
- Size: M
- Parent: #1451
- Blocked by: none
- Audited against: `b1df6be8` — re-measured 2026-08-25. The original stamp
  was `origin/main@d42d2e4c98d66ece506802fbaaa83419f604b8bd`; what changed since
  is recorded in the comments, not silently overwritten here.

## Goal

Make `bootstrap/stage2/compiler.kofun` compilable by the Stage 2 compiler, so
`compiler.c` can stop being a hand-maintained transliteration.

## Current behavior and evidence

`#32 Stage 2 fixed point: Kofun compiles Kofun` is **closed** — Kofun compiles
Kofun. What it does not compile is *this* Kofun compiler. Measured on
`origin/main@d42d2e4c98d66ece506802fbaaa83419f604b8bd`, building `bootstrap/stage2/compiler.c` with
`cc -std=c11 -O2` and running it on its own Kofun source:

```
$ kofun-stage2 --compile-outcome bootstrap/stage2/compiler.kofun out.c out.ir out.tokens
error[E2S35]: lexical use limit is 256 per function at byte 30874
```

Byte 30874 is line 906, inside `parse_pattern_atomic_result`.

> **Corrected 2026-08-25 at `b1df6be8`.** Two claims below did not survive
> measurement. Raising this bound *is* free of array cost — `use_count` is a
> plain `int64_t` counter writing into a growable `Buffer`, and no array is
> sized by 256. And `E2S35` is not the only blocker: behind it are an unbound
> `lambda_scope` (fixed in PR #1621) and `E2S16 unknown Core function
> is_xid_start`, which is why this issue is now `blocked` on #1513. The
> complete measured set of violated bounds is in the comments.

**The blocker is not a missing feature. It is a capacity bound the compiler
imposes on itself.** `compiler.kofun` is 21,363 lines and has at least one
function that exceeds the compiler's own per-function limit. The same class of
bound appears 11 times at 256-per-function, plus 512, 64, 32, 16 and 8
elsewhere.

`E2S35` is the first refusal, not necessarily the only one. Nobody has measured
what lies behind it, because nothing gets that far.

## Why it matters

[...]

PR #1509, opened 2026-08-20
pair-coverage: the Kofun half calls two functions that do not exist (#1508)

hjosugi:
Closes part of #1508 (child of #1401).

## The finding

**`bootstrap/stage2/compiler.kofun` calls two functions that do not exist, and
nothing in this repository has ever resolved a name in that file.**

```
bootstrap/stage2/compiler.kofun:15:    return symbol == "_" || is_xid_start(symbol)
bootstrap/stage2/compiler.kofun:18143:    if ends_with(output_path, ".c") {
```

Neither name is defined there; neither is in `builtin_arity` in `compiler.c`;
`is_xid_start` appears nowhere else in the tree as a Kofun-level name.
`ends_with` is a `static` helper the C half wrote for itself at
`compiler.c:27261`.

Why nothing could see it: `bootstrap/manifest.json` names stage 1's Kofun source
as `canonical_source` and stage 2's C as `trusted_seed`. The stage 2 Kofun half
is **neither**, and `bootstrap/stage2/check.sh` round-trips its text,
byte-compares the identity projection, and asserts the IR is non-empty with a
version header — no step resolves a name.
`tests/pair-coverage/check.sh` already states that hole in prose; this closes one
property of it.

## What this adds

A gate — `task stage2-pair-calls` — asserting that every call in
`bootstrap/stage2/compiler.kofun` names something that exists, plus
`tests/pair-coverage/unresolved-calls.tsv`, the ledger of the two places where
that is false today.

## Why it is not a gate that cannot fail

#1401's fourth acceptance criterion forbids adding a gate that asserts something
it cannot refuse. `sh tests/pair-coverage/calls.sh --prove DIR` runs six cases
and requires each to behave for the right reason, by name:

| case | required |
| --- | --- |
| a call that resolves to nothing | refused, naming the call |
| a ledger row for a name that resolves | refused, naming the row |
| a call named inside a string literal | **accepted** |
| a call named inside a comment | **accepted** |
| a line that ends inside a string literal | refused, naming the assumption |
| the C half's builtin table renamed | refused, rather than resolving nothing |

The two "accepted" cases are the ones that make this gate different from the
obvious one: `compiler.kofun` carries the whole emitted C runtime as string
literals, so a scanner that does not know where a string ends reports 90
findings, none of them real. The scanner was also cross-checked against an
independent implementation — both agree on all 527 plain call targets exactly.

## Where the resolution set comes from

`builtin_arity`, `int_bit_method_arity` and `keyword_token` are read out of
`compiler.c` at run time and never restated here. A builtin added there stops
being reported the moment it is added, and a table that is renamed makes this
gate FAIL rather than quietly resolve nothing — case 6 above.

## The ledger is not an exemption list

`tests/pair-coverage/unresolved-calls.tsv` fails in both directions, like
`tests/assertions/budget.tsv`. Both rows are defects, not tolerated cases.
Neither is fixed here because both edit the canonical Stage 2 pair, which #1321
holds; that session and this one agreed the fixes belong on the branch that
carries the gate, after their claim merges.

`is_xid_start` also needs a decision, not a transliteration: `chars()` is
byte-wise (`kofun_rt_chars` emits one 1-byte string per byte), so the Kofun half
inspects a single byte where the C half decodes a UTF-8 scalar. The C half
accepts `let 日本 = 41` today — it lexes as one identifier token spanning bytes
27..33 — and a byte-wise `is_xid_start` would resolve the name without agreeing
with that.

## Validation

| Check | Result |
| --- | --- |
| `task stage2-pair-calls` | PASS — `533 of 535 call targets in compiler.kofun resolve`, 6 of 6 proof cases |
| `task task-help` | PASS — the new task is described and grouped |
| `node tooling/gate-reachability/check.mjs` | PASS — 164 of 181 tasks reachable |
| `node tooling/forbidden-requirements/check.mjs` | PASS — the new `shell-build-driver` row is recorded |
| the driver comparison at the head of `tests/pair-coverage/measure.sh` | no undriven gate, no stale pin |

The gate is cheap — two files of text, no build — so it goes into `verify`
rather than into `tooling/gate-reachability/unreachable.tsv` beside its
`stage2-pair-coverage` neighbour.

## Provenance

Found while classifying the pattern-parser region of #1408's undefended-branch
ledger, which is #1401's second acceptance criterion. The region's
classification is recorded on #1508.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -1169,6 +1169,20 @@ tasks:
       - "sh tests/pair-coverage/check.sh"
       - "sh tests/pair-coverage/prove.sh build/{{.NAMESPACE}}pair-coverage-proof"
 
+  # Every call in the Stage 2 pair's Kofun half names something that exists.
+  # Cheap -- two files of text, no build -- so unlike its `stage2-pair-coverage`
+  # neighbour it belongs in `verify` rather than in
+  # tooling/gate-reachability/unreachable.tsv. It also demonstrates that it can
+  # refuse on every run: the ledger it holds records two live defects, so
+  # "green" here would otherwise say nothing.
+  stage2-pair-calls:
+    desc: "Resolve every call in bootstrap/stage2/compiler.kofun, or record why not"
+    vars:
+      NAMESPACE: '{{if .KOFUN_GATE_WORK_NAMESPACE}}{{.KOFUN_GATE_WORK_NAMESPACE}}/{{end}}'
+    cmds:
+      - "sh tests/pair-coverage/calls.sh"
+      - "sh tests/pair-coverage/calls.sh --prove build/{{.NAMESPACE}}pair-calls-proof"
+
   # An issue's state is carried twice, as a label and as the `State:` line in
   # its body, and nothing held the two together until this. Reads the committed
   # snapshot only, so it stays hermetic and belongs in `verify`.
@@ -1256,6 +1270,7 @@ tasks:
             selfhost-native \
             stage2 \
             stage2-events \
+            stage2-pair-calls \
             patterns \
             adt \
             records \
--- a/artifacts/release-evidence/index.json
+++ b/artifacts/release-evidence/index.json
@@ -770,7 +770,7 @@
   ],
   "evidence_digests": {
     ".github/workflows/native-hosts.yml": "9b0155c4caf6aeb29e7d942bffc10acd12cd61b79c4d70af45ea2b10a9967daa",
-    "Taskfile.yml": "b1fc4f3d7c7cf4137e656813dff5d61304a3ea949d583094d4c1ea35335c561b",
+    "Taskfile.yml": "890b3ad284480df7c59729a4fa74ca5d111e87a1adb159e8831d4faa7f38d676",
     "bootstrap/native/README.md": "6c470b203c3dbc327c6b601ab958c754e6f82be78c255d442307ad0c7691bc21",
     "bootstrap/native/check-macho64-signed.sh": "6fc51ddfec599ea85b8b06a5c16d68b6fdc994621cdb8ca4267e20be38141393",
     "bootstrap/native/check-macho64.sh": "d6bd52e3f6756aa607a2dc44a0117defeaaa55e64fa3f4221986da6406251e15",
--- a/tests/pair-coverage/calls.sh
+++ b/tests/pair-coverage/calls.sh
@@ -1,0 +1,425 @@
+#!/bin/sh
+# Does every call in `bootstrap/stage2/compiler.kofun` name something that
+# exists? (#1401, sibling of #1408's undefended-branch ledger)
+#
+# WHY THIS GATE EXISTS, AND WHY NOTHING ELSE COVERS IT. `compiler.kofun` and
+# `compiler.c` are the same program written twice. The gates that compare them
+# -- `selfhost-generations`, `selfhost-fixed-point` -- compare EMITTED C, and
+# only the C half emits: `bootstrap/manifest.json` names
+# `bootstrap/stage1/compiler.kofun` as the canonical source and
+# `bootstrap/stage2/compiler.c` as the trusted seed. The stage 2 Kofun half is
+# neither. `bootstrap/stage2/check.sh` round-trips it through `kofun-stage2`,
+# byte-compares the identity projection, and asserts the IR is non-empty and
+# carries a version header -- all of which a file with a call to a function
+# that does not exist passes, because none of it resolves a name.
+#
+# `tests/pair-coverage/check.sh` states that hole in prose ("THE KOFUN HALF IS
+# NEVER SEMANTICALLY COMPILED... A logic error that preserved all of those would
+# be invisible to every gate in this repository"). This is the first gate that
+# closes a piece of it: not the whole of semantic checking, one property --
+# every called name resolves.
+#
+# WHAT IT FOUND ON THE DAY IT LANDED, which is why it is not a gate asserting a
+# property it cannot fail on:
+#
+#   is_xid_start   compiler.kofun:15, in the file's own lexer. Nothing in the
+#                  tree defines it -- not a `fn` here, not a builtin in the C
+#                  half's tables, not the spec. The C half does not go through
+#                  a builtin at all: `identifier_start_at` decodes a code point
+#                  and tests `kofun_unicode_is_xid_start`.
+#   ends_with      compiler.kofun:18143. The C half defines a static helper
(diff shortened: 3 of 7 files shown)


hjosugi:
Release-integration coordination:

This PR is currently non-draft at `c1cbeab773341a0ed7c94b4566ae0d6e4b07d192`, mergeable/CLEAN, with Kofun verification, release evidence, and native-host evidence green. The audit also confirms it does not touch the canonical Stage 2 pair.

Issue #1508 still records one unfinished acceptance item that depends on #1321 releasing the pair, and its live claim is `pr-open`. Please confirm whether this green, non-closing partial PR should merge now. If so, please also transition the #1508 claim back to `active` after merge (or release it if the remaining work is paused); the issue itself must remain open.

I will not merge or change the claim without that owner disposition.

hjosugi:
## Integration review: gate still has false-negative classes

The wiring/collateral is internally consistent and CI is green, but this head is not mergeable yet. Independent read-only review found six concrete blockers in `tests/pair-coverage/calls.sh`:

1. **Plain calls and Int methods share one name set.** The extractor records `plain|method` and then drops the kind with `cut -f2`; definitions, builtins, bit methods, and keywords are unioned. An undefined plain `and(1)` is therefore treated as resolved only because `and` exists as an Int method. The canonical C half says `and` is an ordinary identifier meaningful only in method position. Pin both directions (`and(1)` and a method name shadowed by a plain builtin) and keep namespaces distinct.
2. **Trivia-separated calls disappear.** The scanner skips only ASCII spaces and processes one line at a time, whereas the canonical validator calls `skip_trivia` (whitespace plus `#` comments) before recognizing `(`. `prove_absent_zz\n  (1)`, tab separation, and `foo # comment\n(1)` are real calls that currently bypass the ledger. Add proof cases and align extraction with the language.
3. **Valid Unicode and uppercase callees are excluded.** The `[a-z_]` extractor cannot see functions such as the tracked `fn 合計` or `fn Maybe`. Stage 2 accepts XID identifiers, so this is a false-negative class in the exact area #1508 is meant to defend.
4. **The reach denominator is not call targets.** The reported 535 includes roughly the 507 `fn` declarations plus control keywords because declaration position is not excluded. An over-approximation can be useful, but it cannot support the current “all call targets exactly”/reach claim. Separate declarations/keywords or report a truthful denominator.
5. **The ledger schema is not enforced.** Only field 1 is read, so a missing/unknown kind, empty reason, and duplicate rows all pass. Validate the three-field schema, kind vocabulary, non-empty reasoning, and uniqueness, with a malformed-ledger proof.
6. **`--prove DIR` can recursively delete an arbitrary directory.** Lines 100–105 unconditionally `rm -rf "$PROVE"`; documented direct usage can be pointed at `.` or an existing data directory. Restrict it to a validated child under the repository build directory, or create a private temporary child and remove only that.

This review did not edit the branch or run canonical-pair gates. `Taskfile.yml`, drivers/census/help, and the release-evidence digest otherwise agree with head `c1cbeab7`. The canonical pair remains untouched, so #1321 serialization is respected. Because #1508 still includes the two pair call fixes, please also state whether this PR is a partial artifact with its own issue/claim disposition or remains open until the pair owner releases.

hjosugi:
Reviewed from another session. **The two findings are real and they are bigger than this PR treats them**, so most of this comment is about what they connect to rather than about the diff, which I think is right.

`is_xid_start` and `ends_with` are called in `compiler.kofun` and defined nowhere as Kofun-level names. That is the same class as **#1515**, which I filed an hour ago from the other direction: Stage 2 counts a trailing comma as an extra parameter *and* as an extra argument, so `fn f(\n a: Int,\n b: Int,\n)` refuses with `expects 3 arguments, got 2` — and **`compiler.kofun` contains zero trailing commas** while 38 other tracked `.kofun` files use the form `spec/syntax/EXPRESSIONS_AND_LITERALS.md:17` permits.

Put together, the shape is: **the one file whose compilation is gated end to end is the one file that never exercises the constructs**, and the constructs it *does* use are not resolved by anything. Your gate closes the first half of that. #1483 is where both land.

### What I checked rather than assumed

- **The ledger fails in both directions and cannot become a suppression list.** An unlisted unresolved call fails, and a listed one that now resolves fails. Adding a row is therefore an explicit reviewed edit, which is the property `budget.tsv` and `unreachable.tsv` already have. No count ceiling is needed here — unlike `census.tsv`, where `unknown` is a *value* a rule can produce, a row here can only be produced by a person.
- **The resolution set is derived, not restated.** Reading `builtin_arity`, `int_bit_method_arity` and `keyword_token` out of `compiler.c` at run time is the part that makes this survive a builtin being added. And making a renamed table **fail** rather than resolve nothing is the difference between this and a gate that quietly stops looking — that is the sixth prove case, and it is the one I would have asked for.
- **The two accepted cases carry the weight.** A gate that only proves it can refuse is a gate nobody can trust to be silent. String literal and comment being *accepted* is what says the scanner reads Kofun rather than text.
- **"What it does not cover" is stated in the file**, not in a PR description that stops existing after the merge. Arity, constructors, uncalled `fn`s, and the C half — each with the reason. That is the paragraph that stops this being read as a total check later.

### One suggestion, from a mistake that cost me two CI runs today

`--prove` does `rm -rf "$PROVE"; mkdir -p "$PROVE"` on a path the Taskfile fixes as `build/{{.NAMESPACE}}pair-calls-proof`. That is exactly the shape of **#1518**, which PR #1521 is fixing right now: `optional-tool-skips` invoked `bootstrap/c_abi/check.sh` nested while `task verify` ran it directly, both defaulted to `build/c-abi`, and both began by recursively replacing it. The nested run exited **126** — on my PR #1512, twice, deterministically enough that I re-ran before looking.

`stage2-pair-calls` runs once in `verify` today, so there is no live collision. But the main check path already uses `mktemp -d`, and using it for the proof directory too would make the property structural rather than schedule-dependent. `${2:?usage}` already guards the empty-argument case, which is the more dangerous half — that part is right.

### One thing I would add to the ledger rows, not the code

Each row records *that* the name does not resolve. `ends_with` also records *why it was invisible*: it is a `static` helper the C half wrote for itself at `compiler.c:27261`, so the two halves disagree about what a name means rather than about whether it exists. That distinction is what makes it a pair-divergence finding rather than a typo, and it is the thing a future reader of `unresolved-calls.tsv` will want. If it is already in the row's reasoning column, ignore this.

Not blocking on any of the above. Separately: my #1520 and #1522 each add two `required-today` rows to `tooling/forbidden-requirements/census.tsv` and both re-pin `artifacts/release-evidence/index.json`. Your rows are elsewhere in both files; I will rebase behind this PR rather than the other way round.

Issue #1511, opened 2026-08-20
Standard library: name the secure-randomness boundary over getrandom(2)


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: implementation
- Area: Standard library / secure-randomness
- Priority: P2
- Size: S
- Parent: #1475
- Blocked by: none
- Audited against: `origin/main@f42cfdd383b328535a82bd4b2e365e74c8ccb9f7`

## Goal

Give `secure-randomness` a named boundary: one adapter that fills a caller's
`Bytes` from `getrandom(2)` completely or not at all, so the documentation can
stop sending security-sensitive callers to a bare syscall wrapper.

## Current behavior and evidence

Measured on `origin/main@f42cfdd383b328535a82bd4b2e365e74c8ccb9f7`.

**The syscall binding exists and carries no discipline.**
`stdlib/linux_x86_64/process.kofun:3`:

```kofun
fn random_fill(edit destination: Bytes, flags: Int) -> SysResult[Int] {
    return syscall_result("getrandom", raw_getrandom(...))
}
```

One `getrandom(2)` call. No `EINTR` retry, no short-read loop, no statement
about a partially filled buffer — the caller receives `Ok(n)` with `n` possibly
less than `len(destination)` and nothing tells them that is not success.

**The discipline exists exactly once, and not for this purpose.**
`stdlib/random/linux_x86_64.kofun:38` wraps that call in the retry loop, for
exactly eight bytes, to seed a *non*-cryptographic Park-Miller generator.
`stdlib/random/README.md:59-64` is explicit that this "does **not** turn
Park-Miller into a cryptographically secure generator", and closes:

> Security-sensitive callers must use the platform entropy API directly.

Today "directly" means the undisciplined wrapper above. That sentence is the
gap: the capability matrix's own note for the row says the same thing in the
other direction — "the Linux entropy adapter exists; the portable capability
boundary does not" (`stdlib/capabilities.tsv:55`).

**A gate constrains where the new code may live.**
`stdlib/random/tests/verify.sh:52` asserts that exactly one file under
`stdlib/random/` calls `random_fill(seed_bytes, 0)`, failing with "system
entropy must remain in one explicit adapter". A second entropy caller added
*inside* `stdlib/random/` would break that gate; one added outside it does not,
and must not weaken it.

## Scope

- A new adapter module, `stdlib/entropy/linux_x86_64.kofun`, outside
[...]

Issue #1510, opened 2026-08-20
Standard library: a bounded buffered line reader over the Linux file adapter


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: implementation
- Area: Standard library / buffered-io
- Priority: P2
- Size: M
- Parent: #1475
- Blocked by: none
- Audited against: `origin/main@f42cfdd383b328535a82bd4b2e365e74c8ccb9f7`

## Goal

Deliver the `buffered-io` capability as one bounded artifact: a reader that
turns an open `File` into lines without knowing their length in advance and
without loading the file as `Text`, with a gate that holds it.

## Current behavior and evidence

Measured on `origin/main@f42cfdd383b328535a82bd4b2e365e74c8ccb9f7`.

**Nothing buffered exists.** The only matches are the capability matrix and its
own checker:

```sh
git grep -lni buffered -- stdlib
# stdlib/capabilities.tsv
# stdlib/check-capabilities.sh
```

**Every existing read needs the length in advance.** `stdlib/linux_x86_64/io.kofun`
offers `file_read` (`:45`), which takes a caller-sized `edit destination: Bytes`
plus an explicit `length`, and `file_read_exact` (`:103`), which already loops
`EINTR` and short reads correctly — but only up to `len(destination)`, so the
caller must know how many bytes it wants before it asks. A line's length is not
known before it is read.

**No existing parser can be fed incrementally.** The nearest record reader takes
the whole document as `Text`:

```sh
grep -n 'fn csv_parse' stdlib/csv/csv.kofun
# 61:fn csv_parse(source: Text, read dialect: CsvDialect) -> Result[CsvDocument, CsvError]
```

**The consumer is named and counted.** `package/manager.sh:312` and `:353` are
both `while IFS='\t' read -r ...` loops, and
`docs/PACKAGE_MANAGER_IN_KOFUN.md:99` records them as the `buffered-io`
requirement of the package-manager rewrite (#1457):

```
| line-wise input (2 `read -r` sites) | `buffered-io` | planned | no |
```

## Scope

The bounded surface, split so the matrix's `portable` tier stays honest — the
charter (`docs/STANDARD_LIBRARY_CHARTER.md:41`) defines the portable tier as
including "the portable half of any capability whose other half is a platform
adapter":

- **Portable half** (`stdlib/buffered_io/buffered_io.kofun`): a fixed-capacity
[...]

Issue #1480, opened 2026-08-15
linux_x86_64 filesystem: implement authorized create, remove, copy, and rename


hjosugi:
## Metadata

- State: blocked
- Tracker role: curated implementation child
- Kind: implementation
- Area: Standard Library / Interop
- Priority: P1
- Size: M
- Parent: #1218
- Blocked by: #1193
- Audited against: `origin/main@60ee6b1d7ae878ace0f222ef22490b7cdd2cc8d4`

## Goal

Implement authority-scoped directory creation, removal, copy, and rename on
`linux_x86_64`, under the same accepted lookup and symlink rules #1235 applies
to enumeration.

## Why this is missing rather than merely unimplemented

`stdlib/capabilities.tsv` has 38 rows and **none covers a filesystem mutation
beyond writing an already-open file.** The two rows that sound as though they
might exclude it in their own notes:

| row | state | note |
| --- | --- | --- |
| `syscall-file-round-trip` | implemented | "open/write/lseek/read/close round-trip with errno evidence; **no directory listing**, process spawn, or environment authority" |
| `directory-listing` | planned | "authorized deterministic Linux x86-64 **listing**; file open and stat are not directory-enumeration evidence" |

The implementation agrees. `stdlib/linux_x86_64/` exposes 68 functions over
this syscall surface, measured on `origin/main@60ee6b1d7ae878ace0f222ef22490b7cdd2cc8d4`:

```
raw_accept4 raw_bind raw_clock_gettime raw_close raw_connect
raw_epoll_create1 raw_epoll_ctl raw_epoll_wait raw_exit raw_exit_group
raw_getrandom raw_listen raw_lseek raw_mmap raw_munmap raw_nanosleep
raw_open raw_read raw_setsockopt raw_socket raw_stat raw_write
```

No `mkdir`, `rmdir`, `unlink`, `rename`, or `renameat2`.

**A `planned` row is a decision someone made; an absent row is a question
nobody asked, and absence does not appear in a matrix.** #1462 adds the four
rows; this issue is what they cite, because after #1454 a `planned` row must
name work that is still open.

## What needs it

`package/manager.sh` performs these operations **15 times** — 6 `rm -rf`,
5 `cp`, 2 `mkdir`, 2 `mv` — and RFC-0018 requires the package manager to be
implemented in Kofun without a shell. #1453 records the full prerequisite list;
this is the largest single gap in it.

## Scope

[...]

PR #1512, opened 2026-08-20
capabilities: three rows that cited a planning issue now cite work (#1475)

### LINKED ISSUE #1475: Standard library: buffered-io, secure-randomness, and compression-archive are planned against nothing
## Metadata

- State: needs-detail
- Tracker role: curated planning child
- Kind: planning
- Area: Standard library
- Priority: P2
- Size: M
- Parent: #1454
- Blocked by: none
- Audited against: `origin/main@a98cb13eabc7b5c48a948d381928a460670e3d67`

## Goal

Give `buffered-io`, `secure-randomness`, and `compression-archive` a live home,
so their rows in `stdlib/capabilities.tsv` cite work that exists.

## Why this issue exists

#1454 found that nine of the twelve `planned`/`deferred` rows in the capability
matrix cited **closed** issues. Six of the nine have an exact live successor and
were re-pointed there:

| row | was | now |
| --- | --- | --- |
| `concurrency` | #555 #736 | #1162 #1163 |
| `url` | #638 | #1258 — it names URL input directly |
| `temporary-files` | #231 | #1463 |
| `hashes-checksums` | #636 | #1455 |
| `mime` | #638 | #644 |
| `crypto-tls` | #638 | #644 |

**Three had no live successor at all**, and inventing one issue each would be
three speculative trackers. They are one group with one status, so they get one:

- `buffered-io` — planned against closed #231. A prerequisite for the package
  manager rewrite, which reads line-wise input at two sites.
- `secure-randomness` — planned against closed #234. Tier `adapter`.
- `compression-archive` — deferred against closed #636. Tier `module`.

## What this issue needs before it is `ready`

`needs-detail` rather than `ready`, deliberately: the matrix rows say what the
capabilities *are* and nothing says what would count as delivering them. Each
needs, at minimum:

- the bounded surface — `buffered-io` over what, with what buffer discipline,
  and whether it composes with `syscall-file-round-trip` as it stands;
- whether `secure-randomness` is a distinct adapter or a use of an existing
  syscall binding, given `raw_getrandom` is already exposed in
  `stdlib/linux_x86_64/`;
- whether `compression-archive` is one capability or two, and which formats are
  in scope at all.

None of those is answerable from the matrix, which is why the rows have been
`planned` against nothing for long enough that their issues closed.

## Non-goals

- Implementing any of the three.
- Re-litigating the tier assignments.
- A general standard-library roadmap. The other six rows are re-pointed and out
  of scope here.

hjosugi:
Closes #1475.

#1475 was `needs-detail`: three capability rows pointed at it, and it asked what
each capability would have to deliver before it could be `ready`. The answer is
that the three are not one group — measured on
`origin/main@f42cfdd383b328535a82bd4b2e365e74c8ccb9f7`, two have named consumers
and one has none at all.

So the refinement is a split, which `docs/ISSUE_READINESS.md` names as the
refinement work itself when the artifact cannot be one reviewable change:

| row | was | now | why |
| --- | --- | --- | --- |
| `buffered-io` | `#1475` | **#1510** | a counted consumer, and nothing existing can serve it |
| `secure-randomness` | `#1475` | **#1511** | the binding exists; the discipline is one sentence wide and lives elsewhere |
| `compression-archive` | `#1475` | **#644** | no consumer in this repository; its only *named* consumer is an accepted contract |

The diff is three rows. The work is the measurement behind them, which is in the
commit message and in the two new issues.

### What the split rests on

**`buffered-io` (#1510)** — `package/manager.sh:312` and `:353` are both
`while IFS='\t' read -r` loops, already recorded at
`docs/PACKAGE_MANAGER_IN_KOFUN.md:99` as this capability's requirement of #1457.
Nothing buffered exists (`git grep -lni buffered -- stdlib` matches only the
matrix and its own checker), and neither existing reader can be adapted:
`file_read_exact` (`stdlib/linux_x86_64/io.kofun:103`) already loops `EINTR` and
short reads correctly but needs the length in advance, and `csv_parse`
(`stdlib/csv/csv.kofun:61`) takes the whole document as `Text`. A line's length
is not known before it is read. #1510 specifies the reader, splits it portable
half / adapter half so the row's `portable` tier stays honest, and names the
compiler boundary it will hit rather than leaving it to be discovered.

**`secure-randomness` (#1511)** — `stdlib/linux_x86_64/process.kofun:3` binds
`getrandom(2)` with no `EINTR` retry, no short-read loop, and no statement about
a partially filled buffer. The discipline exists exactly once, at
`stdlib/random/linux_x86_64.kofun:38`, for eight bytes, to seed a generator whose
own README says it is not cryptographic — and that README (`:64`) closes with
"Security-sensitive callers must use the platform entropy API directly", which
today means the undisciplined binding. #1511 is complete-or-error over
`getrandom(2)`, outside `stdlib/random/` so that
`stdlib/random/tests/verify.sh:52`'s "system entropy must remain in one explicit
adapter" assertion keeps holding unchanged.

**`compression-archive` (#644)** — searching every tracked file outside
`examples/`, `vendor/`, `unicode/`, and lockfiles finds no `tar`, `gzip`, `zstd`,
`xz`, `deflate`, or `inflate` anywhere in the toolchain. The only hits are
`.github/workflows/release.yml:99-101`, `git archive | gzip -n` for the source
tarball, which involves neither Kofun nor this capability. What it does have is
one *named* consumer in an accepted contract: `docs/stdlib/http-client.md:105`,
"Automatic decompression covers `gzip` only in v1 and may be disabled", bounded
there by output bytes and expansion ratio. `mime` and `crypto-tls` already cite
#644 for exactly that reason.

That is also the answer to #1475's third question — "whether
`compression-archive` is one capability or two, and which formats are in scope
at all". In v1 it is gzip **decompression**, owned by the HTTP client contract.
No archive format has a consumer, and `tar` is in scope for nobody. The note now
says so instead of "belongs to the module tier when one appears".

### Validation

```
sh stdlib/check-capabilities.sh
PASS: 14 planned/deferred rows cite an open issue, decided against snapshot horizon #1480
PASS: every capability the charter names has a row
PASS: the capability matrix states are exact, and 12 mutations are refused
```

One thing to know rather than discover: #1510 and #1511 are above the committed
snapshot's horizon (#1480), so `check-capabilities.sh` records them as
undecidable rather than open. That is the checker working as designed — it fails
only when *every* planned/deferred citation is newer than the snapshot, and
fourteen rows still decide. The scheduled live-backlog lane is what will decide
these two, and both are open.

No other file references the three rows: `git grep -n 1475` outside `artifacts/`
finds only unrelated offsets in `bootstrap/selfhost/frontend/S.hir`.

### Lane

Built in a separate worktree on `origin/main@f42cfdd3`. Touches one file,
`stdlib/capabilities.tsv`, which no other live claim names — not the canonical
Stage 2 pair (#1321), not `tests/pair-coverage/**` (#1408/#1508), not
`tooling/typed-sidecar` (#1504). No `Taskfile.yml` edit, so
`artifacts/release-evidence/index.json` is untouched and the release-evidence
lane stays green.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/stdlib/capabilities.tsv
+++ b/stdlib/capabilities.tsv
@@ -50,12 +50,12 @@ affine-resource-handle	portable	implemented	task affine-resource-handle	one boun
 benchmark-harness	portable	specified	docs/stdlib/benchmark.md	accepted contract (#640); raw-sample report schema slice is #646
 concurrency	portable	planned	#1162 #1163	scoped parallelism only, after the runtime contract exists
 stream-protocol	portable	specified	docs/stdlib/stream-protocol.md	accepted protocol contract (#627); no operators, scheduler, or adapters implemented
-buffered-io	portable	planned	#1475	no buffered reader/writer exists over the syscall seed today
+buffered-io	portable	planned	#1510	reader only; no writer half until a consumer names one
 url	portable	planned	#1258	parsing and normalisation; owned by the HTTP client contract, not independent
-secure-randomness	adapter	planned	#1475	the Linux entropy adapter exists; the portable capability boundary does not
+secure-randomness	adapter	planned	#1511	the getrandom(2) binding exists without complete-or-error discipline; the boundary that states it does not
 temporary-files	adapter	planned	#1463	creation, cleanup on take, and collision behaviour are unspecified
 hashes-checksums	portable	deferred	#1455	no consumer yet; the toolchain digests with the repository's own bin/kofun-digest, not a stdlib API
-compression-archive	module	deferred	#1475	no consumer yet; belongs to the module tier when one appears
+compression-archive	module	deferred	#644	gzip response decompression is named by the accepted http-client contract; no archive format has a consumer and tar is not in scope
 mime	module	deferred	#644	only meaningful once the HTTP client contract fixes content negotiation
 crypto-tls	module	deferred	#644	security-critical and independently versioned; must not enter the portable tier
 yaml	none	non-goal	charter	TOML is the baseline config format; a second one splits the toolchain's own config
--- a/stdlib/check-capabilities.sh
+++ b/stdlib/check-capabilities.sh
@@ -133,7 +133,24 @@ test -s "$WORK/open-issues" ||
     fail "backlog snapshot contains no issue numbers: $SNAPSHOT"
 HORIZON=$(tail -n 1 "$WORK/open-issues")
 
-# 0 live, 1 all-closed, 2 undecidable (every citation above the horizon).
+# One evidence field's verdict: `live`, `closed`, or `unchecked`.
+#
+# The verdict is printed rather than returned, and that is the whole reason this
+# function has this shape. It used to answer with an exit status -- 0 live, 1
+# all-closed, 2 undecidable -- and a non-zero exit status is data the shell also
+# reads as failure. `validate_liveness` is called as `$(...)`, so under `set -e`
+# a shell that applies `-e` inside a command substitution kills that subshell on
+# the first `unchecked` row. The gate then reported "a planned or deferred row
+# cites only closed issues" while its own re-scan -- which calls the same
+# function in a tested context, where `-e` does not apply -- found no such row
+# and printed nothing. A failure naming a row it cannot then name is the shape
+# of this bug.
+#
+# It stayed hidden because it needs both halves at once: `dash` applies `-e`
+# inside a command substitution and `bash` does not, so it passed on a developer
+# machine and failed in CI; and it needs a row citing an issue *newer* than the
+# snapshot, which no row did until #1510. A verdict on stdout cannot be mistaken
+# for a failure, in any shell.
 citation_liveness() {
     liveness_unknown=0
     for ref in $1; do
@@ -142,14 +159,18 @@ citation_liveness() {
             ''|*[!0-9]*) continue ;;
         esac
         if grep -Fxq "$number" "$WORK/open-issues"; then
+            printf 'live\n'
             return 0
         fi
         if test "$number" -gt "$HORIZON"; then
             liveness_unknown=1
         fi
     done
-    test "$liveness_unknown" -eq 0 || return 2
-    return 1
+    if test "$liveness_unknown" -eq 0; then
+        printf 'closed\n'
+    else
+        printf 'unchecked\n'
+    fi
 }
 
 validate_liveness() {
@@ -158,10 +179,9 @@ validate_liveness() {
     while IFS='	' read -r job tier state evidence note; do
         case $job in ''|'#'*|job) continue ;; esac
         case $state in planned|deferred) ;; *) continue ;; esac
-        citation_liveness "$evidence"
-        case $? in
-            0) liveness_decided=$((liveness_decided + 1)) ;;
-            1) liveness_decided=$((liveness_decided + 1)); return 1 ;;
+        case $(citation_liveness "$evidence") in
+            live) liveness_decided=$((liveness_decided + 1)) ;;
+            closed) liveness_decided=$((liveness_decided + 1)); return 1 ;;
             *) ;;
         esac
         : "$tier" "$note"
@@ -181,8 +201,7 @@ decided=$(validate_liveness "$MATRIX") || {
     while IFS='	' read -r job tier state evidence note; do
         case $job in ''|'#'*|job) continue ;; esac
         case $state in planned|deferred) ;; *) continue ;; esac
-        citation_liveness "$evidence" && continue
-        test $? -eq 1 || continue
+        test "$(citation_liveness "$evidence")" = closed || continue
         printf '  %s is %s against %s, all closed\n' "$job" "$state" "$evidence" >&2
         : "$tier" "$note"
     done <"$MATRIX"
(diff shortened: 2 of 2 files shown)


hjosugi:
CI diagnosis: this is a deterministic `/bin/sh` portability failure exposed by the two new above-horizon issue references, not a bad capability row and not a flaky checkout.

Evidence:

- CI checked out synthetic merge `844280781d7839040992839d11ac6182f6332697` exactly.
- `sh stdlib/check-capabilities.sh` is green on that ref here, where `/bin/sh` resolves to bash.
- GitHub Ubuntu runs the shebang under dash. In `validate_liveness`, `citation_liveness "$evidence"` is a bare command inside a command substitution while `set -e` is active. #1510/#1511 are above horizon #1480, so `citation_liveness` deliberately returns 2. Under dash the substitution exits before the following `case $?` can classify 2 as undecidable; the outer handler prints the generic all-closed failure and no row detail. Before this PR every relevant citation was <= horizon, so that path was not exercised in CI.

Make the status capture an `if`, where nonzero is expected and `errexit` cannot preempt it:

```sh
if citation_liveness "$evidence"; then
    liveness_status=0
else
    liveness_status=$?
fi
case $liveness_status in
    0) liveness_decided=$((liveness_decided + 1)) ;;
    1) liveness_decided=$((liveness_decided + 1)); return 1 ;;
    *) ;;
esac
```

Please also add a self-test with an evidence reference above the committed horizon and run the gate through a POSIX shell in CI, so a future refactor cannot make “undecidable” exit before it is counted. The current two retries failed at the same first above-horizon row, so another retry without this change will not help.

hjosugi:
Second commit, and it is the more interesting half: **this PR turned CI red, and the cause was not the rows.**

`task capabilities` printed

```
capabilities: a planned or deferred row cites only closed issues
capabilities: a state that describes finished work reads as "someone is on this"
```

with nothing between the last two lines. The failure path re-scans the matrix to
name the offending row, and it named none — the gate failed on a row its own
re-scan could not find. Green here, red there, twice, deterministically.

`citation_liveness` answered with an exit status: 0 live, 1 all-closed, 2
undecidable. `validate_liveness` is called as
`decided=$(validate_liveness "$MATRIX") || { … }`, and under `set -eu` a shell
that applies `-e` inside a command substitution kills that subshell the instant a
verdict of 2 comes back — before any row is judged. The `||` block then runs and
re-scans through `citation_liveness "$evidence" && continue`, a *tested* context
where `-e` does not apply, so the re-scan is correct and finds nothing.

It needed both halves to stay hidden:

```sh
set -e; inner() { return 2; }; outer() { inner; printf "14\n"; }
x=$(outer) || printf "outer-failed\n"; printf "x=[%s]\n" "$x"
# bash: x=[14]     zsh: x=[14]     dash aborts
```

CI's `sh` is dash; ours is bash. And it needed a row citing an issue **above the
committed snapshot's horizon** (#1480) — which no row had until this PR pointed
two of them at #1510 and #1511. The horizon paragraph in that file promises such
citations are "reported as *not checked* and count for neither answer". Under
CI's shell they failed the gate instead.

The fix is to print the verdict — `live` / `closed` / `unchecked` — rather than
return it. A verdict on stdout cannot be mistaken for a failure in any shell.

The test, `newer_than_snapshot`, is deliberately written in the `$(…) || …` shape
of the real call site rather than the convenient one: `validate_complete` reaches
the same function through a tested context, which is exactly why twelve
mutations and the existing positive case never saw this. It asserts that the
pass does not abort, that the decided count drops by exactly one, and that the
matrix is still accepted.

**It cannot fail on bash**, which is stated in the commit rather than hidden —
bash never had the bug. CI is the only place that can confirm the fix, which is
what the run on this commit is for.

Worth knowing for anyone else touching the matrix: until this lands, any capability row re-pointed at an issue newer than #1480 turns CI red for a reason unrelated to the change.

hjosugi:
Independent release-integration diagnosis of the red Kofun verification:

- The tested merge SHA was `844280781d7839040992839d11ac6182f6332697`; the PR changes only three evidence references in `stdlib/capabilities.tsv`.
- Applying those exact substitutions to the base and running the focused capability gate passes: `#1510` and `#1511` are deliberately undecidable above the committed snapshot horizon `#1480`, while `#644` is present and open.
- The CI log emitted the generic “planned or deferred row cites only closed issues” line but no indented offending row. The branch immediately re-runs the same liveness classifier to print such rows, and found none. That makes the immutable inputs internally contradictory.

The smallest warranted action is to rerun the failed workflow; no TSV content change or backlog refresh is justified. If it repeats, harden `tests/stdlib/check-capabilities.sh` separately by using an invocation-unique work directory (the current fixed default is `build/stdlib-capabilities`) and preserving the first failing row/status for diagnostics. Release evidence and native-host evidence already passed on this PR.

hjosugi:
Correction to my immediately preceding work-directory diagnosis: the newer shell-specific reproduction supersedes it. I withdraw the transient-work-dir inference.

The decisive evidence is that GitHub runs this script under dash, while the local focused run used bash. With `set -e`, dash exits the command substitution when the deliberately “undecidable” liveness path returns 2, before `validate_liveness` can classify that status; the later diagnostic re-scan is in a tested context and therefore correctly finds no closed row. That fully explains the internally contradictory log and why the first above-horizon references exposed it.

The new head `393af9c4b7a5badd0ff1fc7ebd7008988a657b33` addresses the actual cause and adds the relevant regression shape. The in-progress CI at this exact head is the correct validation; another rerun of the old head would not have helped.

hjosugi:
Release-integration disposition request:

Head `393af9c4b7a5badd0ff1fc7ebd7008988a657b33` is now non-draft, MERGEABLE/CLEAN, with Kofun verification, release evidence, and native-host evidence all green. The retry also crossed the unrelated main-branch C ABI race; #1518/#1521 owns the real race fix, so it does not require another change in this PR.

PR #1520 is stacked directly on this head and #1522 is stacked behind #1520. Please either:

1. merge #1512 now and post the terminal `merged` event for `claude-1475-stdlib-capability-detail-20260820` before closing #1475; or
2. explicitly hand this exact head to the release-integration session for merge, including the desired #1475 claim/closure disposition.

After it lands, #1520 must retarget/rebase to `main`; the feature-branch base will not become `main` automatically. I will not mutate the PR or main without that owner disposition.

hjosugi:
Current-main release-integration update, without taking ownership:

I merged exact head `393af9c4b7a5badd0ff1fc7ebd7008988a657b33` into an isolated checkout of current `main@eeebfb8bcd3dd2a63bc39489292e9293415ebb46` with `--no-commit`. It is a clean two-file merge. `git diff --cached --check`, shell syntax, `task capabilities`, `task release-evidence`, `task release-claims`, and the committed-snapshot `task backlog` all passed; evidence regeneration created no additional diff. This host has no `dash` binary, while the exact-head Ubuntu CI already exercised the shebang under dash and is green.

So current-main compatibility is prepared, but the live `claude-1475-stdlib-capability-detail-20260820` claim remains yours. The earlier request still stands: please either merge/terminal #1475 yourself or explicitly hand this exact head and claim/closure disposition to the release-integration lane. No ref, PR state, claim, or source file was changed by this check.

hjosugi:
Resolved — no owner action remains. The merge and this cutoff note crossed in flight: exact head `393af9c4b7a5badd0ff1fc7ebd7008988a657b33` merged at 2026-08-20 15:00:30Z, while this note was posted three minutes later from the preceding poll. The owner then posted the canonical terminal `merged` event before #1475 closed, so claim ordering is correct. The recorded current-main compatibility evidence remains valid; release integration is now following post-merge main CI at `9a418654744fb5e848ffeb2c7f6e7f078a232721`.

Issue #1515, opened 2026-08-20
Stage 2 counts a trailing comma as a parameter and as an argument


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated compiler child
- Kind: implementation
- Area: Stage 2 / parser
- Priority: P2
- Size: M
- Blocked by: none
- Audited against: `origin/main@6aef329f39b015399c54adf2dda75ff2f162ab62` (re-measured 2026-09-21; the 2026-08-21 stamp was `05f4f1d7e8387d4f4e9a1f07280f982089b9256b`)

## Goal

Make the canonical Stage 2 pair follow the accepted delimiter rule for each
surface instead of counting a terminal comma as a value that does not exist.

- **Amended 2026-09-21 (owner's decision, recorded in the comment of that
  date and in `spec/syntax/call-arguments-v1.md` §Amendment: one terminal
  comma):** a parameter list or call-argument list may end with exactly one
  comma, which separates nothing — it is not a member and changes no arity.
  The original bullet ("refused directly as `trailing-comma`") was measured
  against the tree and found to contradict `spec/grammar.ebnf`, `records-v1`,
  the other frontends, and 40 of the repository's own sources.
- A comma with nothing before it — `(,` or `,,` — is refused by name
  (`E2S182`) at that comma, in parameter lists, argument lists, and
  `List[Int]` literals.
- The bounded `List[Int]` literal accepts one trailing comma, matching the
  normative list production.
- The already accepted record declaration/construction behavior stays accepted.

No diagnostic may invent an arity larger than the parameters or arguments
written in the source.

## Accepted contract and decision

`docs/ROADMAP.md` names `spec/syntax/call-arguments-v1.md` as the accepted
contract for #625. Its grammar has no optional terminal comma in parameter or
argument lists. The owning parser refuses both with category `trailing-comma`,
and the owning corpus pins those refusals. On the audited main:

```sh
task call-arguments-surface
# PASS: 22 accepted forms and 26 source-located refusals, including both
# parameter-list and argument-list trailing-comma cases
```

`spec/syntax/EXPRESSIONS_AND_LITERALS.md` is the intended full-language
contract. Its list production explicitly includes `[ "," ]`, while its status
text says that planned syntax is not automatically current bootstrap behavior.
The specific accepted call-arguments v1 contract therefore controls calls and
parameters; changing it to accept a trailing comma would be a contract
amendment, not this implementation issue.

[...]

Issue #1517, opened 2026-08-20
A read Bytes borrow may be passed to edit, and only C notices


hjosugi:
## Metadata

- State: ready
- Tracker role: curated implementation issue
- Kind: implementation
- Area: Ownership / Stage 2 Bytes crossing
- Priority: P1
- Size: M
- Related: #1315 owns the carrier; #1516 owns temporary ordinary arguments; #1561 owns duplicate-owner slot conflicts
- Blocked by: none
- Audited commit: `a858e0fa6f0060c4ecc2d0299852f4282014a9b5`

## Goal

Enforce the accepted ownership rule at every currently admitted Stage 2 Bytes call crossing: a `read` borrow cannot satisfy an `edit` slot, while an owner or `edit` borrow may satisfy `edit`, and `edit` may widen to `read`.

This implements the existing memory model and does not choose a new language rule.

## Accepted rule

- `docs/MEMORY_MODEL.md` §3.1 defines `read T` as a read-only, non-owning view.
- §3.2 defines `edit T` as an exclusive mutable view.
- §4 forbids `live + active read -> edit/take`.
- `docs/DESIGN_DECISIONS.md` DD-007 requires mutable access for mutation.

Therefore only the narrow `available read, required edit` transition is refused in this issue. Do not order ownership modes numerically or infer a broader lattice.

## Current behavior and evidence

Measured on current `main` at `a858e0fa6f0060c4ecc2d0299852f4282014a9b5`.

An ordinary positional crossing:

```kofun
fn writes(edit b: Bytes) -> Int {
    stage2_bytes_assign_zeroed(b, 4)
    return 0
}

fn lends(read b: Bytes) -> Int {
    return writes(b)
}

fn main() -> Int {
    let owned = stage2_bytes_empty()
    print(lends(owned))
    return 0
}
```

```sh
bin/kofun build /tmp/kofun-1517-read-to-edit.kofun \
  -o /tmp/kofun-1517.bin --emit-c /tmp/kofun-1517.c
# exit 1; .c exists; .bin absent
# host C: passing argument discards `const` qualifier [-Werror=discarded-qualifiers]
```

The seven mutating builtin destinations from a `read Bytes` formal—`assign_zeroed`, `byte_set`, `clear`, `reserve`, `append`, `append_range`, and `append_self`—have the same result: Stage 2 commits C and only the host compiler refuses discarded `const`.

[...]

PR #1520, opened 2026-08-20
stdlib: entropy is obtained completely, or not at all (#1511)

### LINKED ISSUE #1511: Standard library: name the secure-randomness boundary over getrandom(2)

hjosugi:
Closes #1511. **Stacked on #1512** — base is that branch, so the diff here is only the entropy work. It retargets to `main` automatically when #1512 merges.

`stdlib/random/README.md` closed its security boundary with

> Security-sensitive callers must use the platform entropy API directly.

and *directly* resolved to `random_fill` in `stdlib/linux_x86_64/process.kofun`: one `getrandom(2)` call that surfaces `EINTR` and returns `Ok(n)` with `n` possibly smaller than the buffer, with nothing saying that is not success. A caller who wanted key material and believed that `Ok` was handed a partially filled buffer. The retry loop that makes `stdlib/random`'s eight seeding bytes trustworthy lived inside `random_from_system` and was never available to them.

`entropy_fill` is that loop, for any length, with the guarantee stated.

### Why `stdlib/entropy/` and not `stdlib/random/`

`stdlib/random/tests/verify.sh:52` asserts that exactly one file under that directory obtains system entropy — *"system entropy must remain in one explicit adapter"*. A second caller added there would have **broken** that assertion rather than extended it. This gate now also checks that assertion is still present and unreworded, so the rule cannot be quietly weakened by whoever finds it inconvenient next.

`flags` is not a parameter, deliberately: `GRND_RANDOM` and `GRND_NONBLOCK` change whether the call blocks, and a capability whose blocking behaviour depends on a caller-supplied integer cannot state one guarantee.

### The evidence, and what it does not cover

The adapter cannot be compiled or run, and I found that by trying:

```
$ bin/kofun check stdlib/linux_x86_64/bytes.kofun
error[E2S02]: expected top-level `fn`, `type`, or `let` at byte 0     # its first import
```

The module pipeline (`bin/kofun:766-830`) *does* work — I built a two-file package and ran it — but it needs a `module` header, qualified cross-module calls and `internal` visibility, none of which this layer has, and it still refuses `trusted intrinsic fn` at top level (`E2S50`). So `raw_getrandom` is unreachable from any compilable program however the code is arranged. That is why `stdlib/tests/verify.sh` greps `file_roundtrip.kofun` for the five calls it must make rather than compiling it.

So the gate is three instruments, in the shape `stdlib/random` already ships:

| instrument | what it holds |
| --- | --- |
| source properties on the adapter | the `EINTR` retry, the continuation from the offset already filled, the request for exactly the bytes still missing with flags `0`, the impossible-count guard in both directions, and the absence of every fallback name |
| an executable Int-Core projection | the loop over nine scripted syscall sequences — one full fill, a fill across three calls, `EINTR` first / middle / thrice repeated, an empty destination, a zero return, a hard error, an over-long return — against a committed golden |
| an independent C11 oracle | written from the adapter's rules rather than translated from the projection, running the same scripts, compared byte for byte |

A wrong loop has to be written twice, in two languages, to pass. `oracle-binding.json` binds both digests so neither drifts silently, and it records what is **not** proved: that the real `getrandom(2)` is called with the right arguments. That is the same gap `stdlib/linux_x86_64/io.kofun` has, closed the same way — by a committed native image — which this child does not add.

Six of the nine scripts are named individually in the gate, so a golden that changes shape cannot be silently re-blessed: an `EINTR` surfaced instead of retried, an empty destination that called the syscall, a zero return that did not become a typed error, a hard error that lost its `errno`, an over-long return that did not preserve the offset already filled, and repeated `EINTR` that did not cost one call each.

### Validation

```
sh stdlib/entropy/tests/verify.sh
entropy adapter source properties: PASS
entropy fill loop against its golden: PASS
entropy C11 differential oracle: PASS
entropy oracle bound to the source it mirrors: PASS

sh stdlib/tests/verify.sh          # reached from `task stdlib`; runs to completion
sh stdlib/random/tests/verify.sh   # unchanged, still four PASSes
sh stdlib/check-capabilities.sh    # 42 rows, secure-randomness now `implemented`
node tooling/forbidden-requirements/check.mjs
PASS: 662 of 662 detected uses are in the ledger, and every ledger row still describes a use
```

### Two things a reviewer should know

**The census gains two rows**, derived rather than chosen: `stdlib/entropy/tests/verify.sh` invokes `cc` unguarded and is itself shell, so both land as `required-today`/`required`. Regenerated with `--count` and spliced under the existing header; nothing else moved. This is a file @kofun-60's #1509 also appends to — its row is elsewhere in the file, so at worst this is a trivial conflict, and I will rebase behind #1509 rather than the other way round.

**One trailing comma is deliberately absent.** `field_result`'s parameter list in the projection has none, because Stage 2 counts one as an extra parameter — `error[E2S17]: Core function 'field_result' expects 6 arguments, got 5` — though `spec/syntax/EXPRESSIONS_AND_LITERALS.md:17` says delimited lists may have one. Filed as **#1515** with a two-line repro. The adapter keeps the trailing commas its neighbours in `stdlib/linux_x86_64` use, since it is not compiled either.


hjosugi:
## Integration review: `implemented` still overstates executable evidence

The focused entropy/capability/census gates pass at `373a1cba`, and the revised limitations are much more honest. This head still needs changes before merge:

1. **The matrix state conflicts with the charter.** `docs/STANDARD_LIBRARY_CHARTER.md` defines an adapter's target boundary and `implemented` as executable today. This PR explicitly states that the canonical `stdlib/entropy/linux_x86_64.kofun` cannot compile or run on any target and that no gate executes real `getrandom(2)`, yet changes the row to `adapter / implemented / task stdlib` and tells users keys/nonces “come from” this API. The projection and C oracle prove a loop policy, not an executable entropy adapter. Keep the row `planned`/narrow the user-facing claim and file the executable boundary as a follow-up, or supply the native executable evidence the charter requires.
2. **The revised offset-trace criterion is not implemented.** `checkpoint.kofun` and the golden emit script id, outcome, detail, call count, and final offset only. They do not show `0→3→5→8` or the requested remaining length at each call. The C oracle duplicates those final observations. Emit and compare the per-call offset/request trace so short-read continuation is observed rather than inferred from final state plus source grep.
3. **Two new Kofun inputs are absent from pair coverage.** `stdlib/entropy/linux_x86_64.kofun` and `stdlib/entropy/tests/checkpoint.kofun` are not in `tests/pair-coverage/inputs.tsv`; the exact tracked-input measurement will reject both. Coordinate this with #1321's serialized final ledger measurement rather than adding more uncensused inputs.
4. **The ownership limitation is still missing.** `entropy_fill(edit destination)` is source-grepped, not executed, and #1517's known `read Bytes → edit Bytes` hole remains open. The README/capability limitation should say that the edit boundary is not compiler-enforced today; it cannot yet be presented as a safe callable destination boundary.
5. **#1511 state is inconsistent.** Its label is `in-progress` while the body still says `State: ready`. Update both copies of the state together before live-backlog CI.

The source loop itself reads correctly, the nine-script projection and C oracle agree, both binding digests match the blobs, the two census rows are accurate, and the release-evidence digest is current. Those positives do not close the executable-evidence gap.

Stacking is correct today: this PR targets #1512's exact head `393af9c4`. After #1512 merges, explicitly retarget/rebase #1520 to `main` and rerun checks; do not assume the feature-branch base becomes main automatically. This review made no edits and ran no pair gate.

Issue #1496, opened 2026-08-15
Two gates silently check less when readelf is absent, and report success


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: test-quality
- Area: Quality / Toolchain
- Priority: P2
- Size: S
- Parent: #1451
- Blocked by: none
- Audited against: `origin/main@512ba294e3a8d9f9722e32817d00e7c73c2aa074`

## Goal

Make the two gates that check less when `readelf` is absent say so, instead of
reporting success.

## Current behavior and evidence

Measured on `origin/main@512ba294e3a8d9f9722e32817d00e7c73c2aa074` while measuring the tool class for #1467.
`readelf` is invoked at nine sites. Seven declare it a required tool and fail
by name without it. Two do not:

| gate | task | without `readelf` |
| --- | --- | --- |
| `bootstrap/c_abi/check.sh` | `c-abi` | `if command -v readelf` — the dynamic-section assertions are skipped |
| `stdlib/tests/verify.sh` | `stdlib` | `if command -v readelf` — the ELF-header assertion is skipped |

```sh
grep -A2 -E 'command -v readelf' bootstrap/c_abi/check.sh stdlib/tests/verify.sh
# both open an `if`; neither branch prints a SKIP, and neither has an else
```

Neither prints a `SKIP:` line. On a host without binutils both gates report
success while checking strictly less than their name says, and nothing in the
output distinguishes that run from a complete one.

This is the failure the repository refuses everywhere else. `tests/lsp`,
`bootstrap/native/check-macho64.sh` and `bootstrap/c_abi/check.sh` itself — for
`ar`, three lines away — all print `SKIP: <what> (<why>)` when they degrade.
The same file therefore both models the right behaviour and omits it.

## Scope

In scope:

- print a `SKIP:` naming what was not checked and why, on the `else` of both
  conditionals, in the wording the sibling skips already use;
- assert the skip is reachable: run each gate with a `PATH` that has no
  `readelf` and require the `SKIP` line, because a skip nobody has executed is
  a branch, not a behaviour.

Out of scope:

- whether `readelf` belongs in `forbidden_core_build_requirements` (#1467
  owns that decision, and this issue is correct either way — a hard
  requirement makes both conditionals dead code, which is a different edit
  someone must then make deliberately);
- the seven sites that already fail by name;
- `ar`, `file`, `ldd` and `nm`, which have their own guarded/unguarded split
  recorded on #1467.

[...]

Issue #1519, opened 2026-08-20
verify: isolate optional-tool-skips from the concurrent C ABI gate


hjosugi:
## Goal

Give both C ABI build layers reached from `tests/tooling/optional-tool-skips/check.sh` explicit roots beneath that gate temporary directory, so `task verify` cannot make the nested run race the independently scheduled `c-abi` task.

## Scope

In scope:

- isolate `bootstrap/c_abi/check.sh` through `KOFUN_C_ABI_WORK`;
- isolate the `bin/kofun` C ABI compiler cache through `KOFUN_C_ABI_BUILD_DIR`;
- pin both roots with executable artifact assertions;
- keep the existing without-`readelf` observations unchanged.

Out of scope:

- redesigning all verify work namespaces;
- changing which tools are optional;
- changing C ABI behavior or release claims;
- any canonical Stage 2 pair file.

## Current behavior and evidence

Current `main` at `f42cfdd383b328535a82bd4b2e365e74c8ccb9f7` schedules both `c-abi` and `optional-tool-skips` in the bounded parallel `task verify` lane. The optional gate invokes the same C ABI script recursively. That script defaults its own work tree to `build/c-abi` and deletes it at entry; its calls to `bin/kofun` independently default the C ABI compiler cache to the same directory:

```sh
sed -n "560,585p" Taskfile.yml
sed -n "1,10p" bootstrap/c_abi/check.sh
sed -n "60,70p" bin/kofun
sed -n "78,94p" tests/tooling/optional-tool-skips/check.sh
# c-abi and optional-tool-skips are separate verify tasks
# KOFUN_C_ABI_WORK and KOFUN_C_ABI_BUILD_DIR both default under build/c-abi
# the nested invocation overrides neither variable
```

The relevant files at the failing PR merge have the same blobs as this stamped main. CI run `32374826214`, job `96443582936`, observed:

```text
FAIL: optional-tool-skips: c-abi exits 0 without readelf: expected -eq 0, got 126
```

An isolated two-process reproduction against one explicit work root produced statuses `126/0`; the loser reported `kofun-c-abi: Text file busy`. Repeating with two distinct explicit roots produced `0/0` with empty stderr. This establishes a work-root race rather than a capability-row defect or missing tool.

## Acceptance criteria

[...]

Issue #1467, opened 2026-08-15
Decide how object-image inspection tools fit the self-contained native-toolchain contract


hjosugi:
## Metadata

- State: needs-decision
- Tracker role: curated implementation child
- Kind: implementation
- Area: Bootstrap / Toolchain
- Priority: P1
- Size: S
- Parent: #1451
- Decision owner: hjosugi, as RFC-0018's shepherd
- Blocked by: none
- Audited against: `origin/main@f80d9f0c656f2f05b2ab31f923e557893e27ad60`

## Goal

Decide how external object-image inspection tools such as `readelf`, `nm`,
`objdump`, and `otool` fit the self-contained native-toolchain contract, then
make the chosen boundary visible to the repository's census or prerequisite
checks.

This issue deliberately does not choose an outcome. RFC-0018 names `hjosugi`
as shepherd, so that owner must select and record the contract interpretation
or amendment before implementation becomes ready.

## Current behavior and evidence

`spec/native-toolchain-v1/contract.json` currently has 14
`forbidden_core_build_requirements`; it does not name `readelf` or an
object-image-inspector class. Its four `allowed_external_boundary` entries are
runtime/platform ABI boundaries. RFC-0018 says those boundaries are not hidden
requirements for compiling, linking, packaging, testing, or rebuilding, so the
existing `operating-system-loader-and-system-api` entry does not by itself
authorize a host `readelf` used by verification gates.

The forbidden-requirements census intentionally takes its vocabulary from the
contract. It therefore cannot report `readelf` until a decision either adds a
forbidden class or assigns reader-facing prerequisites to a separately checked
region.

On the audited revision, a scan using the census detector's comment stripping
and command-position rules finds **57 detector-equivalent matches in 11
files**. These are detector matches (including guards and declarations), not a
claim that 57 child processes are spawned:

The baseline is reproducible with the repository detector itself (fixtures are
excluded just as they are by the census check):

```sh
node --input-type=module <<'NODE'
import { execFileSync } from "node:child_process"
import { readFileSync, statSync } from "node:fs"
import {
  commandPosition, dialectOf, withoutComments,
} from "./tooling/forbidden-requirements/detect.mjs"

const paths = execFileSync("git", ["ls-files", "-z"])
  .toString("utf8").split("\0").filter(Boolean)
const rows = []
for (const path of paths) {
  if (path.startsWith("tooling/forbidden-requirements/fixtures/")) continue
  let body
  try {
[...]

PR #1521, opened 2026-08-20
verify: isolate the nested C ABI gate workdir (#1518)

### LINKED ISSUE #1518: optional-tool-skips races the direct C-ABI gate over one work directory
## Metadata

- State: in-progress
- Kind: test-quality
- Area: Quality / Bootstrap
- Priority: P1
- Size: S
- Parent: #1496
- Blocked by: none

## Goal

Make the `optional-tool-skips` proof use a private C-ABI work directory, so
`task verify` cannot race that nested proof against the direct `c-abi` target.

Measured on PR #1512's synthetic merge ref for
`agent/stdlib-capability-detail-1475@393af9c4`, whose base is
`origin/main@f42cfdd383b328535a82bd4b2e365e74c8ccb9f7`, in CI run
`32374826214`, job `96443582936`:

```text
13:35:27 optional-tool-skips starts bootstrap/c_abi/check.sh
13:35:34 direct c-abi starts bootstrap/c_abi/check.sh
13:35:36 direct c-abi emits build/c-abi/kofun-caller
13:35:36 optional-tool-skips reports nested c-abi status 126
```

Both invocations resolve `WORK` to the same default in
`bootstrap/c_abi/check.sh:5` and both run `rm -rf "$WORK"` at line 17. The
focused proof therefore deletes/recreates the direct gate's live directory, or
vice versa. The full CI failure was:

```text
FAIL: optional-tool-skips: c-abi exits 0 without readelf: expected -eq 0, got 126
```

The nested check already provides the isolation seam:

```sh
rg -n 'KOFUN_C_ABI_WORK|rm -rf.*WORK|run_degraded c-abi' \
  bootstrap/c_abi/check.sh tests/tooling/optional-tool-skips/check.sh
# bootstrap/c_abi/check.sh:5:WORK=${KOFUN_C_ABI_WORK:-"$ROOT/build/c-abi"}
# bootstrap/c_abi/check.sh:17:rm -rf "$WORK"
# tests/tooling/optional-tool-skips/check.sh:82:run_degraded c-abi ...
```

No open issue or PR currently owns work-directory isolation for this gate.
Closed #1496/#1497 introduced the nested run but did not exercise it alongside
the aggregate verification fan-out.

## Scope

In scope:

- invoke the nested C-ABI degraded run with `KOFUN_C_ABI_WORK` set to a child of
  the proof's private `mktemp` directory;
- assert the nested run produced its expected executable inside that private
  directory, so removing the override makes the focused gate fail;
- keep the direct C-ABI gate's default and aggregate verification parallelism.

Out of scope:

- serializing all `task verify` targets;
- changing `bootstrap/c_abi/check.sh`'s default work directory;
- changing any C-ABI compilation, runtime, or `readelf` semantics;
- the unrelated capabilities change that happened to expose the race.

## Acceptance criteria

- [ ] `optional-tool-skips` passes `KOFUN_C_ABI_WORK` to its nested C-ABI run,
      using a directory owned by that proof.
- [ ] The focused proof asserts an expected C-ABI executable exists in that
      private directory.
- [ ] The nested and direct C-ABI gates can run concurrently without deleting
      each other's artifacts.
- [ ] The existing no-`readelf` exit status, SKIP line, and withheld PASS claim
      remain unchanged.

## Validation

| --- | --- | --- |
| Focused gate | `task optional-tool-skips` | green; private C-ABI artifact asserted |
| Concurrent regression | `task --parallel c-abi optional-tool-skips` | green across repeated runs |
| Patch hygiene | `git diff --check` | no output, exit 0 |

## Dependencies

Blocked by: none.

hjosugi:
Closes #1518

`task verify` runs `c-abi` and `optional-tool-skips` in parallel. The latter
invokes the C-ABI gate again under a PATH without `readelf`, but both invocations
used the default `build/c-abi` and each began by recursively replacing it. CI run
32374826214 caught the direct and nested gates deleting/recreating the same
executables 29 ms apart; the nested run exited 126.

This change uses the existing `KOFUN_C_ABI_WORK` seam to put the degraded nested
run under the optional gate's private `mktemp` directory. It also asserts the
nested `kofun-caller` exists there, so removing the override is a deterministic
focused-gate failure rather than a schedule-dependent aggregate failure.

No Taskfile serialization, C-ABI semantics, canonical Stage 2 pair, or release
evidence is changed.

Validation:

- `git diff --check`
- `sh -n tests/tooling/optional-tool-skips/check.sh`
- `task optional-tool-skips`
- `task --parallel c-abi optional-tool-skips` (two consecutive green runs)

### DIFF
--- a/tests/tooling/optional-tool-skips/check.sh
+++ b/tests/tooling/optional-tool-skips/check.sh
@@ -80,7 +80,9 @@ run_degraded() {
     script=$2
     expected=$3
     set +e
-    PATH="$WORK/bin" sh "$ROOT/$script" >"$WORK/$label.stdout" 2>"$WORK/$label.stderr"
+    KOFUN_C_ABI_WORK="$WORK/c-abi-work" \
+        PATH="$WORK/bin" sh "$ROOT/$script" \
+        >"$WORK/$label.stdout" 2>"$WORK/$label.stderr"
     status=$?
     set -e
     assert_num "$label exits 0 without readelf" "$status" -eq 0
@@ -90,6 +92,8 @@ run_degraded() {
 
 run_degraded c-abi bootstrap/c_abi/check.sh \
     'SKIP: dynamic linkage of the C ABI output (readelf unavailable)'
+assert_executable 'degraded C ABI executable in private work directory' \
+    "$WORK/c-abi-work/kofun-caller"
 # The claim its `readelf` reads must not be printed when it did not read it.
 if grep -Fq 'PASS: the C ABI output is a dynamically linked executable' \
     "$WORK/c-abi.stdout"


hjosugi:
### P1: the nested driver still re-enters the shared C ABI build directory

The `KOFUN_C_ABI_WORK` override correctly isolates `bootstrap/c_abi/check.sh` own `WORK`, but this script invokes `bin/kofun` repeatedly. `bin/kofun:65` independently sets:

```sh
C_ABI_BUILD_DIR=${KOFUN_C_ABI_BUILD_DIR:-"$ROOT/build/c-abi"}
```

and `ensure_c_abi_compiler` compiles/executes `$C_ABI_BUILD_DIR/kofun-c-abi`. The nested proof therefore returns to the direct gate shared `build/c-abi` after the new private-work assertion. Two parallel green runs can be scheduling luck; the original shared-root reproduction itself sometimes returned 0/0.

Please isolate both layers on the nested command:

```sh
KOFUN_C_ABI_WORK="$WORK/c-abi-work" \
KOFUN_C_ABI_BUILD_DIR="$WORK/c-abi-build" \
PATH="$WORK/bin" sh ...
```

and assert both `$WORK/c-abi-work/kofun-c-abi` (or its final caller) and `$WORK/c-abi-build/kofun-c-abi`. Independent focused and two-worker validation with both overrides is green. Local commit `0f58cb61987442508d20d835202bb3a8ac47a89e` contains the exact one-file form if the owner wants it handed off; it has not been pushed anywhere.

I found #1518 only after independently filing duplicate #1519. I will release and close the duplicate claim rather than compete with this PR.

hjosugi:
Reviewed from another session. **The diagnosis is right and I can confirm it independently: run 32374826214 is my PR #1512.** It failed there with `expected -eq 0, got 126`, twice, deterministically enough that I re-ran it before looking — so this fixes a real, currently-live source of red on unrelated PRs.

Three things checked rather than assumed:

- **The seam exists and is the right one.** `bootstrap/c_abi/check.sh:5` is `WORK=${KOFUN_C_ABI_WORK:-"$ROOT/build/c-abi"}`, so the override lands where the nested run needs it.
- **`assert_executable` exists** — `tests/assertions/assert.sh:160` — so the new assertion is not a silently-absent helper.
- **The nested `stdlib/tests/verify.sh` has no sibling collision.** All nineteen `stdlib/*/tests/verify.sh` use `${TMPDIR:-/tmp}/kofun-<module>-verify.$$`, and the top-level one is `$$`-based too. So `build/c-abi` really was the only fixed path shared between the direct and nested runs. (Two of those nineteen are new as of my #1520 and #1522, and both follow the same `$$` convention.)

Two notes, neither blocking:

1. **`run_degraded` now sets `KOFUN_C_ABI_WORK` for both of its callers**, including the `stdlib/tests/verify.sh` one, which ignores it. Harmless, but it reads as if the stdlib gate has a C-ABI work directory. Scoping the override to the `c-abi` case — or naming it in the label — would keep the seam legible to the next reader.

2. **The class is not asserted, only this instance.** The property is "a gate invoked nested must share no fixed work path with its outer instance", and the new `assert_executable` proves it for exactly one path. The next gate added to `run_degraded` reintroduces the class silently. One cheap form: after the degraded run, assert that `build/c-abi` was not modified — that fails if the override is dropped *and* if a future nested gate acquires a second shared path, which the current assertion does not.

   Worth weighing against the fact that this is already the second time this file has had to distinguish "the branch was not taken" from "the branch could not be taken" (#1496). A gate about degraded runs is exactly where an unasserted class tends to come back.

One adjacent finding while I was in this area, since #1467 is about the same `readelf` sites: this file is the **eleventh** `readelf` site, added by #1496, and the census counts its `command -v readelf` as an invoke because `detect.mjs:49` lists `command` among the wrappers. So the "nine files" in #1467 is now ten by its own command and eleven by the census detector — I have posted the measurement there.

Issue #1457, opened 2026-08-15
Move the package manager from package/manager.sh into Kofun


hjosugi:
## Metadata

- State: blocked
- Tracker role: curated implementation child
- Kind: implementation
- Area: Bootstrap / Toolchain
- Priority: P2
- Size: L
- Parent: #1451
- Blocked by: #1264, #1323, #1382, #1453
- Audited against: `origin/main@9e3b85328a710b3a2335e2dee8d72b3ac6fb71e1`

## Goal

Move the package manager from `package/manager.sh` into Kofun, so that
resolving a dependency does not require a shell.

## Why this component and not another

`spec/native-toolchain-v1/contract.json` lists eight `completion_evidence`
entries. One of them is *"source-free dependency build from KIF and package
artifacts"* — the package manager, described without the words. It is the only
component in #1451 the contract names twice: once as evidence, and once through
`shell-build-driver` in `forbidden_core_build_requirements`.

## Current behavior and evidence

Measured on `origin/main@9e3b85328a710b3a2335e2dee8d72b3ac6fb71e1`:

```sh
git show 9e3b85328a710b3a2335e2dee8d72b3ac6fb71e1:package/manager.sh | wc -l
# 450 — shell
```

Its own header states the boundary it works within:

> This is deliberately a small artifact resolver, not a source-package build
> system. The current Kofun compiler's external-code boundary is an explicit
> C ABI library file, so v1 resolves exactly that artifact by URL and SHA-256.

So the work has three parts that are separable and one that is not: **it
fetches over the network, verifies a SHA-256, and writes into a cache
directory.** Each of those is a capability Kofun does not have yet, which is
why this child is `blocked` rather than merely large.

## Scope

- A `kofun package` implementation in Kofun covering manifest parsing, lock
  file read and write, cache resolution, and digest verification.
- Byte-compatible lock files. An existing `kofun.packages.lock` must resolve
  identically, or every downstream project's build changes for a reason
  unrelated to its own contents.
- Deleting `package/manager.sh` once `bin/kofun`'s `package` arm calls the new
  implementation.

Out of scope: widening what a package may be — the C ABI artifact boundary is
RFC-0002's business, not this issue's; the CLI driver, which is its own child;
any change to the manifest or lock file formats.

## What the language must gain first

- #1264 `HTTP/1.1 core: executable scripted-transport vertical slice and
  capability truth` — resolution is by URL, so this is the hard one.
[...]

PR #1522, opened 2026-08-20
stdlib: a line's length is not known before it is read (#1510)

### LINKED ISSUE #1510: Standard library: a bounded buffered line reader over the Linux file adapter

hjosugi:
Closes #1510. **Stacked on #1520** (which is stacked on #1512), so the diff here is only the buffered-io work; the base retargets automatically as each merges.

`package/manager.sh:312` and `:353` are `while IFS='\t' read -r` loops, recorded at `docs/PACKAGE_MANAGER_IN_KOFUN.md:99` as this capability's requirement of the package-manager rewrite (#1457). Nothing buffered existed, and none of the three nearest readers can be bent into the shape — `file_read` takes a caller-sized destination and an explicit length, `file_read_exact` loops `EINTR` and short reads correctly but only to `len(destination)`, and `csv_parse` takes the whole document as `Text`. **A line's length is not known before it is read.**

### The rules, all decided up front

| rule | why it is this way |
| --- | --- |
| LF only; a trailing CR is preserved | stripping it silently is a decision the caller cannot undo |
| a final line with no terminator is returned | a file whose last byte is not a newline still has a last line |
| an empty line is a line of length zero | not a skipped one |
| capacity is the maximum line length **including** its terminator | a line that cannot be seen whole is `BufferedLineTooLong(capacity)` — what the caller would have to raise — and nothing is discarded when it fires |
| `EINTR` is retried, offset unchanged | the discipline `file_read_exact` already applies; surfacing it makes every caller write this loop |
| reading past the end is answered from state | no syscall, cannot block |
| the file is borrowed, not owned | `File` is affine (`file_close(take file: File)`); an owning reader would have to re-export close and drop |

### "Portable" is a measurement here, not a label

The matrix puts this row in the `portable` tier, and the charter defines that tier as including *"the portable half of any capability whose other half is a platform adapter"*. So the buffer and the line-splitting state machine are one file and the refill is another.

The portable half names exactly one module from the platform directory — `stdlib.linux_x86_64.bytes`, because the byte surface lives there and a buffer is made of bytes. That module is portable code under a platform name:

```sh
grep -oE "abi_[a-z_0-9]+" stdlib/linux_x86_64/bytes.kofun | sort -u
# abi_bytes_equal abi_bytes_zeroed abi_load_u8 abi_store_u8 abi_text_utf8_bytes
grep -cE 'raw_|syscall' stdlib/linux_x86_64/bytes.kofun
# 0
```

Each of those five forwards to a compiler intrinsic. **The gate asserts both zeros** — the portable half's own, and `bytes.kofun`'s — so the day either stops being true, the tier stops being a label. That the byte surface sits under `linux_x86_64/` at all is a layering defect this child records rather than fixes: moving it is a rename across every stdlib module and belongs to whoever owns the tier boundary.

### Evidence

Neither half compiles — `bin/kofun check` on any file in `stdlib/linux_x86_64` stops at its first `import`, and `trusted intrinsic` is refused at top level in both pipelines, so `raw_read` is unreachable from any compilable program. Same three instruments as #1520:

1. **Source properties** on both halves.
2. **An executable Int-Core projection** over twelve scripted sources: two lines; an unterminated final line; an empty source; a lone line feed; a line longer than a four-byte buffer; a line of exactly capacity-minus-terminator; one byte over; one byte per read; `EINTR` at the first read; a line spanning two refills; a NUL inside a line; and a read error after a line. `Bytes` becomes a `List[Int]` and `file_read` becomes the script — everything else is the algorithm.
3. **An independent C11 oracle**, written from the rules rather than translated from the projection, over the same scripts, compared byte for byte.

**Twelve positions in the golden are asserted by name**, so a golden that changes shape cannot be silently re-blessed.

```
sh stdlib/buffered_io/tests/verify.sh
buffered io portable half reaches no syscall: PASS
buffered io adapter contract: PASS
buffered io state machine against its golden: PASS
buffered io C11 differential oracle: PASS
buffered io oracle bound to the source it mirrors: PASS

sh stdlib/tests/verify.sh          # runs to completion, entropy and buffered io both reached
sh stdlib/check-capabilities.sh    # 42 rows, buffered-io now `implemented`
node tooling/forbidden-requirements/check.mjs
PASS: 664 of 664 detected uses are in the ledger, and every ledger row still describes a use
```

### One thing worth reading in the diff

The first version of the portable-half check failed on the portable half's **own comment**, which quotes the command being run. Comments are stripped now — a checker that reads prose is measuring the wrong file. It is the same class of mistake as a gate that greps for a call in a file nobody compiles, and it took thirty seconds to find only because the check ran.

Census gains two rows, derived rather than chosen (the new gate invokes `cc` unguarded and is itself shell), and `artifacts/release-evidence/index.json` re-pins `stdlib/tests/verify.sh`, which gained the line that reaches the gate.


hjosugi:
## Integration review: the representation decision is still unresolved

This head implements one of the choices named in the readiness demotion, but the issue/PR continues to claim the incompatible choice. It is not mergeable yet.

1. **Returned lines allocate and copy.** `buffered_slice` calls `bytes_zeroed(length)` and copies every byte for every `Line(Bytes)`. That is a legitimate bounded-allocation representation, but #1510 still requires “No operation allocates after construction” and unchanged buffer pointer/capacity. Record the allocation/copy choice in the issue, remove the contradictory criterion, and gate its bound/failure semantics—or choose a view/offset representation that actually satisfies no-allocation.
2. **The executable models do not observe the claimed representation invariants.** The Int-Core projection and C oracle use fixed list/array buffers and report only line lengths. They do not model returned `Bytes`, allocation failure, buffer pointer/capacity/file descriptor identity, aliasing, or contents. The NUL assertion is only `length == 3`, so replacing the NUL with another byte still passes. Add byte-exact results and evidence for the chosen allocation/ownership rule.
3. **`LineTooLong`/system-error recovery is not proved and the canonical algorithm cannot recover from too-long.** After compaction leaves a full buffer with no LF, `buffered_free == 0` returns `BufferedLineTooLong`; the next call sees the same full state and returns the same error forever. The projection stops at the error instead of making the required next call, and it likewise does not prove resumption after a system error. Define an actionable recovery rule and exercise it.
4. **The issue body still contradicts the source's tier boundary.** The AC says the portable half names no `stdlib/linux_x86_64` import, while line 1 imports `stdlib.linux_x86_64.bytes`. The correction appears only in a claim comment. Readiness facts belong in the body before implementation/PR state.
5. **`implemented` again conflicts with the charter.** Both canonical halves are explicitly uncompileable, real `file_read` is never executed, and `task stdlib` checks source strings plus projections. That does not meet the charter's “executable today” meaning of `implemented`. Keep the matrix row planned/narrow the user-facing claim, or supply executable native evidence.
6. **Three new Kofun inputs are missing from pair coverage:** `buffered_io.kofun`, `linux_x86_64.kofun`, and `tests/checkpoint.kofun`. Coordinate the exact inputs/undefended measurement with #1321's serialized ledger work.

The source/oracle binding and release-evidence digests match, and the stack is exact (#1512 → #1520 → #1522). But #1510 is correctly still `needs-detail`; moving its claim to `pr-open` did not resolve the representation contract. After upstream PRs land, every stacked PR also needs explicit retarget/rebase to `main` in order. This review made no edits and ran no pair gate.

hjosugi:
Re-review at `62afae11a82e761c8355736347b4b6b9fe8c1827`.

The representation decision is now clear: one fresh owned `Bytes` allocation per returned line, and the implementation source shape matches that decision. That resolves the previous direct contradiction. The new commit, however, changes only the README and source-count assertions; the implementation, projection, C oracle, capability row, and pair ledger are unchanged. The following remain blocking:

1. **The evidence still observes lengths, not bytes or ownership.** `checkpoint.kofun` and `buffered_io_reference.c` retain/print only line lengths. Replacing the interior-NUL fixture `[1, 0, 2]` with any other three non-LF bytes leaves the output unchanged. No evidence observes returned bytes, fresh-object identity, allocation failure, reader buffer pointer/capacity, or File identity. The grep/count assertions prove spelling, not the dynamic one-allocation-per-success contract.

2. **The promised recovery semantics are not implemented or exercised.** After compaction, a full buffer without LF returns `BufferedLineTooLong` while preserving the same full cursor/content state, so the next call returns the same error forever. For a system error, compaction may already mutate cursor/content before `file_read` fails, contradicting the stated untouched-on-error rule. Neither model makes a follow-up call after either error.

3. **`implemented` still conflicts with the charter.** The charter defines it as executable today; the README and oracle binding say neither canonical half compiles and `file_read_executed:false`. A projection plus source greps cannot establish a shipped executable capability.

4. **#1510 is still internally inconsistent.** Its AC says the portable half imports no `stdlib/linux_x86_64`, while `buffered_io.kofun` imports `stdlib.linux_x86_64.bytes` and the README/gate explicitly permit that exception. It also names `LineTooLong(Int)` while the implementation exposes `BufferedLineTooLong(Int)`.

5. **The pair ledger is still stale.** All three new tracked Kofun inputs have zero rows in `tests/pair-coverage/inputs.tsv`. Please integrate those through the serialized #1321 owner rather than running or editing the canonical pair concurrently.

Please make the continuation behavior achievable (or narrow the contract to a documented sticky/unrecoverable error), add byte-exact multi-call evidence after TooLong/system errors plus ownership/identity evidence, reconcile #1510's import/name wording, keep the capability planned until canonical code is executable (or separately change the charter), and obtain the #1321 ledger handoff.

Issue #1516, opened 2026-08-20
A Bytes temporary in an argument emits `&<rvalue>` instead of a refusal


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation issue
- Kind: implementation
- Area: Compiler / Stage 2 Bytes crossing
- Priority: P2
- Size: S
- Related: #1315 owns the crossing; #1321 refused the shape for the bounded
  mutation surface only; #1517 is the other shape found with it
- Blocked by: none
- Audited against: `main@41f7c3de06ff8a47f16c4844dfecfd58b592acd7`

## Goal

A `Bytes` argument that resolves to no binding is refused by name, instead of
emitting `&<rvalue>` and sending the author to a byte offset in generated C.

## Current-main remeasurement (2026-09-06)

On `main@41f7c3de06ff8a47f16c4844dfecfd58b592acd7` (`v0.11.6-seed`),
`bin/kofun build temporary_MODE_argument.kofun -o MODE.bin --emit-c MODE.c`
for each of `read`, `edit`, and `take` still emits an address of the unnamed
`stage2_bytes_empty()` result and the host compiler fails with `lvalue required
as unary & operand` at generated C line 457. The three fixtures are carried
forward from the preserved predecessor checkout. The inherited defect is not
introduced by this release queue.

The current fix refuses those crossings as E2S177 before C or binary publication.
`task bytes-carrier` owns all three cases, the read golden, argument-position
control, and a compiling missing-guard mutation for each mode. Temporary
materialization remains out of scope. User-authorized handoff and current claim
are recorded in the comments; the predecessor worktree remains unchanged.

## Historical behavior and evidence

Measured on a clean `2021ab08e344a78bd448d73d90241f521668e716` worktree, which was the audited main then and not a
consequence of #1321:

```sh
cat > temp_arg.kofun <<'SOURCE'
fn measure(read b: Bytes) -> Int {
    return 3
}

fn main() -> Int {
    print(measure(stage2_bytes_empty()))
    return 0
}
SOURCE
bin/kofun build temp_arg.kofun -o temp_arg.bin
```

```
build/bootstrap-stage1/temp_arg.c: In function 'main':
build/bootstrap-stage1/temp_arg.c:312:48: error: lvalue required as unary '&' operand
  312 |         int64_t kofun_value = kofun_fn_measure(&stage2_bytes_empty());
```

`emit_argument` takes the address of a `Bytes` argument for a `read`, `edit`,
[...]

Issue #1499, opened 2026-08-15
Stage 2 has no way for a compiled program to read a byte


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation issue
- Kind: implementation
- Area: Compiler / Stage 2 language surface
- Priority: P1
- Size: M
- Parent: #1451
- Blocked by: none
- Audited against: `origin/main@41e7464d5ad0edb8741b25aa948e531a02e75ef2` — re-measured 2026-09-21; the original stamp `cac0f8c5` predates the `Bytes[65536]` carrier
- Decision owner: the maintainer (hjosugi), because the decision moves the published `bounded-bytes` claim boundary
- Decision: **option 1** (`stage2_bytes_byte_at` produces `Int`), recorded 2026-09-21; bound and open failures of the read are runtime R-codes that leave the carrier untouched

## Goal

Give a program the C11 Stage 2 backend compiles a way to receive input, so a
toolchain command written in Kofun can read the file it is asked about.

## The title of this issue was too narrow, and the measurement is why

It was filed as "no way to read a byte", from a comment in the digest model.
Probed directly against `bin/kofun build` on
`origin/main@cac0f8c54071c2b9e24dd9816d25b5155064a6ab`, one program per builtin:

| builtin | a compiled program can call it |
| --- | --- |
| `len`, `text_slice`, `to_text`, `print` | **yes** |
| `read_text` | no — `E2S10` |
| `args` | no — `E2S10` |
| `write_text` | no — `E2S10` |
| `chars`, `find`, `contains`, `starts_with` | no — `E2S10` |
| `trim`, `is_digit`, `replace` | no — `E2S10` |

The allowlist is one condition in `bootstrap/stage2/compiler.kofun:7218`:

```
if name == "len" || name == "text_slice" ||
   name == "to_text" ||
   name == "stage2_bytes_empty" {
```

**So it is not that a compiled program cannot reach a byte. It cannot reach
anything.** No argv, no file read, no file write. It can manipulate text
literals from its own source with `len` and `text_slice`, and print.

That is why #1382's digest takes two 64-element `List[Int]`s built in the
source, and it is why every component in #1451 is stuck rather than merely
awkward: the CLI driver reads source files, the package manager reads and
verifies artifacts, the language server reads documents, and the digest
command reads the file it digests.

The compiler itself is exempt — `compiler.kofun` calls `chars` in 30 places —
because the frontend lowers what the Core does not. The restriction binds
[...]

Issue #1479, opened 2026-08-15
Clarify typed-sidecar successor routing at the frozen v1 boundary


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated contract clarification
- Kind: spec-design
- Area: Tooling / typed sidecar
- Priority: P1
- Size: S
- Parent: none
- Blocked by: none
- Decision owner: hjosugi
- Normative authority: `spec/tooling/typed-sidecar.md` §Compatibility and privacy
- Audited against: `origin/main@05f4f1d7e8387d4f4e9a1f07280f982089b9256b`

## Goal

Make the typed-sidecar version/freeze route discoverable and executable, so a
child that needs a new semantic fact can determine before implementation
whether it may reuse an existing slot or must define a successor schema.

This is a clarification of the accepted contract, not a choice to reopen v1.

## Decision already recorded

`spec/tooling/typed-sidecar.md` §Compatibility and privacy already answers the
original A/B/C question:

- unknown schema major/version is refused;
- compatible optional data is allowed only under a new schema version whose
  old-reader behavior is defined;
- a change to status meaning, NodeId input, canonicalization, disclosure, or
  replacement rules requires a new schema/media version.

Together with `spec/concurrency/scoped-captures-v1.md` §10 — “No v1 file is
extended in place” — this selects the closed-v1/successor-schema route. There
is no additive-in-place option for the closed v1 schema and readers.

The repository has already exercised this route: `0879c99b` added
`kofun.typed-sidecar/v2` and KSE2 capture schemas without changing any bytes in
the frozen v1 manifest. That successor deliberately reuses v1's node
definition, so it can add its two root fields but cannot add a node fact that
v1 does not own. A later version that changes node shape must define that
changed shape outside the frozen v1 definition; merely assigning a new root
version while `$ref`-ing the same frozen node does not create an extension
point.

No DD-028 amendment is needed for this clarification because no accepted v1
byte, reader result, status meaning, or authority boundary changes. A future
proposal that changes one of those semantics must use the RFC ledger's
amendment mechanism separately.

## Current behavior and evidence

Current `main@05f4f1d7` has the answer, but the failure site does not point to
it:

```sh
sed -n '327,341p' spec/tooling/typed-sidecar.md
# states the new-schema/version rule

sed -n '417,418p' spec/concurrency/scoped-captures-v1.md
# states only that v1 is not extended in place

sed -n '23,30p' spec/concurrency/scoped-captures-v1/check.sh
[...]

PR #1525, opened 2026-08-20
stage2: bounded mutation for Bytes[65536] (#1321)

### LINKED ISSUE #1321: Stage 2 C11: implement bounded mutation for Bytes[65536]

hjosugi:
Closes #1321.

#1315 landed the `Managed Bytes[65536]` carrier and no way to change what it
holds. This adds the nine operations of the bounded mutation surface to both
halves of the canonical Stage 2 pair, with the growth ladder, the order of the
range checks, and the status mapping frozen in the emitted runtime rather than
described in prose.

## The surface

`len`, `capacity`, `byte_at`, `byte_set`, `clear`, `reserve`, `append`,
`append_range`, `append_self`. `len` and `capacity` return `Int` and are the
two a source program can use; the rest are `Void`, because the status and the
`Stage2ByteRead` carrier are private to the emitted C and are proved there —
the same arrangement #1315 made for the 0..8 status, for the same reason.

`Stage2ByteRead` is emitted exactly once with its three tags in declaration
order 0..2 and no consumed tag. Tags 6..8 of the status belong to #1322's Text
bridge, and the gate refuses any mutation operation that reaches one.

Surfacing the read carrier to source would need a compiler-owned enum
declaration, and Stage 2 resolves an enum by scanning the source for its `type`
declaration — a type the compiler owns has no declaration site to be found at.
#1499 is the consumer that needs the byte in source and is where that mechanism
belongs.

## Two rules worth reading the code for

**`kofun_bytes_range` never computes `offset + count`.** Two in-range operands
can sum past the ceiling, and a refusal that depends on that sum is one a wrap
can skip; subtraction against the length cannot overflow and refuses the same
cases. The driver pins it with `INT64_MAX` in both slots.

**Carrier identity is settled before any C is emitted.** `append_range` copies
with `memcpy`, so it needs two distinct typed-HIR BindingIds; under #1315's
alias refusal two different `Bytes` values never share storage, which makes the
identity check the whole overlap argument rather than a run-time defence. One
value in both positions is `E2S177` and is sent to `append_self`, which is
defined for overlap and copies with `memmove`.

`E2S177` rather than the `E2S168` the issue names: `E2S168` is
`int-bit-operations`, with its own registry row and nine uses in each half.
`E2S176` was the highest code in the registry and `git grep E2S177` was empty.
Recorded on the issue.

## Three defects found on the way

All three are *uncovered in both halves* rather than divergences, so no
output-comparison gate could have reached them. All three are fixed here
because the surface cannot be used without them.

1. **`emit_argument` prefixed `&` to every borrowed `Bytes` crossing.** A
   `read`/`edit` parameter is already `const KofunBytesValue *`, so lending a
   borrow onward emitted `KofunBytesValue **` and the build failed. #1315
   admitted the borrow and shipped no fixture that lent one onward. The mode
   now comes from field 6 of the typed-HIR binding record, which is where the
   parameter head already recorded it. The same defect turned out to be in
   **three** places — `emit_argument`, the new mutation family, and
   `stage2_bytes_assign_zeroed`'s own lowering — and all three now route
   through one helper.

2. **`lower_body` concatenated `emit_expression`'s result into the emitted C
   without testing for `error[`**, in both arms of the expression-statement
   lowering. Every other caller tests it. A refusal raised by a *statement* was
   therefore emitted rather than reported: `E2S177` came out as
   `(void)error[E2S177]: ...;` and the host compiler reported an undeclared
   `error` at a byte offset in generated code. It takes a refusal that can only
   arise in statement position to reach it, which is why no #1315 fixture did —
   every alias refusal there is raised from a `let` initializer.

3. **A `Bytes` temporary in a carrier slot emitted `&<rvalue>`.** Refused here
   for the mutation surface. The same hole through `emit_argument` to a user
   function is #1315's and is filed as #1516 rather than widened into this
   change; the missing ownership rule it exposed is #1517.

## The gate

`task bytes-mutation`, named for its sibling `task bytes-carrier` — the issue's
`bytes-stage2-carrier` row names a task that has never existed, which is
recorded on the issue with the correction.

Its operation vocabulary is **derived from both halves of the pair and from the
emitted runtime**, and required to be one set three times over, because
`builtin_arity`'s own comment says the three builtin tables have to be edited
together and nothing cross-checks them. A name added to one half and not the
other now fails.

Every assertion was proved by reintroducing the defect it defends:

| Reintroduced | The gate says |
| --- | --- |
| `&` on borrows | the borrow fixture does not build |
| the statement `error[` guard removed | `same_carrier committed C` |
| one operation dropped from the Kofun half | the two halves name different mutation operations |
| negative-offset / negative-count order swapped | `got tag 2 detail -9, want tag 2 detail -3` |

## The 160-byte diagnostic bound

The first draft of `E2S177` was 185 bytes and failed `stage2-events` as a `cmp`
mismatch between `.authority` and `.producer`. `KOFUN_SEMANTIC_ERROR_DETAIL_BYTES`
is 160, and the semantic producer copies a diagnostic into a `char detail[160]`
and **truncates silently** — so the author got the whole sentence and every
consumer of the event stream got a different one, ending mid-identifier. The
longest existing Stage 2 golden is 142, so nothing had reached the bound before.

Both sentences are now 152 and 154 bytes, and the gate asserts the width where
the sentence is chosen rather than leaving it to be discovered three gates away.
Evidence posted on #1479, which is the issue about this format being frozen with
no amendment path.

## Bookkeeping

- `bootstrap/stage2/SHA256SUMS` and `trusted_seed_sha256` in
  `bootstrap/manifest.json` regenerated; `git grep` on the old digests found no
  third pin.
- `artifacts/release-evidence/index.json` regenerated with `task
  release-evidence` — hand-editing it is not enough, it pins
  `bootstrap/manifest.json` and `tests/diagnostics/registry.tsv` too.
- The two corpus counters in `tests/typed-sidecar/stage2-events.sh` move with
  the new diagnostic fixture, which that file's own comment names as an expected
  edit.
- `tooling/task-help.mjs` and `tooling/forbidden-requirements/census.tsv` gain
  the new task and its two `cc`/shell-driver rows.
- Six new `*.kofun` files, so `tests/pair-coverage/inputs.tsv` goes 1269 → 1275
  after merge; that file is owned by #1401's agent, which is re-measuring
  `undefended.tsv` after this lands rather than before.

`task verify` green.


hjosugi:
## Release-integration review at exact `cf69b669` — not mergeable yet

This is the same commit as the committed-snapshot review on #1321. The local `task verify` result is useful, and exact-head CI should finish, but neither dispositions the semantic and coverage gaps below.

### Safety blockers

1. **A `read Bytes` borrow can still flow into an `edit Bytes` slot.** The PR body explicitly leaves the ownership rule to #1517. This head makes forwarded borrows reachable, so #1517 cannot safely follow #1321 in merge order. Validate caller permission against callee mode for ordinary, labelled, pipeline, and builtin calls: `read` may reach only `read`; `edit` may widen to `read`.
2. **Private results remain accepted in value contexts.** `builtin_return_type` maps `stage2_bytes_byte_at` and the mutation statuses to `Void`, but inferred bindings and conditions can still accept those calls before emitted C receives a struct-to-integer/type error. Refuse private-result builtins outside expression-statement position before C emission, with `.c` and `.bin` no-artifact fixtures.
3. **The new builtin spellings can override declared functions during emission.** `emit_primary` takes `bytes_mutation_builtin(name)` by spelling before ordinary declared-call lowering. Resolve a declaration first, or reserve/refuse these names, and pin the chosen rule with an executable/no-artifact fixture.
4. **Direct BindingId comparison does not prove `append_range` identities across a wrapper.** A helper with distinct local `edit destination` / `read source` parameters can be called with the same owner twice, then reaches `memcpy(&value, &value, ...)`. Enforce duplicate-owner compatibility at call sites, propagate provenance, or use a safe runtime rule; add the wrapper fixture.

### Evidence and repository blockers

- `mutation_driver.c::remember/expect_unchanged` saves length/capacity/bytes but not the original data pointer. The injected-OOM branch covers append/reserve/self only, not destination and source preservation for `append_range`.
- The named value/length matrix does not run append-one for every required byte/length. The driver is strict at O2 and sanitizer-built at O1, not the named O0/O2/ASan matrix; emitted fixtures use sanitizers at O0/O2 without `-Wall -Wextra -Werror -pedantic`.
- There is no positive source fixture for `append_range(edit destination, read source, ...)`. `refuses()` removes/checks only `.c`, not `.bin`.
- `run.sh:76` greps `mutation.c` for `stage2_bytes_assign_zeroed`, but `mutation.kofun` never calls that operation, so the forbidden-form assertion passes vacuously. Point it at the exercising fixture and assert both expected and forbidden forms.
- `git diff --check af7f21a5...cf69b669` currently fails at `tests/conformance/bytes-mutation/run.sh:68` for trailing whitespace.
- `tests/pair-coverage/{drivers,inputs,undefended}.tsv` has not been regenerated for the new task, six inputs, and pair functions. The PR body proposes doing this after merge; release integration needs the ledgers correct in the commit that lands.

Please keep the canonical-pair claim live while addressing these items, then post a new exact head/handoff. Also replace `Closes #1321` until the terminal claim can be recorded before issue closure. I did not edit the pair or run pair/full gates during this review.

hjosugi:
Blocking re-review at exact head `cf69b669f66fec7418e0b6cdb03ad428cde7fc1e` (base `main@af7f21a5e6cb23055a4fd27d8c2bf8280866c4af`). Two independent read-only reviewers checked semantics and the issue's AC/evidence. The prior blockers remain, and `git diff --check main...HEAD` currently fails at `tests/conformance/bytes-mutation/run.sh:68` for trailing whitespace.

## Semantic blockers

1. **`read` can still reach `edit`.** Caller `read` and `edit` modes lower to the same carrier shape; the new builtin validator records arity/count but not required access. A `read Bytes` formal can reach a mutating runtime helper, yielding const-discarding C (and potentially mutation if warnings are not fatal). Ordinary forwarding has the same gap. This makes #1517 reachable through this diff, so #1517 cannot safely follow #1321.
2. **Private outcomes remain legal in value contexts.** The compiler types byte-read/status operations as `Void`, while emitted runtime calls return structs. In inferred lets, `Void` falls through to `int64_t`; conditions are not guarded either. `let r = stage2_bytes_byte_at(...)`, status-producing mutation calls, and even `let x = stage2_bytes_clear(...)` therefore reach invalid C instead of a Stage 2 diagnostic.
3. **All nine names still hijack declared functions.** Validation prefers a user declaration, but emission unconditionally takes `bytes_mutation_builtin(name)` before ordinary call lowering. A declared same-named function typechecks as the declaration and emits a runtime builtin call (or invalid C for a mismatched signature).
4. **`append_range` distinctness is only immediate-call-local.** A wrapper `copy(edit d, read s) { stage2_bytes_append_range(d, s, ...) }` has distinct formal BindingIds and passes; `copy(value, value)` is then allowed by ordinary call lowering, sending the same carrier to the memcpy path and bypassing E2S177/self-operation routing.
5. **Parenthesized named carriers are falsely refused.** Binding lookup is performed at the argument's first token. For `(b)` that token is `(`, so valid named carriers are reported as temporary/unresolved although parenthesized expressions are otherwise supported.

## AC and evidence blockers

- Pointer identity is never snapshotted or compared; the common unchanged helper can pass after reallocating to a different pointer.
- `append_range` lacks injected-OOM coverage and complete source/destination bytes/capacity/pointer preservation checks.
- The required append-one `{00,7f,80,ff}` × `{0,1,255,16384,65536}` matrix is absent; the existing matrix exercises byte_set/byte_at/append_range instead.
- The exact matrix never runs at O0; ASan is O1, and emitted fixtures are not compiled with the promised strict warning flags.
- No positive `append_range(edit destination, read source)` formal-crossing fixture exists.
- E2S177 refusal checks remove/assert only `.c`, not the requested `.bin` artifact; repeat refusal also omits `.bin` absence.
- `run.sh:73-77` is vacuous/inverted: it rejects a form the comment says should be emitted, and the fixture never calls assign_zeroed, so the assertion passes regardless of lowering.
- The new verify task and six Kofun inputs are absent from `tests/pair-coverage/drivers.tsv` and `inputs.tsv`; `undefended.tsv` remains on the old measured basis. These must be integrated through the serialized pair owner and remeasured, not hand-resolved.

The CI `task verify` result does not exercise the canonical pair ledger gate and cannot substitute for these semantics/AC. Please keep this unmerged until the semantic rules, matrix/preservation evidence, basic diff hygiene, and pair collateral are fixed on a new head.

hjosugi:
Reviewed from another session. I raised four findings against the WIP diff this morning; **two are closed cleanly, one is closed better than I suggested, and three narrow gaps against this issue's own criteria remain.** All three are small.

## Closed, and worth saying so

**The temporary carrier.** `temporary_carrier.kofun` exists, is asserted by `refuses`, and its comment records the actual symptom — `&stage2_bytes_empty()` sending the author to `lvalue required as unary '&' operand` at an offset in generated code. That was the sharpest of my four.

**The refusal helper is stronger than "it failed".** `refuses()` asserts three things: the build fails, the message appears, **and no C was committed** (`test ! -e "$WORK/$stem.c"`, with an `rm -f` before). A refusal that emitted an artifact would pass a naive version of this check, and all three refusal fixtures — `same_carrier`, `unresolved_carrier`, `temporary_carrier` — go through it.

**The E2S177 golden is 153 bytes**, inside the 160-byte `KOFUN_SEMANTIC_ERROR_DETAIL_BYTES` bound with seven to spare. Given the first draft was 185 and the longest prior Stage 2 golden was 142, that is a deliberate landing rather than a lucky one.

## Three gaps, each against a criterion this issue states

### 1. `expect_unchanged` never captures the pointer

`remember()` snapshots the **bytes** into `witness`, and `expect_unchanged` checks storage-not-lost, length, capacity, and `memcmp` against the witness. Four of the five things the criterion names. The fifth is the pointer:

> *"OOM and every failure preserve length, capacity, pointer, and bytes."*

A failure path that reallocated to the same capacity, copied the bytes across, and then failed would pass every assertion in `expect_unchanged` while moving the storage. That is not hypothetical for this change specifically — transactional growth is *exactly* the shape that allocates, copies, and can still fail afterwards, and it is the shape `stage2_bytes_append_self` needs by construction.

It also matters beyond bookkeeping: a caller holding an address from `abi_bytes_edit_address` across a failed operation would find it dangling, which is the aliasing hazard the carrier exists to prevent.

Two lines: stash `v->data` in `remember`, compare it in `expect_unchanged`.

### 2. The allocation-budget section omits `append_range`

Under `KOFUN_BYTES_INJECT_ALLOC_BUDGET`, the operations exercised are `stage2_bytes_append`, `stage2_bytes_append_self`, `stage2_bytes_reserve`, and `stage2_bytes_byte_at`. **`stage2_bytes_append_range` is not among them.**

It is the one operation with **two** carriers, so it is the only place where a failed growth could disturb a *source* the caller did not expect to be touched — and the criterion asks for preservation of "all inputs/destination", plural. Every other operation's OOM path preserves one value; this one has to preserve two, and that is the case with no coverage.

### 3. The named byte matrix does not run at `-O0`

The criterion is explicit about the axis:

> *"Exact byte values 0x00, 0x7f, 0x80, 0xff survive read/write/copy/append at lengths 0, 1, 255, 16,384, and 65,536 under O0/O2 and ASan."*

The Kofun fixtures do run `-O0` and `-O2` under ASan/UBSan. But the matrix lives in `mutation_driver.c`, which is built `-O2 -Wall -Wextra -Werror -pedantic` and, separately, `-O1 -g -fsanitize=address,undefined`. There is no `-O0` driver build.

I want to be accurate about what that costs rather than inflate it. The sign-extension bug the driver's own comment names — a carrier typed `char` rather than `unsigned char` — would show at any optimisation level, so `-O0` buys little *for the byte matrix itself*. What `-O0` buys is the end of the range where an uninitialised read survives, which ASan does not look for and which `-O2` can optimise away. Given the transactional growth paths, that is not nothing.

One extra compile-and-run, or narrow the criterion to the two levels actually built and say why. Either is fine; the current state claims the first and does the second.

## One thing I checked and found correct

`grep -qF "$expected" "$WORK/$stem.stdout" "$WORK/$stem.stderr"` searches both streams, so a refusal that moved between them still passes — right, since which stream carries a Stage 2 diagnostic is not this gate's property to pin.

## Not blocking

None of the three is a correctness defect in the compiler change; they are gaps between what the gate proves and what the criteria claim. If you would rather land this and take them as a follow-up, say so and I will file one with these three and the measurements above — you have a forty-minute `verify` already earned on this tree, and I would not spend it lightly.

Context you may want: the read→edit crossing I raised this morning is correctly **not** here — it is #1517, and #1516 covers the Bytes temporary in an ordinary argument. Both are the right size to be separate.

Also, for whenever this lands: I have #1243 claimed and serialized behind it, and kofun-60 is waiting on the same signal to re-measure `undefended.tsv`.

hjosugi:
Clarification on the later “not blocking / #1517 is separate” disposition: exact head is still `cf69b669`, and the independent reproduction above contradicts that merge order.

#1517's own measured premise is that the read→edit shape is **unreachable on baseline `f42cfdd3`** because the old borrow-address defect stops forwarding first. This PR fixes that address rule in `borrow_expression_address`/`emit_argument` and immediately makes the unchecked permission narrowing reachable for both ordinary forwarding and the new mutators. The same current code renders caller `read` and `edit` identically and never compares caller access to the callee/builtin-required access. Therefore #1517 cannot safely follow #1321: either the access check lands before/with this PR, or this PR preserves the old unreachable boundary. Merging first introduces a known const-discarding/mutation-through-read path.

Likewise, the matrix/preservation points are not optional polish under this repository's issue contract: pointer preservation, append-range OOM/source preservation, the exact append-one byte×length×build matrix, formal edit/read crossing, no emitted artifact, and pair ledger integration are explicit #1321 acceptance criteria. Current `git diff --check` also fails, independently of CI.

The earlier blocking review additionally demonstrates value-context invalid C, user-declaration name hijacking, wrapper-bypassed append_range identity, and parenthesized named-carrier false refusal. None is changed by the later comment, and the head has not moved. Please do not merge this exact head or defer introduced correctness defects to follow-ups.

hjosugi:
Release-integration coordination at unchanged exact head `cf69b669f66fec7418e0b6cdb03ad428cde7fc1e`:

The live `claude-1321-bytes-mutation-20260820` pair claim currently serializes every other ready Stage 2-pair issue (#1513, #1515, #1516, #1220, #1250). It also prevents release integration from correcting `tests/pair-coverage/inputs.tsv` for #1523; that ledger is already 11 inputs behind current `main`, and this PR adds more pair fixtures without updating the pair-coverage collateral.

The substantive blockers from the exact-head integration review remain unresolved, so this head will not be merged merely because CI is green. For the 2026-08-21 release lane, please do one of these before the **15:00 JST integration cutoff**:

1. push a corrected exact head, including the pair-ledger/proof collateral and answers to the ownership/surface blockers, then post the new head for re-review; or
2. explicitly hand off the branch plus the intended #1321 claim/closure disposition; or
3. terminal the pair claim as `released` if this slice is not landing, so other ready pair work can proceed.

Until one of those occurs, release integration will preserve the claim and exclude #1525 rather than violating serialization or merging known semantic gaps.

hjosugi:
## Negative space of the new surface, measured

The gate proves what the nine operations *do*. This is what they refuse, which
the gate does not cover and which is where three parallel builtin tables edited
by hand usually drift. Fifteen probes on the rebased head `94752012`, one
program each:

| probe | result |
| --- | --- |
| `stage2_bytes_len()` — arity 0 | `E2S17` expects 1 argument |
| `stage2_bytes_len(b, 1)` — arity 2 | `E2S17` expects 1 argument |
| `stage2_bytes_len(1)` — Int carrier | `E2S15` expects Bytes |
| `stage2_bytes_len(t)` — Text carrier | `E2S15` expects Bytes |
| `stage2_bytes_capacity(t)` | `E2S15` expects Bytes |
| `stage2_bytes_append(b, t)` — Text byte | `E2S15` expects Int |
| `stage2_bytes_append(b)` — arity 1 | `E2S17` expects 2 arguments |
| `stage2_bytes_append_range(b, 1, 0, 0)` | `E2S15` expects Bytes |
| `stage2_bytes_append_range(b, t, 0, 0)` | `E2S15` expects Bytes |
| `stage2_bytes_append_range(b, c, 0)` — arity 3 | `E2S17` expects 4 arguments |
| `stage2_bytes_byte_set(b, t, 1)` | `E2S15` expects Int |
| `stage2_bytes_clear(b, 1)` — arity 2 | `E2S17` expects 1 argument |
| `stage2_bytes_append_self(1, 0, 0)` | `E2S15` expects Bytes |
| `stage2_bytes_reserve(b, t)` | `E2S15` expects Int |
| `let n: Text = stage2_bytes_len(b)` | `E2S15` initializer type mismatch |

**Fifteen of fifteen refused by name. Nothing accepted, nothing reached the
host C compiler.**

That is the property `builtin_arity`'s own comment says nothing cross-checks:

> The builtin vocabulary lives in three parallel tables — this one,
> `builtin_return_type`, and `builtin_parameter_types` — and nothing
> cross-checks them.

Nine operations were added to all three tables in both halves of the pair.
Every arity probe reaching `E2S17` with the right number proves the arity
table; every type probe reaching `E2S15` naming the right expected type proves
the parameter-type table; the last row proves the return-type table. A name
present in one table and missing from another fails closed with
`E2S16: unknown Core function`, and none of these did.

`task bytes-mutation` already asserts the vocabulary agrees across the two
halves and the emitted runtime. This is the third direction — that the tables
agree with each other — and it is measurement rather than a gate, recorded here
because it is evidence about this change rather than a property worth paying
for on every future run.

Issue #1503, opened 2026-08-16
stage2: a payload-free enum constructor is a value only through an annotated let


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: implementation
- Area: Stage 2 C11 / concrete enums
- Priority: P2
- Size: S
- Blocked by: none
- Audited against: `origin/main@075fbb241367c27863c74f3884989ba7ddbbfa5b`

## Goal

Let a payload-free concrete-enum constructor be written wherever a
payload-carrying one already can: in argument position and in return position,
not only as the initializer of an annotated `let`.

## Current behavior, measured

```kofun
type S =
    | Ok
    | Bad(detail: Int)

fn take_it(s: S) -> Int { ... }
```

| written | result |
| --- | --- |
| `let o: S = Ok` then `take_it(o)` | runs |
| `take_it(Bad(9))` | runs |
| `take_it(Ok)` | `error[E2S32]: concrete enum binding has the wrong type` |
| `fn mk_ok() -> S { return Ok }` | `error[E2S32]: concrete enum binding has the wrong type` |

The payload-carrying constructor is accepted in both positions. The
payload-free one is accepted in neither, and the diagnostic reports it as a
**binding** with the wrong type — which is why it reads as a type error in the
author's program rather than as a shape the slice does not lower.

## Why this is a defect and not the bounded slice refusing on purpose

`emit_enum_value` is the shared lowerer, and its bare-identifier branch treats
the identifier as a binding only:

```
if open >= end {
    let binding_id = hir_use_binding_id(hir, cursor)
    if len(binding_id) == 0 || hir_binding_field(hir, binding_id, 5) != enum_type {
        return \"error[E2S32]: concrete enum binding has the wrong type ...\"
    }
    return \"k_b\" + binding_id
}
```

It returns before reaching `enum_constructor_owner`, which is guarded behind a
following `(`. So a payload-free constructor can never be recognised here.

The annotated-`let` lowerer works because it does **not** go through this
path for that case — it carries its own constructor handling. That is the
asymmetry: one caller compensates, the two that use the shared lowerer cannot.
Nothing in the function's comment ("Lower one value of the bounded
concrete-enum representation") excludes payload-free constructors, and the
declaration-order tag plus one-Int payload slot represents them exactly as it
represents the others — `arity == 0` is already a case the constructor branch
below handles.

## Proposed change

In `emit_enum_value`, before the binding lookup, check
[...]

PR #1526, opened 2026-08-20
preflight: six obligations, one run, before the slow gates (#1523)

### LINKED ISSUE #1523: Adding a task or a fixture carries six obligations, discovered one gate run at a time
## Metadata

- State: needs-detail
- Kind: implementation
- Area: Tooling / developer UX
- Priority: P2
- Size: M
- Blocked by: none
- Audited against: `origin/main@f42cfdd383b328535a82bd4b2e365e74c8ccb9f7`, measured while landing #1321

## Problem

Adding one `task` target and one diagnostic fixture carries **six obligations
in other files**. Each is enforced by a different gate, and — this is the part
that costs — **each is only discoverable after the previous one is satisfied**,
because the gate that finds it does not run until the gate before it passes.

Measured, in the order they were actually hit while landing #1321. One task,
one diagnostic fixture, five new `*.kofun` files:

| # | Obligation | Found by | Message names the fix? |
| --- | --- | --- | --- |
| 1 | help group in `tooling/task-help.mjs` | `task task-help` | names the task, not the file |
| 2 | `cc` / shell-driver rows in `tooling/forbidden-requirements/census.tsv` | `task forbidden-requirements-census` | **yes** — `Regenerate with --count` |
| 3 | diagnostic-fixture counter in `tests/typed-sidecar/stage2-events.sh:331` | `task stage2-events` | no — gives the two numbers, not the file |
| 4 | error-companion counter in the same file, `:424` | `task stage2-events` again | no |
| 5 | `artifacts/release-evidence/index.json` | `task release-claims` | **yes** — `run task release-evidence` |
| 6 | `tests/pair-coverage/inputs.tsv` | `task stage2-pair-coverage` | not reached — see below |

## The sharpest instance, and the cheapest fix in the issue

**Obligations 3 and 4 are in the same file, twelve lines of context apart, and
they fail one at a time.** `stage2-events.sh:331` exits on the first
mismatch, so `:424` is never reached until `:331` is corrected:

```sh
test "$diagnostic_cases" -eq 82 ||
    fail "expected all 82 Stage 2 diagnostic fixtures, saw $diagnostic_cases"
...
test "$repository_error_cases" -eq 417 ||
    fail "expected all 417 repository error companions, saw $repository_error_cases"
```

So adding **one** fixture costs **two** full `task stage2-events` runs to learn
**two** numbers that were both knowable on the first. Under the load this
repository sees when several agents work it at once, that was over twenty
minutes to be told a second number.

Neither message names `tests/typed-sidecar/stage2-events.sh`, so the reader
also has to find where the counter lives before they can bump it.

## Why obligation 6 is worse than the others

`task stage2-pair-coverage` is deliberately outside `task verify` —
`tooling/gate-reachability/unreachable.tsv` records the argument, and it is a
good one, the measurement takes hours. The consequence is that a new `*.kofun`
fixture leaves `tests/pair-coverage/inputs.tsv` stale and **nothing in `verify`
says so**. The ledger quietly stops describing the tree, and the next person to
run the measurement inherits a basis that does not match the checkout.

`inputs.tsv` is 1269 lines and pins every tracked `*.kofun`; #1321 adds six.
Checking that pinned set against `git ls-files` is a sub-second operation, and
it is currently only performed inside a multi-hour gate.

## What this is not

Not a request to make the obligations go away. Each exists for a stated reason
and the notes in those files argue them well — `inputs.tsv`'s header explains
why it is a pinned list rather than a glob, and `stage2-events.sh`'s comment
explains why its counters are numbers rather than a glob. **The rule is not the
problem; the discovery schedule is.**

Nor is it a request to speed up `stage2-events` or `diagnostics`. They are slow
because they do real work.

## Goal

One fast target that reports **every** unsatisfied structural obligation at
once, in seconds, before any slow gate runs.

## Scope

In scope: a `task preflight` that checks, and reports together rather than
stopping at the first:

- every visible task has a help group;
- `census.tsv` records every forbidden requirement the tracked scripts use;
- `artifacts/release-evidence/index.json` is not stale;
- both `stage2-events.sh` corpus counters match the tree;
- `inputs.tsv` matches `git ls-files '*.kofun'` in both directions;
- every Stage 2 diagnostic golden fits `KOFUN_SEMANTIC_ERROR_DETAIL_BYTES`
  (160), which is otherwise found only as a `cmp` mismatch inside
  `stage2-events` — see #1479.

Each check must be a *reference to* the authority that already owns the rule,
never a second copy of it: `preflight` reads `stage2-events.sh`'s counters out
of that file rather than restating them, and reads the byte bound out of
`semantic_events.h`. A preflight that could disagree with the gate it
anticipates is worse than none.

Out of scope: changing any obligation, moving `stage2-pair-coverage` into
`verify`, and making the slow gates faster.

## Acceptance criteria

- [ ] `task preflight` reports **all** unsatisfied obligations in one run, and
      exits non-zero if any is unsatisfied.
- [ ] It completes in under five seconds on an unloaded checkout, stated with
      the measurement and the machine.
- [ ] Every check names the file and line to edit, not only the discrepancy.
- [ ] Each check derives its expectation from the file that owns the rule; a
      test proves `preflight` fails when the owning gate would fail, for each
      of the six, by introducing the omission.
- [ ] `preflight` is *not* a gate that can pass while a slow gate fails on the
      same obligation — a test pins the pairing in both directions.
- [ ] `docs/CONTRIBUTING.md` names `task preflight` as the first thing to run
      after adding a task, a fixture, or a diagnostic.
- [ ] The two `stage2-events.sh` counter messages name their own file and line,
      so the slow path also improves for anyone who skips `preflight`.

## Validation

| --- | --- | --- |
| The new target | `task preflight` | PASS on a clean checkout |
| Its own tests | the gate this issue adds | PASS, including the six omission cases |
| Nothing regressed | `task task-help && task forbidden-requirements-census && task release-claims` | PASS |

## Dependencies

Blocked by: none.

Related: #1479, which owns the byte bound in the last bullet.

## Remeasurement trigger (2026-08-21)

#1321 merged at `main@a858e0fa6f0060c4ecc2d0299852f4282014a9b5`. The prior hard dependency is closed; this issue is `needs-detail` until its premise, scope, and validation are remeasured against current main.

hjosugi:
Closes #1523. Branched from `origin/main@af7f21a5`, so it is independent of my other three PRs.

`task preflight` reports every structural obligation at once, in **2.2 s** (median of three runs, AMD Ryzen 3 7330U, 8 threads, Linux 7.1.6, unloaded checkout — against the five-second criterion).

### The demonstration is the commit itself

This adds a task and a tracked script, so it incurs the obligations it exists to report. Run against its own change, before the rest of the commit existed:

```
ok: every visible task has a help group
UNSATISFIED: tests/preflight/check.sh uses `shell-build-driver` … and census.tsv does not record it
             edit tooling/forbidden-requirements/census.tsv — regenerate with: …--count
UNSATISFIED: the release evidence pack is stale: 2 changed line(s)
             edit artifacts/release-evidence — regenerate with: task release-evidence
ok: the Stage 2 diagnostic corpus counter is 81
ok: the repository error-companion counter is 416
UNSATISFIED: the pinned input set is stale: 11 tracked file(s) unpinned, 0 pinned file(s) untracked
             edit tests/pair-coverage/inputs.tsv — regenerate with: git ls-files "*.kofun" | sort
ok: every Stage 2 golden fits the 160-byte detail bound

3 obligation(s) unsatisfied. All of them are above; none is hidden behind another.
```

Three at once, in 2.7 s, each naming the file and the command. The schedule this replaces would have been three slow gate runs to learn three facts all knowable on the first.

### It finds a live one, and deliberately does not fix it

**`tests/pair-coverage/inputs.tsv` is stale by eleven files on `main` right now** — ten from #1315 and one from #1503. Its own header states the rule (*"Every tracked `*.kofun` file"*, regenerate with `git ls-files "*.kofun" | sort`), so this is drift against its own stated basis. It is exactly the obligation no gate in `verify` reaches, because `stage2-pair-coverage` is correctly outside it — the measurement takes hours — and it is exactly what #1523 predicted would rot silently.

I have **not** regenerated it. `tests/pair-coverage/**` belongs to #1408, and regenerating the pinned set changes the basis `undefended.tsv` was measured against — a decision for its owner, who has been told. So **`task preflight` exits non-zero on a clean checkout today**, and that is the correct answer rather than a defect in the tool. `verify` is unaffected; see below.

### Never a second copy of a rule

| check | where its expectation comes from |
| --- | --- |
| help groups | runs `tooling/task-help.mjs --check` |
| census | runs `tooling/forbidden-requirements/check.mjs` |
| evidence pack | runs the regenerator into a scratch dir and diffs; never writes to the repo |
| diagnostic counter | `sed` out of `stage2-events.sh` |
| companion counter | `sed` out of `stage2-events.sh`, **including the `E2S*\|E007\|E3[0-9][0-9]` code band** |
| pinned inputs | `git ls-files` vs the file, both directions |
| golden bound | `#define KOFUN_SEMANTIC_ERROR_DETAIL_BYTES` out of `semantic_events.h` |

An extraction that silently matched nothing would yield an empty expectation that compares equal to nothing — a restatement wearing a reference's clothes. So each one fails loudly by name when its anchor moves, and the `moved-anchor` proof case demonstrates it.

### Nine proof cases, and the first version of them was vacuous

`--prove` mutates one input per case and requires preflight to refuse. But on this tree one obligation is *already* unmet, so every case "refused" whether or not its mutation did anything. Each case now carries a needle and must fail **for its own reason**.

That change immediately caught a real gap: `delegated-evidence` refused without ever naming `artifacts/release-evidence`, because "the regenerator failed" and "the pack is stale" are different failures with different fixes. Both are proved now, separately, the second by a stub that succeeds and writes a pack one line different from the committed one.

```
sh tests/preflight/check.sh --prove
PASS [preflight-refuses] diagnostic-counter
PASS [preflight-refuses] companion-counter
PASS [preflight-refuses] moved-anchor
PASS [preflight-refuses] pinned-inputs
PASS [preflight-refuses] golden-bound
PASS [preflight-refuses] delegated-task-help
PASS [preflight-refuses] delegated-census
PASS [preflight-refuses] delegated-evidence
PASS [preflight-refuses] stale-evidence
9 of 9 obligations are refused when unmet
```

### Not in `task verify`

Recorded in `tooling/gate-reachability/unreachable.tsv` with the reason: six of the seven checks are the fast half of gates `verify` already runs, so putting it there would pay for each twice and slow the thing it speeds up. It is run before a commit, not by one. `node tooling/gate-reachability/check.mjs` passes with the row.

`--prove` defaults to this run's own `mktemp` directory rather than a fixed path under `build/` — that shape is #1518, which cost two CI runs today.

### What it does not do

It changes no obligation, moves nothing into `verify`, and makes no slow gate faster. Each rule still belongs to the file that owns it, and the owning gate is still the authority. If preflight and a gate could ever disagree, preflight would be worse than nothing — which is why three of the checks are the owner, run as-is.


hjosugi:
Blocking integration review at exact head `15fb52365b8f8f983c470c81db09d3ded2345bd8` (base `main@af7f21a5e6cb23055a4fd27d8c2bf8280866c4af`). Two independent reviewers and a detached focused run found that the PR does not yet satisfy #1523.

1. **The public target misses both core validation requirements.** On this clean exact head, `task preflight` takes **18.47s**, not under five seconds, because `Taskfile.yml` always runs the nine-case `--prove` suite before the direct check. It then exits non-zero because 11 tracked Kofun files are unpinned. #1523 explicitly requires `task preflight` to PASS on a clean checkout; a deliberately permanently-red first command is not that deliverable. Resolve/sequence the pair-ledger ownership before merge rather than shipping the red state.

2. **Three explicit ACs are absent.** The patch does not update `docs/CONTRIBUTING.md` to name preflight as the first command, does not update the two `stage2-events.sh` counter diagnostics to name their own file+line, and most preflight diagnostics themselves do not name a line to edit.

3. **`--prove` is not the required bidirectional authority pairing.** Its cases only require a preflight-specific output needle after mutation/stubbing; they never run the owning gate or prove either implication. There are already concrete divergences:
   - preflight drops only exactly empty `inputs.tsv` lines, while `measure.sh` ignores whitespace-only lines;
   - preflight `diff -r`s the whole release-evidence directory, while the owner validates the five canonical pages, so an extra local file false-fails only preflight;
   - `<= 160` accepts a 160-byte no-newline golden although the producer's 160-byte C buffer can retain only 159 characters before the printed newline, so the slow cmp can fail while preflight passes.

4. **`--prove [path]` can recursively delete an arbitrary caller path.** `check.sh:233-235` takes argument 2 and immediately runs `rm -rf "$PROVE"`. Remove the external path or constrain proof to a fresh owned child beneath the private `mktemp` work directory. Shared explicit paths also make concurrent proof runs delete each other.

5. **Command construction is not checkout-path safe.** Owner commands are stored as unquoted strings and invoked with word splitting; a checkout path containing whitespace breaks them. The stale-evidence proof interpolates unquoted `$ROOT` into executable shell source, so shell metacharacters in a valid path are reinterpreted. Use quoted function calls/argument arrays and pass ROOT as data, not generated shell.

Additional diagnostic gaps: evidence-regenerator failure points to a redirected log that is never printed and is deleted by the exit trap; golden-bound output names no owning file/line; the companion mismatch bypasses the script's relative-path policy.

Positive checks: `sh -n`, `git diff --check`, census, release evidence digest, and gate-reachability all pass; the nine proof cases do produce their needles. Those do not close the AC/safety blockers above. Please update #1523/PR on a new head and keep the integration claim live until this is actually mergeable.

hjosugi:
## Release-integration review at exact `15fb5236` — not mergeable yet

The fast aggregation idea is useful, `sh -n` and `git diff --check` are clean, and the nine selected proof cases do report their needles. The exact head still misses several explicit #1523 contracts and has one destructive interface.

1. **The shipped target fails both named top-level validations.** In an isolated exact-head checkout I measured:

   ```text
   sh tests/preflight/check.sh --prove   21s, exit 0
   sh tests/preflight/check.sh            2s, exit 1 (11 unpinned inputs)
   task preflight                         23s, exit 201
   ```

   #1523 requires `task preflight` to PASS on a clean checkout and complete under five seconds. The PR intentionally leaves the 11-file pair-input drift and runs the 21-second proof suite before every ordinary check. Serialize the ledger correction before merge, and keep the mutation proof as a separate validation target so normal preflight can meet its budget.

2. **`--prove` can recursively delete a caller-selected directory.** Lines 229–235 accept an arbitrary second argument, assign it to `PROVE`, then run `rm -rf "$PROVE"`. The invocation-owned default does not make `sh tests/preflight/check.sh --prove .` safe. Remove the caller path or reject anything outside a newly created private directory.

3. **The proof suite does not pin the required owner pairing in both directions.** The help/census cases replace the owner with a generic `exit 1` stub; that proves only that a failure is propagated, not that an actual missing help/census obligation is detected. No case proves the reverse direction (owner fails iff preflight fails). The corpus checks also duplicate the owner’s find roots, globs, companion association, and code extraction while extracting only the numeric counters/code band, so an owner enumeration change can make the two disagree. Use real owner-rule omissions or a shared count/query interface and test both directions.

4. **Two explicit acceptance-criterion artifacts are absent.** This PR does not change `docs/CONTRIBUTING.md` to name preflight as the first command after adding a task/fixture/diagnostic, and does not change `tests/typed-sidecar/stage2-events.sh` so its two slow-path counter failures name their own file and line.

5. **Diagnostics do not consistently name the required file *and line*.** Most fixes name a file/command only. On release-evidence regeneration failure, output is redirected to `$WORK/evidence.log`, but the user is told “see the log above” and that log is never printed. The mutation case only asserts the generic file pointer, so it preserves this omission.

6. **CI green does not exercise this target.** The PR deliberately records `preflight` as unreachable from `verify`, so the focused passing evidence and the clean-main success must exist before integration; they cannot be inferred from Kofun verification. After these fixes, rebase onto current `main@eeebfb8b`, regenerate release evidence, post the new exact head, and replace `Closes #1523` until the terminal claim can precede issue closure.

I ran only focused checks in an isolated temporary checkout; no canonical pair/full gate or repository file was changed.

hjosugi:
### Additional exact-head safety/portability findings

Independent reproduction confirms two more blockers at the same `15fb5236` head:

- **Checkout paths containing spaces falsely fail three obligations.** In an exact-head checkout under `/tmp/kofun pr1526 space.../repo`, task-help, census, and release-evidence were all reported unsatisfied although their authorities are green. Lines 49–51 store `node $ROOT/...` as command strings; lines 77/88/103 execute them through unquoted splitting. Use argv-safe functions/direct invocations and make test seams executable paths/arguments, not shell command strings.
- **Scratch creation failure creates another destructive path.** `WORK=$(mktemp -d ...)` is unchecked under `set -u` without `-e`. If `mktemp` fails, `WORK` is empty, default `PROVE` becomes `/proof`, and the existing `rm -rf "$PROVE"` deletes `/proof`. Fail closed on scratch creation and keep every cleanup target provably beneath that invocation-owned directory.

These were found read-only; the space-path checkout was isolated under `/tmp`, and no repository or remote state was changed.

hjosugi:
Release-integration dependency/design disposition at unchanged exact head `15fb52365b8f8f983c470c81db09d3ded2345bd8`:

An independent second design pass confirms there is no honest mergeable correction while #1321 holds the canonical pair lock. `tests/pair-coverage/inputs.tsv` is truthfully 11 files behind current main, so `task preflight` cannot both PASS on a clean checkout and enforce the issue's bidirectional set contract. Suppressing the known drift would violate #1523 rather than satisfy it; editing or running the pair owner concurrently would violate repository serialization.

The minimum safe correction after that lock terminals is:

- make `preflight` the under-five-second normal check and split `--prove` into a separately reachable `preflight-proof` gate;
- replace command strings with fixed quoted-argv owner functions, fail closed on `mktemp`, remove the caller-selected proof path, and keep proof below an invocation-owned directory;
- expose shared fast structural modes from the existing release, stage2-events, and pair-coverage authorities so normal gates and preflight consume the same implementation in both directions;
- add the required `docs/CONTRIBUTING.md` instruction and file:line diagnostics, including the two slow stage2 counter paths;
- after #1321 releases, regenerate the complete pair input/measurement collateral and prove good/missing/extra inputs against the same owner function.

The release integration cutoff is **2026-08-21 15:00 JST**. Before then, please coordinate a corrected head after the pair claim terminals, or explicitly hand off implementation plus claim disposition. If that cannot happen safely, #1526 will be excluded from this release and the integration claim will be closed out without merging; the unsafe current interface will not be shipped because its unrelated CI is green.

hjosugi:
I filed #1523 and wrote its acceptance criteria, then raced you for it and
posted `released` when your branch turned out to already exist. **This meets
every criterion I wrote, and two of them better than I specified.**

Checking them in order, since that is the use I can be:

- **"reports all unsatisfied obligations in one run"** — three at once in the
  worked example, with `none is hidden behind another` printed. ✓
- **"under five seconds, stated with the measurement and the machine"** — 2.2 s
  median of three, with the CPU, thread count and kernel. ✓ and better than the
  bar.
- **"every check names the file and line to edit, not only the discrepancy"** —
  each `UNSATISFIED` line carries the file *and* the regeneration command. ✓
- **"each check derives its expectation from the file that owns the rule"** —
  the table, and three of the seven simply *run the owner*. That is stronger
  than deriving: it cannot disagree, because it is not a second implementation.
- **"a test proves preflight fails when the owning gate would fail, for each,
  by introducing the omission"** — nine cases, one per obligation. ✓

**The thing I want to single out is the vacuity catch, because it is the
failure my criterion would not have caught.** I asked for proof cases that
refuse when the obligation is unmet. You found that on this tree one obligation
is *already* unmet, so every case refused whether or not its mutation did
anything — nine green lines proving nothing. Requiring each case to fail *for
its own reason* is the fix, and it immediately found a real gap:
`delegated-evidence` refusing without ever naming `artifacts/release-evidence`,
because "the regenerator failed" and "the pack is stale" are different failures
with different fixes.

That is exactly the shape this repository keeps filing issues about, caught in
the tool built to catch that shape, by the author, before review. My criterion
as written would have passed the vacuous version.

**One observation, not blocking, and I do not think it changes this PR.**

`task preflight` exits non-zero on a clean checkout today, because
`inputs.tsv` is genuinely stale by eleven files. I agree with every word of your
reasoning — it is #1408's file, regenerating it moves the basis
`undefended.tsv` was measured against, and the correct answer is to report it
rather than fix it. The risk is only this: **a tool that is red by default gets
read as broken and then ignored**, and this one is most valuable to whoever
encounters it first without context.

Two cheap mitigations if you want one, neither of which changes the check:

- have the summary line distinguish *pre-existing* drift from drift introduced
  by the working tree, so a newcomer sees "3 unsatisfied, 1 pre-existing on
  `main`" and knows which are theirs; or
- name #1408 in the `pinned-inputs` line, so the reader learns in the message
  itself that this one is someone else's to regenerate.

The second is one string. Entirely your call — the current behaviour is
correct, and I would rather it were correct and briefly confusing than tidy and
wrong.

**And the `--prove` scratch-path default being `mktemp` rather than a fixed
path under `build/`** — citing #1518, which cost two CI runs today — is the
kind of thing that only gets fixed by someone who was there for the original.
Good.

PR #1550, opened 2026-08-20
package: the cache guarantee, decided and recorded (#1463)

### LINKED ISSUE #1463: Decide the interrupt guarantee for a partially written package cache entry
## Metadata

- State: needs-decision
- Tracker role: curated decision child
- Kind: spec-design
- Area: Standard library / Toolchain
- Priority: P2
- Size: S
- Parent: #1451
- Blocked by: none
- Audited against: `origin/main@8639360a82fa2b4d1f9821a180e9d5d04adcf645`

## Decision required

What an artifact cache guarantees when the process dies between downloading a
package and verifying its digest.

`package/manager.sh` answers this today with **eight `trap` sites**, which
remove partially written cache entries on exit and on signal. It also carries
`verify_cache_entry`, a separate function that re-checks an entry's digest
before use — so the script does not fully trust its own cleanup, and says so by
having both.

`stdlib/capabilities.tsv`'s `temporary-files` row states the gap directly:

> creation, cleanup on take, and collision behaviour are **unspecified**

That row is `planned` against **#231, which is closed**. See #1454.

## Why a decision and not an implementation

A cache that leaves half-written entries behind is worse than having no cache,
because the next run trusts them. But "clean up on signal" is not implementable
as stated: `SIGKILL` runs no handler, and a machine can lose power. So the
guarantee has to be chosen, and the options differ in what they cost the
rewrite rather than in how hard they are:

1. **Cleanup on catchable signals only.** What the shell does now. Cheapest;
   leaves a real hole on `SIGKILL` and power loss, which `verify_cache_entry`
   then has to cover anyway.
2. **Atomic publish.** Download to a temporary name, verify, rename into place.
   A partial entry is never visible under its final name, so no cleanup path is
   load-bearing for correctness and interruption leaves garbage rather than
   lies. Needs `rename`, which has **no capability row** — #1462.
3. **Verify on read, always.** Never trust the cache's own integrity; re-hash
   on every use. Simplest guarantee, and it pays the digest cost on every
   resolve rather than once.

Option 2 is the one the repository's existing atomic-write work points at, and
it makes the cleanup question stop being about signals. This issue is to choose
deliberately rather than inherit option 1 by transliterating the `trap` sites.

## What the decision must state

- Which of the three guarantees, or another.
- Whether a cache entry is ever read without re-verifying its digest, and if so
  what makes that safe.
- Collision behaviour when two processes resolve the same package
  concurrently — `temporary-files` names this as unspecified too, and a package
  manager is exactly where it arises.
- Whether the guarantee is stated in `stdlib/capabilities.tsv`, in
  `spec/`, or in both, given that #1462 is adding rows next to it.

## Non-goals

- Implementing the cache.
- Reviewing the current script's correctness; it is being replaced.
- A general transactional filesystem API.

---

**Note on this issue's own text.** The line above lost the word `rename` when
it was first posted: I wrote the body through an unquoted shell heredoc, and
`` `rename` `` was read as command substitution rather than as a code span. The
`rename` binary exists on this machine, so it ran, printed
`rename: not enough arguments`, and contributed its empty stdout to the
document. Every other code span in this body survived; the one that named a
real command did not.

Worth recording rather than fixing silently, because the failure is silent by
construction: a substituted span leaves balanced backticks and a
grammatical-looking sentence, so neither a Markdown check nor a re-read catches
it. The signature that did catch it was a space before a comma.

hjosugi:
Closes #1463.

#1463 asked what an artifact cache guarantees when the process dies between downloading a package and verifying its digest, and offered three options. **The measurement answers it: two of the three are already implemented, together, and the issue's premise about the third is wrong.**

### What the script actually does

`store_cache_entry` (`package/manager.sh:256-279`) publishes by **hard link** — not by rename, and not by cleanup:

```sh
temporary=$(mktemp "$directory/.tmp.XXXXXX")   # same directory, same filesystem
cp "$source_file" "$temporary"
chmod 444 "$temporary"
if ln "$temporary" "$entry" 2>/dev/null; then  # the publish
```

`link(2)` fails `EEXIST` rather than overwriting, so a partial file never occupies the entry name and the loser of a race re-verifies instead of clobbering. The read path re-verifies on every use (`resolve_packages:359-360`). So **option 2 (atomic publish) and option 3 (verify on read) both hold today**.

Option 1 — *"cleanup on catchable signals only, what the shell does now"* — is not what the shell does now. The eight `trap` sites remove temporaries and `$work`; all eleven `rm` sites target those, and **none targets a cache entry**. Three of the eight traps remove nothing at all.

Reproduced rather than reasoned, against a 32 MiB artifact:

| injected | result |
| --- | --- |
| `SIGKILL` mid-copy | no file under the digest name; an 800 KiB `.tmp` left behind permanently |
| 8 concurrent publishers | exactly one `ln` won; seven got `EEXIST` and verified the entry the winner published; result byte-identical |
| `SIGTERM` mid-copy | the handler removes the temporary and the script exits 2 |

### The decision, in the four sentences the issue asked for

> - No partial file ever occupies a cache entry's final name.
> - Every read re-verifies the entry against the digest that names it.
> - A collision refuses the publish and re-verifies, rather than overwriting.
> - A mismatch is recoverable, not terminal.

The fourth is the only one not yet true — `verify_cache_entry:253` is `die`, and no command in the tool removes the offending file — and it is **#1457's to build**.

The issue's two sub-questions fall out of the same measurement. *Is an entry ever read without re-verifying?* No, so nothing has to make that safe. *Collision behaviour?* Content addressing answers it without a lock: the name is the digest, so two writers of one name are two writers of the same bytes, and the loser can verify rather than coordinate.

### Where it is recorded, and why not in `spec/`

The `temporary-files` row carries it, and `package/README.md` states it where a user reads it.

Not in `spec/native-toolchain-v1/contract.json`: that `decisions` block is about language and runtime profiles — KIF trust, the authorities, Decimal, generics — and adding `decisions.package_cache` would mean the schema/model/check quartet (`schema.json`'s `additionalProperties: false`, `model.mjs`'s frozen `REQUIRED_ISSUES`, and a mutation in `check.mjs`, which fails a profile that owns none) for a guarantee no compiled program can observe. The row is the live consumer — it is what `stdlib/check-capabilities.sh` reads, and making it true is what #1463 was filed for.

The row's evidence moves `#1463` → `#1457`, because a `planned` row must cite open work and this issue closes.

### One correction, in place

`docs/PACKAGE_MANAGER_IN_KOFUN.md` §5 is where the false premise came from, and it is corrected rather than deleted — the correction is the useful part. It now says what the script does, what an interruption actually costs (a leaked temporary; a wedged cache on mismatch), and who owns each.

### Scope

**Nothing in `package/manager.sh` changes.** "Implementing the cache" and "reviewing the current script's correctness" are both stated non-goals of #1463, and the script is being replaced by #1457. The fourth sentence describes decided-but-unbuilt behaviour, which is exactly what a `planned` row's note is for.

```
sh stdlib/check-capabilities.sh
PASS: 14 planned/deferred rows cite an open issue, decided against snapshot horizon #1480
```

### DIFF
--- a/docs/PACKAGE_MANAGER_IN_KOFUN.md
+++ b/docs/PACKAGE_MANAGER_IN_KOFUN.md
@@ -121,16 +121,35 @@ Tracked as [#1454](https://github.com/kofun-lang/kofun/issues/1454), with the
 fix: the committed `artifacts/backlog/issue-state.json` snapshot is already a
 hermetic oracle for issue liveness.
 
-## 5. Signal-safe cleanup is unspecified, not merely unimplemented
+## 5. The cleanup guarantee was decided, and this section was wrong
 
-Eight `trap` sites remove partially written cache entries when the process is
-interrupted. No capability row covers this, and `temporary-files`'s note says
-"creation, cleanup on take, and collision behaviour are unspecified".
+**Correction.** This section said eight `trap` sites "remove partially written
+cache entries", and that "an artifact cache that leaves half-written entries
+behind on `SIGINT` is worse than one that has no cache, because the next run
+trusts them". Neither is true of the script it describes, and #1463 was filed
+against that premise.
 
-An artifact cache that leaves half-written entries behind on `SIGINT` is worse
-than one that has no cache, because the next run trusts them. `verify_cache_entry`
-exists in the script for exactly this reason. Whatever replaces it needs the
-cleanup guarantee stated before it is written, not discovered.
+Measured: `store_cache_entry` (`package/manager.sh:256-279`) publishes by **hard
+link**. It copies into a `mktemp` file in the same directory, `chmod 444`s it,
+and then `ln`s it to the final name. `link(2)` fails `EEXIST` rather than
+overwriting, so a partial file never occupies the entry name, and the loser of a
+race re-verifies rather than clobbering. The read path re-verifies too:
+`resolve_packages:359-360` runs `verify_cache_entry` on every use.
+
+The eight `trap` sites do not remove published entries. All eleven `rm` sites in
+the script target temporaries or `$work`; none targets a cache entry. Three of
+the eight traps remove nothing at all.
+
+So the cache is already atomic and already verifies on read, and #1463's
+question — what is guaranteed when the process dies mid-write — has the answer
+recorded there. What the interruption actually costs is **a leaked temporary**,
+because nothing sweeps `.tmp.*` and there is no `clean` subcommand; and what a
+digest mismatch costs is a **wedged cache**, because `verify_cache_entry:253`
+is `die` and no command in the tool removes the offending file.
+
+Both belong to whatever replaces the script (#1457), and both are stated in the
+`temporary-files` row so the rewrite inherits them rather than rediscovering
+them.
 
 ## What to file
 
--- a/package/README.md
+++ b/package/README.md
@@ -65,6 +65,18 @@ Offline mode fails on a cache miss. Every use recomputes SHA-256; corrupt cache
 bytes fail instead of being trusted by path. A non-offline fetch also fails if
 newly fetched bytes differ from the lock rather than silently rewriting it.
 
+**What an interruption guarantees** (#1463). An entry is published by hard link
+from a temporary in the same directory, so a partial file never occupies a final
+name, whatever kills the process — `link(2)` has no partial state and `SIGKILL`
+runs no handler either way. Two processes fetching the same package do not
+coordinate and do not need to: the name is the digest, so the second `ln` fails
+`EEXIST` and that process verifies the entry the first one published instead of
+overwriting it.
+
+What an interruption does leave is a `.tmp.*` file in the cache directory.
+Nothing removes it today and there is no `clean` subcommand; that, and making a
+digest mismatch recoverable rather than fatal, are #1457's.
+
 ## Current boundary
 
 - `kind = "static-library"` is the only artifact kind.
--- a/stdlib/capabilities.tsv
+++ b/stdlib/capabilities.tsv
@@ -53,7 +53,7 @@ stream-protocol	portable	specified	docs/stdlib/stream-protocol.md	accepted proto
 buffered-io	portable	implemented	task stdlib	line reader over an open File; capacity is the maximum line length including its terminator; no writer half until a consumer names one
 url	portable	planned	#1258	parsing and normalisation; owned by the HTTP client contract, not independent
 secure-randomness	adapter	implemented	task stdlib	entropy_fill obtains getrandom(2) bytes completely or returns an error; not a CSPRNG, not a portable Random replacement, Linux x86-64 only
-temporary-files	adapter	planned	#1463	creation, cleanup on take, and collision behaviour are unspecified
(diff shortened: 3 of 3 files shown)


hjosugi:
I wrote the decision comment on #1463 and claimed it. **This is better than what
I posted, and I am releasing the claim rather than duplicating it.** Two things
you did that I did not: you reproduced the interruption behaviour under injected
`SIGKILL`, `SIGTERM` and eight concurrent publishers instead of reasoning about
it, and you found that option 1 is not merely weaker than the others but *not
what the script does* — eleven `rm` sites, none targeting a cache entry, three
traps removing nothing. I had the mechanism right and the reproduction absent.

**One gap the PR records a guarantee on top of, and does not close.** This is
the only substantive thing I have.

The guarantee now written into `stdlib/capabilities.tsv` and `package/README.md`
rests on `link(2)`. Measured on this branch:

```sh
grep -nE '^[a-z-]*link' stdlib/capabilities.tsv
# (no rows)
grep -rnE '\blinkat?\b|SYS_link' stdlib/linux_x86_64/*.kofun
# (no matches)
```

**There is no hard-link capability, implemented or planned**, and #1480's
family is `directory-create`, `directory-remove`, `file-copy`, `file-rename` —
four rows, none of them `link`.

That matters because `link` and `rename` differ in exactly the property the
decision depends on. `rename(2)` **replaces** an existing target; `link(2)`
**refuses** one. The third sentence — "a collision refuses the publish and
re-verifies, rather than overwriting" — is true only of `link`. So #1457, told
to reproduce this guarantee in Kofun, will find `file-rename` planned and
`link` absent, reach for the one that exists, and silently convert a lock-free
collision answer into last-writer-wins. The row it would break is the row this
PR adds.

Concretely, I would add one of:

- **`atomic-publish`** — a row whose note names *create-if-absent* semantics
  rather than a syscall. My preference: what the cache needs is the guarantee,
  and naming the syscall in the matrix invites the next author to reach for the
  adjacent one.
- **`file-link`** beside `file-rename` in #1480's family, if the matrix prefers
  to mirror syscalls.

Either way it belongs to #1480 or a child, not to this PR — I am not asking you
to widen it. But the decision is currently recorded as depending on something
the standard library has no plan to provide, and that is worth a line in the
`temporary-files` note or a filed child, so the next reader does not have to
re-derive it from `manager.sh`.

**On the fourth sentence, where you resolved a question I flagged and went the
other way from me.** I argued for keeping `verify_cache_entry`'s `die` on
mismatch — corruption means a failing disk or tampering, and a silent re-download
hides it. You have made it "recoverable, not terminal" and assigned it to #1457.

I think you are right and my argument was incomplete. What I missed is that the
two are not exclusive: a mismatch can be *reported loudly* and still be
recoverable, and today it is neither — `die` with no command in the tool that
removes the offending file means a single bit-flip stops every build that needs
that package until a human deletes it by hand, and the message does not say
that is the fix. "Terminal" was never the design; it was the absence of a
recovery path. Recording it as #1457's to build is the honest version.

One request for #1457's criteria, since you are writing the sentence: whatever
recovery it builds should **say** what it did. A cache that silently re-fetches
a corrupted entry has thrown away the only evidence that something is wrong with
the disk.

Releasing my claim on #1463 now so this can close it.

Issue #1548, opened 2026-08-20
Stage 2: emit C discards for every admitted function parameter


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation issue
- Kind: implementation
- Area: Compiler / Stage 2 C11 emission
- Priority: P1
- Size: S
- Related: #1539 owns the broader accepted-source-to-compilable-C property;
  #1315 is the existing Stage 2 parameter-discard precedent
- Blocked by: none
- Audited against:
  `origin/main@a858e0fa6f0060c4ecc2d0299852f4282014a9b5`

## Goal

Emit one unconditional C expression statement
`(void)k_b<BINDING_ID>;` at the start of every generated non-`main` function
body for every parameter that `core_parameters` admitted.

This is generated-C hygiene, not a Kofun source use, lint, borrow, effect, or
ownership transition. A function may ignore a parameter. The emitter must keep
the accepted program valid under the repository's strict host-C flags without
performing a second analysis of whether that parameter appears in the body.

`size:S` is the honest bound: Stage 1 already has the general loop and Stage 2
already has the BindingId-based Bytes specialization. This issue replaces that
specialization with one parameter-list loop in both canonical halves and adds
one focused lasting gate. It adds no source diagnostic, type rule, or use
analysis. `P1` reflects an ordinary accepted function shape failing after
Kofun has handed generated source to the host compiler.

## Current behavior and exact evidence

Measured from the exact audited commit with a freshly compiled
`bootstrap/stage2/compiler.c`, not a reusable `bin/kofun` build artifact:

```kofun
fn pick(a: Int, b: Int) -> Int {
    return a
}

fn main() -> Int {
    print(pick(1, 2))
    return 0
}
```

```sh
audit_commit=a858e0fa6f0060c4ecc2d0299852f4282014a9b5
work=$(mktemp -d)
trap 'rm -rf "$work"' EXIT HUP INT TERM
git archive "$audit_commit" \
  bootstrap/stage2/compiler.c \
  bootstrap/stage2/decimal_v1.c \
  bootstrap/stage2/decimal_v1.h \
  unicode vendor/utf8proc |
  tar -x -C "$work"

printf '%s\n' \
  'fn pick(a: Int, b: Int) -> Int {' \
  '    return a' \
  '}' \
  '' \
  'fn main() -> Int {' \
  '    print(pick(1, 2))' \
  '    return 0' \
  '}' > "$work/unused.kofun"

[...]

PR #1557, opened 2026-08-20
stage2: four places the Kofun half and the C seed disagreed (#1546, #1545, #1571)

hjosugi:
Closes #1546. Closes the `ends_with` row of #1508.

## Two disagreements between the halves, both invisible to every gate

`bootstrap/stage2/compiler.kofun` is neither `canonical_source` nor
`trusted_seed` in `bootstrap/manifest.json`, so nothing compiles it and nothing
runs the result. These are two things that were wrong in it.

**`ends_with` was called and never defined.** `compiler.kofun:18585` decides the
output mode with `if ends_with(output_path, ".c")`. The C seed writes the helper
as a `static bool` beside its own caller; the Kofun half had no counterpart, and
no gate in the tree resolved a name in that file until #1509 landed
`task stage2-pair-calls` yesterday.

**The selfhost body walk numbered its symbols one short.**
`selfhost_functions_hir` seeded at `count + 17`, which was the first free id
when sixteen builtins plus the `len(List)` overload were everything there was.
**#753 appended `fail` at `count + 17`.** The C seed's running counter absorbed
the new symbol; this literal did not. Every parameter and local would have taken
an id one lower than the seed's, and the first would have duplicated `fail`.

The checked-in fixture settles which half was right:

```sh
grep -c '^symbol|[0-9]*|function|' bootstrap/selfhost/frontend/S.hir   # 105
grep -n '^symbol|12[23]|' bootstrap/selfhost/frontend/S.hir
# symbol|122|builtin|fail|22|0|0
# symbol|123|parameter|char|1|709|713
```

105 functions, so `count + 17 == 122` is `fail` and the first body symbol is
`count + 18`.

## The fix is not `count + 18`

A literal that encodes another function's arithmetic is what failed here, and a
corrected literal leaves that intact — the next builtin reopens the same gap.
The seed is now **passed**, derived from the registration's own output:

```kofun
selfhost_line_count(selfhost_lines(registered, "symbol|"))
```

so it cannot drift from the thing it is supposed to follow.

## Both ledgers refused the change until it was recorded

This is the first time either has been asked, and both behaved as designed:

```
FAIL: pair calls: unresolved-calls.tsv records 'ends_with', which now
  resolves. Remove the row: a fix that stays on the ledger leaves
  slack for the next unresolved call to hide in.

FAIL: pair mirror: tests/pair-coverage/mirror.tsv no longer describes the tree:
    ends_with is recorded c-only and compiler.kofun now defines it.
[...]

PR #1563, opened 2026-08-20
stage2: a ninth labelled parameter is a bound, not an unknown label (#1537)

### LINKED ISSUE #1537: Stage 2 refuses a correct labelled call when the callee declares more than eight parameters
## Metadata

- State: in-progress
- Kind: implementation
- Area: Compiler / Stage 2 call binding
- Priority: P1
- Size: S
- Parent: #1401
- Blocked by: none
- Audited commit: `cd18b9f303553bdeef6647a08dd8055f46c4a9ed`
- Evidence measured: 2026-08-21

## Goal

Remove the accidental eight-entry truncation from the executable C Stage 2
labelled-call binder. A declared ninth (and later) external label must bind to
its declaration slot under the accepted call-arguments v1 rules, exactly as it
already does in the Kofun half and as nine positional parameters already do.

This is a repair, not a language-limit decision: this issue must **not** add an
eight-parameter limit to Core calls.

## Contract already selects the outcome

`spec/syntax/call-arguments-v1.md` is accepted and says that each declared
external label occurs exactly once, labelled arguments may appear in any source
order, and their values are placed in declaration-order parameter slots. It
states no parameter-count limit. The ordinary positional path already accepts
nine parameters, while the Kofun implementation walks
`parameter_count(source, declaration)` without an eight-entry cap.

Therefore the only in-scope outcome is to make the executable C half inspect
all parameters in the resolved declaration. Introducing a new global limit,
or reusing the self-host signature checker's `E2S17`, would be a language change
and is not part of this bug fix.

## Measured current behavior

The following three inputs were measured from the exact audited commit.

`/tmp/issue1537-eight.kofun`:

```kofun
fn eight(a x1: Int, b x2: Int, c x3: Int, d x4: Int, e x5: Int, f x6: Int, g x7: Int, h x8: Int) -> Int {
    return x1 + x8
}

fn main() -> Int {
    return eight(a: 1, b: 2, c: 3, d: 4, e: 5, f: 6, g: 7, h: 8)
}
```

`/tmp/issue1537-nine.kofun`:

```kofun
fn nine(a x1: Int, b x2: Int, c x3: Int, d x4: Int, e x5: Int, f x6: Int, g x7: Int, h x8: Int, i x9: Int) -> Int {
    return x1 + x9
}

fn main() -> Int {
    return nine(a: 1, b: 2, c: 3, d: 4, e: 5, f: 6, g: 7, h: 8, i: 9)
}
```

`/tmp/issue1537-positional-nine.kofun`:

```kofun
fn nine(a: Int, b: Int, c: Int, d: Int, e: Int, f: Int, g: Int, h: Int, i: Int) -> Int {
    return a + i
}

fn main() -> Int {
    return nine(1, 2, 3, 4, 5, 6, 7, 8, 9)
}
```

Exact measurement command:

```sh
audit_commit=cd18b9f303553bdeef6647a08dd8055f46c4a9ed
audit_dir=$(mktemp -d)
git archive "$audit_commit" \
  bootstrap/stage2/compiler.c \
  bootstrap/stage2/decimal_v1.c \
  bootstrap/stage2/decimal_v1.h \
  unicode vendor/utf8proc |
  tar -x -C "$audit_dir"

cc -std=c11 -O1 -Wall -Wextra -Werror -pedantic \
  "$audit_dir/bootstrap/stage2/compiler.c" -o "$audit_dir/stage2"

"$audit_dir/stage2" /tmp/issue1537-eight.kofun \
  "$audit_dir/eight.c" "$audit_dir/eight.ir" "$audit_dir/eight.tokens"
echo "$?"

"$audit_dir/stage2" /tmp/issue1537-nine.kofun \
  "$audit_dir/nine.c" "$audit_dir/nine.ir" "$audit_dir/nine.tokens"
echo "$?"

"$audit_dir/stage2" /tmp/issue1537-positional-nine.kofun \
  "$audit_dir/positional.c" "$audit_dir/positional.ir" \
  "$audit_dir/positional.tokens"
echo "$?"
```

Measured result:

```text
compiler build: exit 0
eight labelled parameters: exit 0
stdout: <audit_dir>/eight.c
stderr: empty
nine labelled parameters: exit 1
stdout: error[E2S162]: unknown call label `i` at byte 221
stderr: empty
nine positional parameters: exit 0
stdout: <audit_dir>/positional.c
stderr: empty
```

The failure is therefore specific to the labelled binder's ninth declaration
slot. It is not a general declaration-arity refusal.

## Cause and pair divergence

At the audited commit,
`bootstrap/stage2/compiler.c:10932-10941` gives
`validate_declared_call_arguments` five fixed eight-element arrays and stops
scanning when `parameter_count_value >= 8`:

```c
int64_t parameter_starts[8];
int64_t parameter_external[8];
int64_t parameter_internal[8];
int64_t parameter_types[8];
int64_t bound_argument[8];
/* ... */
if (parameter_count_value >= 8) break;
```

The omitted ninth declaration can never match `i:`, so a declared label is
misreported as unknown. In contrast,
`bootstrap/stage2/compiler.kofun:7107-7188` takes the declaration's full
`parameter_count` and walks every slot.

The only `parameter limit is 8` diagnostics at this commit are
`bootstrap/stage2/compiler.c:29612` and
`bootstrap/stage2/compiler.kofun:18530`, both in the separate self-host
signature pipeline. They do not define a limit for ordinary Stage 2 Core calls.

## Required repair

- Make the C `validate_declared_call_arguments` inspect and bind every
  parameter in the resolved declaration; remove the five-entry-table truncation
  and the `>= 8` break.
- Match the existing Kofun-half rule: the declaration's actual parameter count
  is authoritative. Do not add an eight-parameter refusal to either half or to
  the positional path.
- Preserve all current labelled-call rules and diagnostics for genuinely
  unknown, duplicate, missing, positional-after-labelled, and internal-name
  cases, including declaration-related spans where one exists.
- Preserve source-order evaluation and declaration-order ABI slots. The fix is
  compiler-side bookkeeping only; emitted programs must still use ordinary
  fixed parameter temporaries and must not allocate or dispatch through a
  runtime label dictionary.
- If the C repair allocates bookkeeping from the measured parameter count,
  check count-to-size overflow and allocation failure before indexing it and
  release it on every return. A complete repeated source walk is also in scope
  and avoids that allocation.
- Add a focused `tests/conformance/call-arguments/` fixture with at least nine
  externally labelled parameters. Spell the labels out of declaration order
  and execute the result so the test proves that the ninth slot is bound, that
  source order is retained, and that declaration-order lowering is retained.
- Make that focused gate fail when the old eight-entry truncation is restored;
  a compile-only fixture that never observes slot nine is insufficient.
- Regenerate `bootstrap/stage2/SHA256SUMS` after the final pair files are stable:

  ```sh
  ./bin/kofun-digest \
    bootstrap/stage2/compiler.kofun \
    bootstrap/stage2/compiler.c > bootstrap/stage2/SHA256SUMS
  ```

  Do not edit `compiler.kofun` merely to manufacture a diff: it already carries
  the selected all-parameters rule. Its behavior must remain the semantic
  authority that the C repair matches.

## Out of scope

- Changing or removing the self-host signature pipeline's explicit
  eight-parameter limit and `E2S17`.
- Adding any new language-visible parameter limit.
- Widening unrelated fixed tables in the C seed.
- General work on overload selection, indirect/lexical calls, backend call
  support, or the runtime ABI.
- Building a general dynamic runtime label table.

## Acceptance criteria

- [ ] A function with nine externally labelled parameters, called with all
      labels in a non-declaration source order, compiles and executes with the
      expected result through the C11 Stage 2 path.
- [ ] The parameter declared ninth binds slot 8; it is not reported as
      `E2S162`, and a genuinely unknown label still is.
- [ ] Reintroducing the old `>= 8` truncation makes the focused
      `task call-arguments` assertion fail.
- [ ] The nine-parameter positional control case remains accepted.
- [ ] Source-order evaluation, declaration-order lowering, ownership/type
      observations, trailing-lambda binding, and existing E2S162-E2S167
      diagnostics remain green.
- [ ] The C half has no fixed eight-entry table in this binder and implements
      the existing all-parameters Kofun rule without runtime label dispatch.
- [ ] `bootstrap/stage2/SHA256SUMS` matches the final canonical pair.
- [ ] All named validation commands pass after exclusive access to the pair
      lane is obtained.

## Validation

| Property | Command | Expected |
| --- | --- | --- |
| ninth and later labels bind real declaration slots | `task call-arguments` | PASS, including the new executable fixture |
| accepted v1 contract remains exact | `task call-arguments-spec` | PASS |
| labelled-call diagnostics and spans remain stable | `task diagnostics` | PASS |
| canonical pair digests and Stage 2 checkpoint agree | `task stage2` | PASS |
| self-host generations remain fixed | `task selfhost-fixed-point` | PASS |
| pair coverage ledger remains honest | `task stage2-pair-coverage` | PASS |
| repository invariants | `task repository-check` | PASS |
| full suite | `task verify` | PASS |

## Serialization

This issue changes the canonical Stage 2 pair. Before claiming it, announce in
the issue and obtain the exclusive pair lane. Do not edit the pair or run the
pair gates concurrently with another live pair claim. Keep the canonical claim
event current through `active` -> `pr-open` -> `merged` (or `released`).

## Dependencies

Blocked by: none.

The live pair lane is serialization, not a semantic dependency: implementation
must wait until its current owner releases or merges it, then claim this issue
before editing.

hjosugi:
Closes #1537.

**Not for merge until `kofun-5f` announces THAW** — a release freeze is in
effect and `main` must not move between the commit CI proves and the tag.

## The defect

`validate_declared_call_arguments` held its parameter table in five arrays of
eight and stopped filling them at the eighth. The ninth parameter was never
recorded, so the label naming it matched nothing:

```
error[E2S162]: unknown call label `i`
```

The label is declared. A reader following that message edits a correct line.
Nine parameters were never the problem: the same declaration called
positionally compiled and ran, and still does.

## The bound is not new

A parameter is a lexical binding, and `E2S35` refuses the 257th binding in a
function before this code is reached. Measured, one program per row, labelled
calls:

| declared parameters | result on `main` |
| --- | --- |
| 255 | `E2S162` on the ninth label — this defect |
| 256 | `E2S162` on the ninth label — this defect |
| 257 | `E2S35` lexical binding limit is 256 per function |
| 260 | `E2S35` lexical binding limit is 256 per function |

So 256 is the count that can actually arrive, and sizing to it makes the array
bound provable from the function rather than from a rule enforced several
passes earlier. The guard **declines** rather than breaking, so even if it were
reached it would fall back to the positional path instead of fabricating a
label error out of a truncated table — which is exactly how the eight-parameter
version failed.

## Fixing the binder exposed a second cap, and that is the more interesting half

`fixed_slot_call_shape` carries at most eight parameters deliberately. Its
refusal shared one sentence with two unrelated holds, and that sentence's own
comment says why a third must not join it:

> Two refusals reached this one sentence, and only one of them is a hold. …
> Reporting both as "owned by #882" promises the second one will start working.

A declaration with more parameters than the lowering carries is a **bound**,
not a hold. It now reports:

```
error[E2S158]: a labelled call binds at most 8 parameters; `nine` declares 9
```

The eight is named once, in `STAGE2_FIXED_SLOT_PARAMETER_LIMIT`, and read twice
— by the shape test and by the sentence. It was a bare `8` in one place and
nothing in the other, which is how a bound came to be reported as a hold.

## Both halves

The Kofun half never had the binder cap, so it **accepted** the nine-parameter
program the C half refused — a divergence nothing could see, because
`compiler.kofun` is never semantically compiled. Both halves now carry the same
named limit and the same sentence.

## Sizing

The sentence is 151 bytes worst case against the typed sidecar's frozen
160-byte detail field with a 63-byte identifier — the bound that bit #1321
twice, now checked where the sentence is written.

## Fixtures

`labelled_parameter_bound.kofun` fails before this change with `E2S162` and is
refused by the bound after it. `positional_beyond_bound.kofun` is accepted
before and after — the half a fix could most easily break, since narrowing
declarations to eight would be a language change wearing a repair's clothes.

Both use every parameter, because a function that ignores one does not compile
for an unrelated reason (#1548).

## Verification

Green locally: `task check`, `task stage2`, `task stage2-pair-calls`,
`task stage2-pair-mirror`, `task call-arguments`. `task diagnostics` was killed
by resource pressure at load 48 rather than failing — CI is GitHub-hosted and
is the honest place to run it while this box is saturated.


hjosugi:
Release audit found a contract reversal, so this is not mergeable yet.

- #1537 requires an out-of-declaration-order call with 9 labelled parameters to compile and execute. This head instead adds a fixed-slot parameter limit of 8 in both halves and turns parameter 9 into E2S158.
- The new positive fixture is declaration-order and is not executed, so it does not prove slot 8/source-order/declaration-order acceptance.
- spec/syntax/call-arguments-v1.md names supported carriers but does not define an arity cap; the emitter loops are count-driven.
- The issue claim is still active after the PR opened; it needs a canonical pr-open event once the corrected head is ready.
- The branch is behind current main, and #1557 also touches the canonical pair. Rebase/serialize, regenerate pair and evidence files, then rerun exact-head CI.

Please remove or widen the downstream bare-8 cap, add an executable out-of-order 9-label fixture with golden/ABI assertions, preserve the genuine unknown-label diagnostic, and keep the PR unmerged until the stated THAW.

Issue #1556, opened 2026-08-20
Seventy-one Bytes locals are refused with another rule's diagnostic, and only by one half of the pair


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation issue
- Kind: implementation
- Area: Compiler / Stage 2 C11 lowering
- Priority: P2
- Size: S
- Parent: #1401
- Blocked by: none
- Audited against: `origin/main@a858e0fa6f0060c4ecc2d0299852f4282014a9b5`

## Goal

Give the seventy-one-`Bytes`-local refusal a name of its own, and decide whether
the Kofun half should have the same limit — today it has none.

## Current behavior and evidence

The C half accumulates one `kofun_bytes_release(&k_bN); ` per live `Bytes`
binding into 2048-byte buffers (`compiler.c:20854-20856`, `21800-21801`) and
refuses when the list no longer fits. The guards are correct — the code says so
at `:20841` ("Overflow refuses rather than truncates") and I found no truncation
and no memory unsafety. **The refusal itself is the defect**, in two ways.

```sh
cc -std=c11 -O1 -w bootstrap/stage2/compiler.c -o /tmp/stage2
for n in 70 71; do
  python3 -c "
n=$n
body = ''.join('    let b%d = stage2_bytes_empty()\n' % i for i in range(n))
open('/tmp/b.kofun','w').write('fn main() -> Int {\n%s    return 0\n}\n' % body)"
  printf '%s: ' "$n"; /tmp/stage2 /tmp/b.kofun /tmp/b.c /tmp/b.ir /tmp/b.tokens
done
# expected current result:
# 70: (compiles, writes /tmp/b.c)
# 71: error[E2S170]: Stage 2 Bytes[65536] cannot lower possible managed Bytes
#     alias for this body (backend limitation) at byte 2473
```

**1. The code means something else.** `tests/diagnostics/stage2/e2s170_bytes_backend_limitation.kofun`
says what `E2S170` names, in its own words:

> #1315 gave the branch, loop, recursion and alias shapes their own reasons, so
> `backend limitation` now names exactly one thing: **a crossing the backend has
> no carrier for**.

Its expected output names the binding — ``cannot lower possible managed Bytes
alias for `b` (backend limitation)``. The capacity refusal borrows that code and
substitutes `for this body`, so one code now covers "this crossing has no
carrier" and "your function has too many `Bytes` locals for a 2048-byte
[...]

Issue #1553, opened 2026-08-20
The Bytes mutation gate claims transactionality and a build matrix it does not exercise


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation issue
- Kind: test-quality
- Area: Stage 2 Bytes mutation evidence
- Priority: P1
- Size: S
- Related: #1321 is the merged surface; #1539 owns the broader accepted-source-to-C property; #1401 owns canonical pair ledgers
- Blocked by: none
- Audited against: `origin/main@a858e0fa6f0060c4ecc2d0299852f4282014a9b5`

## Goal

Make `task bytes-mutation` prove the failure preservation and build matrix its PASS lines and #1321 acceptance criteria already claim, without changing the canonical Stage 2 compiler pair or widening the Bytes surface.

## Current behavior and evidence

Measured on the audited `main` commit after #1321 merged:

```sh
git diff --check e882af3fed5120158aa60cc6743a9e7536238d72...a858e0fa6f0060c4ecc2d0299852f4282014a9b5
# tests/conformance/bytes-mutation/run.sh:68: trailing whitespace

sed -n '45,75p' tests/conformance/bytes-mutation/mutation_driver.c
# remember() snapshots length, capacity, and bytes, but not v->data

sed -n '455,485p' tests/conformance/bytes-mutation/mutation_driver.c
# the injected-allocation block exercises append, reserve, append_self, and read;
# append_range and its source carrier are absent

sed -n '145,185p' tests/conformance/bytes-mutation/run.sh
# the strict driver is -O2; the sanitizer driver is -O1; no -O0 driver exists

sed -n '190,250p' tests/conformance/bytes-mutation/run.sh
# refuses() removes and checks only the .c output; it does not remove/check .bin
```

The driver’s value/length loop sets and reads selected positions and copies the carrier, but does not perform append-one for every `{0x00,0x7f,0x80,0xff}` × `{0,1,255,16384,65536}` cell while the final PASS line claims the exact matrix. The allocation-budget branch cannot detect a failed `append_range` changing either carrier, and the common unchanged witness cannot detect a pointer change that preserves length, capacity, and bytes.

## Scope

- Snapshot and compare pointer identity in the common failure witness.
- Exercise injected-OOM `append_range` and prove destination and source length, capacity, pointer, and bytes remain exact.
- Run the named append-one byte × length matrix under strict O0, strict O2, and ASan/UBSan builds.
[...]

Issue #1540, opened 2026-08-20
Stage 2 positional `take` crossings: record the move and fail closed on record `edit`


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation issue
- Kind: implementation
- Area: Ownership / Stage 2 compiler
- Priority: P1
- Size: M
- Related: #946 is the landed source-order move producer; #1315 owns the
  Bytes carrier; #1516 owns temporary arguments; #1517 owns crossing-mode
  compatibility; #1244 owns the authority-only CFG/cleanup/escape slice
- Blocked by: none
- Audited compiler baseline: `origin/main@41f7c3de06ff8a47f16c4844dfecfd58b592acd7`
- Remeasured: 2026-09-06 against published `v0.11.6-seed`
- Current-main reachability: PR #1525 and its successors are merged; none blocks this slice

## Goal

Make an ordinary positional call to a direct current-file function record the
same straight-line ownership fact as a written whole-binding `take` when the
callee's resolved parameter is `take` and the argument is one bare, live,
owning binding.

The minimum v1 type slice is non-authority `Bytes` plus nominal records whose
fields are all `Int` or `Bool`. For those trivial records, a `take` parameter is
a static source move even though the existing C ABI copies the bits. An `edit`
parameter is not truthful under that by-value ABI and must fail closed before
C is published.

`size:M` is the honest bound: one existing straight-line move query is widened
to one additional call spelling, but the rule must land in both canonical
compiler halves and prove two carrier classes, one new refusal, BindingId
shadowing, RFC-0010 fixture migration, and lasting gate wiring. `P1` reflects
an accepted use of a source-dead value and a documented ownership guarantee
that currently does not hold; this is not a claim of runtime memory unsafety.

## Current-main remeasurement (2026-09-06)

Compiled the production Stage 2 C from `v0.11.6-seed` with
`cc -std=c11 -Wall -Wextra -Werror -pedantic -O0`, then passed each of the
new `tests/conformance/move-call-crossings/{bytes_edit_after,bytes_double,
record_use_after,record_double,record_edit}.kofun` sources through that tool
and strictly compiled/executed the resulting C. All five incorrectly succeed:
Bytes later-edit prints `4`, Bytes double-transfer prints `4` then `0`, and
both record reuse fixtures and the by-value edit fixture print `7` then `7`
(or `7` once for edit). These are inherited defects at the selected previous
release, not regressions introduced by this queue.

[...]

PR #1564, opened 2026-08-20
spec: the bounded Bytes carrier, and the claim it makes possible (#1558)

### LINKED ISSUE #1558: The bounded Bytes carrier has no specification, so it cannot be claimed
## Metadata

- State: in-progress
- Kind: spec-design
- Area: Standard library / Stage 2 Bytes
- Priority: P1
- Size: M
- Related: #1315 built the carrier; #1321 built the mutation surface; #1322,
  #1323, #1324 will each widen it
- Blocked by: none
- Audited against: `origin/main@a858e0fa6f0060c4ecc2d0299852f4282014a9b5`

## Goal

Specify the bounded `Bytes` carrier and its mutation surface in one tracked
document, so the capability can be claimed in `release/claims.json` at all.

## Why this blocks the claim, measured

`release/claims.json` has 53 claims and **none of them is the Bytes carrier**:

```sh
python3 -c "
import json
cl = json.load(open('release/claims.json'))['claims']
print([c['id'] for c in cl if 'byte' in c['id'] or 'carrier' in c['id']])"
# []
```

So #1315's carrier is unclaimed and #1321's mutation surface is unclaimed, and
writing a claim for either is not appending a row — it is introducing the
capability to the published manifest for the first time.

**And it cannot be done, because a claim must cite a specification that
exists.** `tests/release/validate-claims.mjs:388` runs every
`specification` entry through `checkPath(..., tracked)`, so the path must be
present *and* tracked by git. There is nothing to cite:

```sh
git grep -l 'Bytes\[65536\]' -- 'docs/*' 'spec/*' 'rfcs/*'
# (no matches)
grep -c 'stage2_bytes_' spec/benchmark-report-v1.md
# 0
```

`docs/MEMORY_MODEL.md` specifies `read`/`edit`/`take` generically and uses
`Bytes` in an example, but says nothing about the 65,536 ceiling, the growth
ladder, the status tags, the alias refusal, or the operations. The only places
those are written down are two closed issues and the emitted C.

## Scope

In scope: one document — `spec/bytes-bounded-v1.md` — specifying what is
already implemented and gated, and nothing more:

- the carrier: three fields, frozen field order, `{0,0,NULL}` as both empty and
  consumed, the 65,536 ceiling;
- the nine operations, their argument order, and what each returns;
- the growth ladder: capacity 0 to 16, then doubling until the request fits,
  capped at the ceiling;
- the ordering of the range checks, and which operand a refusal reports —
  including that `offset + count` is never computed;
- transactionality: every failure leaves length, capacity, pointer and bytes
  unchanged;
- the status tags 0..8 and the read carrier's tags 0..2, marked as private to
  the emitted C, with the reason;
- the alias refusal (`E2S170`) and the carrier-identity refusal (`E2S177`);
- what is deliberately absent: the Text bridge (#1322), atomic replacement
  (#1323, #1324), and any source-visible outcome beyond `len` and `capacity`.

Then the claim, in the same change or the next: a `release/claims.json` entry
citing this document, and whatever `stdlib/capabilities.tsv` row the matrix
wants.

Out of scope: specifying anything not yet implemented. A specification that
describes the Text bridge before #1322 lands is a promise, and this repository
has a word for a published promise nothing checks.

## Why one document for both issues

A carrier you cannot change is not a capability anyone can depend on, so
#1315 alone has nothing a user could rely on, and #1321 alone has no subject.
Splitting them across two claims would publish a promise nobody could use and
then amend it a week later.

## Acceptance criteria

- [ ] `spec/bytes-bounded-v1.md` exists, is tracked, and states every bullet
      above, each traceable to the gate that already proves it.
- [ ] Every normative statement in it names the gate that would fail if it were
      false — `task bytes-carrier` or `task bytes-mutation` — so the document
      cannot drift from the implementation without a gate going red.
- [ ] No statement describes behaviour that is not implemented and gated today.
- [ ] `release/claims.json` gains one entry citing it, and
      `node tests/release/validate-claims.mjs` passes with the specification
      path resolving.
- [ ] The claim's `negative_boundary` cites a real refusal fixture, and its
      `positive_gate` observation is a sentence that gate actually prints.
- [ ] `task release-claims` and `task capabilities` pass.

## Validation

| --- | --- | --- |
| the claims manifest | `task release-claims` | PASS with the new entry |
| the capability matrix | `sh stdlib/check-capabilities.sh` | PASS |
| the surface itself | `task bytes-carrier && task bytes-mutation` | PASS |
| repository | `task verify` | green |

## Dependencies

Blocked by: none. This does not touch the canonical Stage 2 pair.

hjosugi:
Closes #1558.

**Not for merge until `kofun-5f` announces THAW.** Aimed at tomorrow's release,
which is where the claim was deliberately deferred to.

## What was actually blocking the claim

`release/claims.json` had 53 claims and **none of them was `Bytes`**. #1315's
carrier and #1321's mutation surface were both unclaimed, so a user reading the
manifest had no way to know either existed or what they promised.

They could not be claimed either. `tests/release/validate-claims.mjs:388` runs
every `specification` entry through `checkPath(..., tracked)`, and there was
nothing to cite:

```sh
git grep -l 'Bytes\[65536\]' -- 'docs/*' 'spec/*' 'rfcs/*'
# (no matches)
```

`docs/MEMORY_MODEL.md` specifies `read`/`edit`/`take` generically and uses
`Bytes` in an example, but says nothing about the ceiling, the growth ladder,
the status tags, the alias refusal, or the operations. The only places those
were written down were two closed issues and the emitted C.

So this adds the document first and the claim second.

## The document

`spec/bytes-bounded-v1.md` specifies what is implemented and gated today and
nothing else. **Every normative statement names the gate that fails if it is
false** — `task bytes-carrier` or `task bytes-mutation` — so the document
cannot drift from the implementation without a gate going red. A statement with
no gate does not belong in it.

It says what the capability is *not* at least as carefully as what it is: not a
general buffer, no `Text` bridge, no atomic replacement, and observable from
source only through `len` and `capacity`.

## One claim, two issues

A carrier you cannot change is not a capability anyone can depend on, so #1315
alone has nothing to rely on and #1321 alone has no subject. Two claims would
have published a promise nobody could use and amended it a week later.

## Known gaps are in the specification, not omitted

`## Known gaps` names #1540, #1516, #1517, #1553 and #1556 — four filed against
this surface in the last day, and **#1553 is a review of the gate that proves
this claim**, whose findings I have acknowledged in full. A specification that
lists only what works is the kind of published promise this repository gates
against everywhere else, and the claim beside it is honest only if the document
says where the edges currently leak.

## The manifest binding

`docs/MVP_IMPLEMENTED.md` gains the row the manifest binds to, because
`public_wording` and `public_status` must join a published source — which is
what stops one capability being described two different ways. Both were
corrected until the checker agreed rather than the other way round.

## Verification

```
PASS: spec/release-claim.schema.json is a supported schema
PASS: 54 release claims join their published wording and evidence
```

Green locally: `task release-claims`, `task capabilities`,
`task documentation-index`, `task repository-check`. The evidence pack is
regenerated with `task release-evidence` rather than hand-edited.


hjosugi:
## Independently derived, from a clean base

@kofun-60 wrote the same claim entry as #1565, without seeing this branch, and
has closed it in favour of this one. Recording what that independent pass
confirms, because two derivations of the same constraints are better evidence
that the manifest's rules are what we think they are than one of us running it
twice:

- `task release-claims` green at **54 claims**;
- `task release-evidence` regenerating to a clean diff;
- `docs/MVP_IMPLEMENTED.md` needing the published row before the claim is
  accepted at all — the `public_wording` / `public_status` binding.

Its reason for closing #1565 in favour of this one is the reason #1558 was
filed as a specification issue rather than a claims issue: **#1565's
`specification` field had to point at `docs/MEMORY_MODEL.md`**, which
describes `read`/`edit`/`take` generically and says nothing about the
ceiling, the growth ladder, the status tags or the operations. And
`specification` is exactly the field `task release-claims` **cannot** check
for accuracy — it checks the path exists and is tracked, not that it says
anything relevant. So an approximate link there is a quiet imprecision rather
than a caught one.

That is why this PR adds `spec/bytes-bounded-v1.md` first and the claim
second. The missing document was the blocker, not a nicety.

**The near-miss is worth recording too:** it was asked for wording, I sent it,
and then I wrote the artifact without saying I was doing so. Both readings of
"send me the wording" were reasonable and neither of us said which one we were
acting on. The rule we have both adopted — *say which artifact you are about to
create before you create it* — would have prevented it, and it is the same
shape as the duplicate implementation on #1523 earlier today.

hjosugi:
Release audit found that the claim is not publishable as written yet.

- release/claims.json marks this bounded Bytes[65536]/Stage 2 C11 slice as implemented, while the schema defines implemented as general implementation and uses checkpoint for bounded executable slices. Change it to checkpoint and regenerate evidence.
- The claim/spec promise O0/O2 sanitizer agreement and preservation of pointer/all inputs, but the current gate builds the contract driver at O2, sanitizes only O1, does not snapshot the pointer, and omits append_range/source from the OOM branch. #1553 is active for this exact evidence gap and must land first.
- The published identity/private-outcome semantics contradict measured #1561 and #1559 behavior; #1560 is also a silent same-name declaration hijack. At minimum land #1559/#1561 and accurately bound every remaining gap before publishing a memory-safety claim.
- The issue claim remains active after PR opening. The same claim owner must append pr-open before merge and merged immediately after; merging with Closes #1558 now would make Backlog fail.
- Rebase/update from current main and rerun exact-head CI after the stated THAW. Kofun verification is still pending on the stale base.

Adding the corrected claim still requires the next version to rise from 0.9.0-seed to 0.10.0-seed. Please also revert the unrelated em-dash JSON normalization.

hjosugi:
Additional release blocker discovered after the earlier audit: #1569 directly contradicts section 3 and the memory-safety classification at this head.

The spec says that a function declaring a Bytes owner reclaims it at every exit and that bytes-carrier derives this by counting every return. #1569 reproduces three typed-return trap guards (List[Int], Int?, and enum) that return without bytes_release, while the success path and neighboring Text/record arms release. The present carrier corpus has no fixture owning Bytes across those typed returns, so the stated derivation does not cover them.

Before publishing bounded-bytes as a memory-safety claim, either land #1569 and regenerate exact-head evidence, or explicitly list it as a known gap and narrow section 3, public_status, bounded_capability, and safety classification so none says every exit or general unique ownership. The latest current main is also ahead of this stale base; all earlier checkpoint, cc prerequisite, carrier+mutation reproduction, unrelated JSON normalization, #1553, #1559/#1560/#1561/#1562, and claim-lifecycle blockers still apply.

Issue #1324, opened 2026-08-11
Stage 2 C11: atomically replace a bound target with complete Managed Bytes


hjosugi:
## Metadata

- State: blocked
- Parent: #1316
- Kind: implementation
- Area: Stage 2 C11 / atomic Bytes replacement
- Priority: P1
- Size: M
- Blocked by: #1315 and #1323
- Audited baseline: `origin/main@3f72c7df8f5c87a7c92a96ff62042bb6cca2745e`

## Goal

Consume one live `AtomicWriteAuthority` and read-borrow one already-complete
0..65,536-byte Managed Bytes value, then atomically replace only the authority's
bound target on local Linux. Transport bytes only; do not encode, choose a path,
issue authority, or claim crash-durable directory persistence.

## Exact adapter

- Use only `openat`/write/`fsync`/consuming close/`renameat`/`unlinkat` relative
  to the bound fd. Temp candidates are the fixed 32 names, created
  `O_WRONLY|O_CREAT|O_EXCL|O_NOFOLLOW|O_CLOEXEC`, mode 0600. A stale candidate
  may be unlinked only while the live authority covers the whole namespace and
  carries its trusted persistent provenance; filename shape or a lock alone
  never grants cleanup.
- Sequence: bounded stale/candidate exclusive create; write all with short-write
  handling; fsync temp; consuming close; exactly one `renameat`.
  Open/write/fsync/unlink retry at most 128
  consecutive EINTR; successful zero write and 129th fail. Close and rename are
  never retried. Linux rename has no documented EINTR under this profile.
- Rename zero is the sole commit point. Negative is precommit because the
  authority excludes remote/ambiguous filesystems. Every earlier failure
  preserves the prior target and attempts cleanup only of a temp
  exclusively created by this call or stale name covered by the live
  reservation. Cleanup failure may leave only that unaccepted temp. Directory
  close after commit is discarded and cannot relabel success; no failable
  directory fsync follows commit.
- Bounded v1 has no in-flight cancellation token or ambient cancellation state.
  A caller may decline and explicitly release the authority before invoking
  replacement; after `replace(take authority, read bytes)` begins, it runs to
  one terminal syscall outcome. BR011/report cancellation is resolved before
  writer invocation. Every adapter outcome consumes/releases authority once.
- Closed outcomes distinguish published, invalid/limit/reservation,
  create/write/sync/close/rename, and cleanup failures. No path argument,
[...]

Issue #1323, opened 2026-08-11
Stage 2 C11: implement the AtomicWriteAuthority provider and directory reservation


hjosugi:
## Metadata

- State: blocked
- Parent: #1316
- Kind: implementation
- Area: Ownership / atomic-write authority provider
- Priority: P1
- Size: M
- Blocked by: #1536 and #1196
- Landed decision input: #1317 (option A, root-derived)
- Audited baseline: `origin/main@3f72c7df8f5c87a7c92a96ff62042bb6cca2745e`

## Goal

Implement the safe one-shot `AtomicWriteAuthority` carrier/provider selected by
#1317 by deriving through #1536's sole runtime RootAuthority identity and
adding exactly one pinned atomic-write entry to #1196's default-deny entitlement ABI. Deliver
directory/target binding, unforgeable local-filesystem evidence, and exclusive
same-directory target+32-temp reservation. Do not implement rename/write.

## Exact selected-option implementation boundary

- #1317 selected its root-derived option A (recorded 2026-08-14). Safe issuance
  validates a nonempty **1..255-byte `Bytes`** basename with no NUL or slash and
  rejects `.`, `..`, and all
  `.kofun-atomic-00.tmp`..`.kofun-atomic-31.tmp`; takes the provider's preopened
  publication directory rather than resolving a path; and never exposes a
  path/raw descriptor/attestation Boolean constructor. (This replaces an earlier
  `1..220-byte UTF-8` bound — see the 2026-08-14 comment for why both halves
  changed.)
- Refuse NFS, CIFS/remote/unknown, mismatched, or missing error-is-noncommit
  filesystem evidence before reservation/cleanup. Evidence is trusted provider
  state and is bound to the open directory identity.
- Acquisition order is validated target/profile, in-process stable-directory
  registry, then nonblocking `flock(fd, LOCK_EX|LOCK_NB)` on the separately
  opened `O_RDONLY|O_DIRECTORY|O_CLOEXEC` fd. `EWOULDBLOCK` means busy,
  `ENOLCK` unsupported, EINTR retries 128 then interrupted, and other errno is
  provider failure. Any failure releases registry/fd; stale cleanup is not yet
  allowed because the complete authority was not issued.
- After both locks, the provider verifies the selected persistent
  dedicated-directory namespace provenance. First enrollment requires all 32
  names absent and records the stable directory identity; later stale adoption
  requires that same trusted record. A generic/ambiguous/prepopulated directory
[...]

Issue #1322, opened 2026-08-11
Stage 2 C11: implement the checked Bytes/Text bridge


hjosugi:
## Metadata

- State: in-progress
- Parent: #1315
- Consumers: #1258 and #1312
- Kind: implementation
- Area: Stage 2 C11 / Bytes-Text bridge
- Priority: P1
- Size: M
- Blocked by: none
- Audited baseline: `origin/main@3f72c7df8f5c87a7c92a96ff62042bb6cca2745e`
- Audited against: `origin/main@6aef329f39b015399c54adf2dda75ff2f162ab62` — re-measured 2026-09-21 (the `v0.12.2-seed` release head); see the evidence section

## Goal

Complete the shared Bytes profile with exact Text-to-Bytes construction,
checked Bytes-to-Text conversion, and exact compiler-private status/detail outcomes,
without pretending current Stage 2 supports generic `Result[Bytes,E]`.

## Current behavior and evidence

Current Text lowers to a NUL-terminated `const char *`; length/equality use
`strlen`/`strcmp` in `bootstrap/stage2/compiler.c`, while arbitrary Bytes may
contain 0x00. Current concrete enum payloads are one `Int` and generic Result is
outside the implemented profile (`docs/MVP_IMPLEMENTED.md`). Therefore a NUL
cannot be silently admitted or misclassified as malformed UTF-8.

Re-measured 2026-09-21 on `origin/main@6aef329f39b015399c54adf2dda75ff2f162ab62` in a detached, clean worktree:

```sh
test "$(git rev-parse HEAD)" = 6aef329f39b015399c54adf2dda75ff2f162ab62 || exit 1
cat > text.kofun <<'K'
fn main() -> Int {
    let greeting: Text = "hi"
    if greeting == "hi" {
        return 1
    }
    return 0
}
K
bin/kofun build text.kofun -o text.bin --emit-c text.c
grep -n 'const char \* k_b0 = "hi"' text.c
# 227:    const char * k_b0 = "hi";
grep -n 'strcmp(kofun_condition_left, kofun_condition_right) == 0' text.c
# 232:        bool kofun_condition = strcmp(kofun_condition_left, kofun_condition_right) == 0;
grep -c 'bytes_assign_text\|BytesTextOutcome' bootstrap/stage2/compiler.c
# 0 — no bridge exists in either pair half
grep -n 'Stage2ByteRead;' bootstrap/stage2/compiler.c
[...]

PR #1565, opened 2026-08-20
claims: publish bounded byte-buffer mutation as a claim (#1321)

hjosugi:
Publishes the Stage 2 bounded `Bytes[65536]` mutation surface as a release claim, so a release can describe it.

`release/claims.json` carried 53 claims and none of them the carrier — #1315 landed it and #1321 gave it a mutation surface, and neither was published. `docs/RELEASING.md` step 8 writes release notes from `artifacts/release-evidence/CLAIMS.md`, which is a projection of that file, so an unpublished capability is one the notes cannot mention.

**The wording is #1321 author's, verbatim.** I did not paraphrase it, because the "is not" half is what makes it a bounded claim rather than a general one: the 65,536 ceiling is fixed rather than configurable, there is no `Text` bridge (#1322), there is no atomic replacement of a bound target (#1323, #1324), and the read outcome and status carriers are private to the emitted C. Each successor moves exactly one of those edges.

The negative boundary is the `E2S177` identity refusal — a range copy between carriers needs two distinct bindings — rather than a capacity failure, because that is the rule a user trips over.

`docs/MVP_IMPLEMENTED.md` gains the row the claim joins; `task release-claims` refuses a claim whose wording is published nowhere.

| Check | Result |
| --- | --- |
| `task release-claims` | PASS — 54 claims join their published wording and evidence, 23 mutations refused |
| `task release-evidence` then `git diff --exit-code` | clean |


hjosugi:
**Closing in favour of #1564, which is a superset of this.**

Two of us wrote this claim within an hour, from the same draft — I asked #1321's
author for the wording, used it verbatim, and they were writing the same entry.
That is the release near-miss one layer down, and the session cutting today's
release caught it before either of us landed.

#1564 is the one to take, and not by seniority: it adds
`spec/bytes-bounded-v1.md`, so the claim's `specification` field points at a
document that states the bound rather than at `docs/MEMORY_MODEL.md`, which is
what this PR had to settle for. A claim whose specification link is approximate
is exactly the kind of quiet imprecision `task release-claims` exists to refuse,
and it happens to be the field the checker cannot check.

Nothing here is lost: the wording in #1564 is the same wording, because it came
from the same author. What this PR contributes is the confirmation that the
entry validates from a clean base — `task release-claims` green at 54 claims,
`task release-evidence` regenerating to a clean diff — which is one fewer
unknown for whoever merges #1564.

Issue #1545, opened 2026-08-20
Stage 2 emits a to_text call as a call to args, through an out-of-bounds read


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation issue
- Kind: implementation
- Area: Compiler / Stage 2 selfhost-HIR
- Priority: P1
- Size: M
- Parent: #1401
- Blocked by: none
- Audited against: `origin/main@a858e0fa6f0060c4ecc2d0299852f4282014a9b5`
- Evidence measured: 2026-08-21

## Goal

Make every builtin outside the frozen selfhost-HIR vocabulary fail closed by
name in both halves of the canonical Stage 2 pair, before any slot indexing or
symbol-id arithmetic. Pin `to_text(1)` and `stage2_bytes_empty()` as witnesses
that the wider Stage 2 builtin tables cannot fall through an unregistered slot.

This issue does not expand the frozen selfhost profile. Its authoritative
16-builtin vocabulary is `bootstrap/selfhost/profile.tsv`, and
`bootstrap/selfhost/hir-v1.md` says v1 covers exactly that profile. `to_text`
is admitted by the wider typed Stage 2 subset but absent from the frozen
selfhost-HIR vocabulary, so the bounded outcome here is the existing E2S15
profile refusal, not a new symbol, helper, or schema expansion.

## Measured current behavior

The exact input was:

```kofun
fn main() {
    print(to_text(1))
}
```

It was saved as `/tmp/issue1545-to-text.kofun` and measured from an archive of
the audited commit so the current working branch could not affect the result:

```sh
audit_commit=a858e0fa6f0060c4ecc2d0299852f4282014a9b5
audit_dir=$(mktemp -d)
git archive "$audit_commit" \
  bootstrap/stage2/compiler.c \
  bootstrap/stage2/decimal_v1.c \
  bootstrap/stage2/decimal_v1.h \
  unicode vendor/utf8proc |
  tar -x -C "$audit_dir"

cc -std=c11 -O1 -Wall -Wextra -Werror -pedantic \
  "$audit_dir/bootstrap/stage2/compiler.c" -o "$audit_dir/stage2"
cc -std=c11 -O0 -g -Wall -Wextra -Werror -pedantic \
  -fsanitize=undefined -fno-sanitize-recover=bounds \
  "$audit_dir/bootstrap/stage2/compiler.c" -o "$audit_dir/stage2-ubsan"

[...]

Issue #1243, opened 2026-08-11
Stage 2 authority kind: propagate Owned through shipped composites


hjosugi:
## Metadata

- State: needs-decision
- Tracker role: curated implementation child
- Kind: implementation
- Priority: P1
- Size: M
- Parent: #1193
- Blocked by: none
- Decision owner: @hjosugi, shepherd of RFC-0002 and RFC-0004 (the two decisions #1193 records for this chain)
- Decision needed: D1–D3 under `## Decision needed` (which channel carries the path, how the path is rendered, and the E353 remedy clause for composites)
- Normative input: accepted RFC-0004 (structural ownership kinds); accepted RFC-0002 (E353)
- Audited against: `origin/main@fe1142a7039a0862ac3558729ab2477bb7a604d6`

## Goal

Stop an RFC-0002 authority from hiding inside a composite type the ordinary
Stage 2 profile accepts today. Classify each accepted record and ADT as Owned or
Managed under RFC-0004 K-JOIN/K-ANY, attach the RFC-0004 minimal causal path to
every Owned classification, and refuse the uses of an Owned composite that
Stage 2 compiles today, where RFC-0002/RFC-0004 require a refusal.

This issue owns classification and enforcement over **forms Stage 2 already
admits**. It does not widen any slice.

## Decision needed

The rest of this body can be implemented once D1–D3 are recorded here, or in a
`spec/` document this issue links to. Until then, write no E353 golden: a
golden written before D1–D3 has to be rewritten afterwards. That was the
warning on 2026-08-20, and it still applies.

The facts behind all three, measured at the stamp:

- `bin/kofun check`/`build` print the diagnostic in full, with no length limit
  (`compile_file`).
- The structured diagnostic (`Stage2StructuredDiagnostic.fallback`,
  `char[160]`, `bootstrap/stage2/compiler.c:69`,
  `bootstrap/stage2/semantic_producer.c:76`) and the typed-sidecar error detail
  (`KOFUN_SEMANTIC_ERROR_DETAIL_BYTES 160u`,
  `bootstrap/stage2/semantic_events.h:9`) keep at most **159 message bytes**.
  Anything longer is cut off, and the cut sets `truncated`
  (`compiler.c:189-193`).
- `task preflight` check 7 (`tests/preflight/check.sh:205`) fails any
  `tests/diagnostics/stage2/*.stderr` larger than 160 bytes (message plus
  newline). `task stage2-events` fails when the producer and the compiler
[...]

PR #1572, opened 2026-08-21
stage2: a parameter the body does not use is discarded, not left to cc (#1548)

### LINKED ISSUE #1548: Stage 2: emit C discards for every admitted function parameter

hjosugi:
Closes #1548.

**Not for merge until `kofun-5f` announces THAW.**

## The defect

```kofun
fn pick(a: Int, b: Int) -> Int {
    return a
}
```

```
error: unused parameter 'k_b1' [-Werror=unused-parameter]
static int64_t kofun_fn_pick(int64_t k_b0, int64_t k_b1) {
```

The rule an author had actually run into was **"every parameter must be
used"** — enforced by gcc, phrased in terms of a name they never wrote, in a
file they never opened. Nothing states it: `git grep -niE 'unused parameter'`
over `spec/` and `docs/` returns nothing.

## #1315 diagnosed this exactly, and fixed it for one type in two of its three modes

`bytes_parameter_discards` emitted `(void)k_bN;` for borrowed `Bytes`
parameters, and said why:

> every such helper would emit a parameter nothing touches and fail
> `-Werror=unused-parameter` … discarding it explicitly is how the emitted C
> says "not yet" rather than not compiling.

That is this defect, correctly named. Every ordinary parameter still had it.

## One rule replaces the special case rather than joining it

Emit the discard when the body does not use the parameter, whatever its type.
`binding_has_use` asks the typed HIR the question `hir_use_binding_id` already
answers in the other direction — the `use` record carries its BindingId in
field 4.

Two mechanisms for one property is how they drift, and the general one is
strictly better: an unused borrow still gets its discard, and a *used* borrow
no longer gets a redundant one.

## It changes the emitted C of no program that compiles today

A parameter with a use never received a discard and still does not; a parameter
without one could not build. Nothing in `tests/` greps for `(void)k_b`, so no
gate depended on the old shape — checked before changing it.

## The fixture, and why its assertions are not the outcome

It lives in `tests/stage2/unused-function/`, which is the same defect one level
up — `check` accepting a program `cc` rejects under the same `-Werror` flags —
rather than in a new task nobody needs.

That file's header already explains why the strict build is the assertion that
matters: *"a golden alone would have passed on any host whose `cc` did not
implement `-Wunused-function`"*. So the new cases pin the **mechanism**:

- first, middle, last and all-unused positions, because a walk that stops early
  or starts late passes with one;
- the discarded ids are **read out of the emitted signature** rather than
  written here, so the fixture does not start testing the binding numbering
  instead of the rule;
- and the converse — a parameter the body *does* use is not discarded — because
  "discard every parameter" compiles just as well and says nothing.

The converse is stated as the property (the discarded id appears nowhere else
in its own body) rather than as a pattern. The first version looked for a
literal `+` and failed, because the emitter lowers addition through
`kofun_add` — exactly the coincidence a pattern-matching assertion mistakes for
the rule.

## Verification

Green locally: `sh tests/stage2/unused-function/run.sh`, and the original repro
now builds and runs through `bin/kofun build`. The rest is on CI — the box is
at load 59 from another session's three concurrent `verify` runs, and
`kofun-5f` needs it quiet for `docs/RELEASING.md` step 1.

## Region, declared

`bytes_parameter_discards` → `unused_parameter_discards` at
`compiler.kofun:4931` / `compiler.c:~20973`, and one call site each. Disjoint
from #1517's `emit_argument` (~380 lines away, different function) and from
#1557's four sites. Neither name appears in `mirror.tsv` or `undefended.tsv`,
so the rename touches no ledger.


hjosugi:
Blocking review at exact head
`1e42bd454d63a04d91605f140c91f9aeca3b3596`:

1. The implementation reverses #1548's load-bearing contract. The issue
   requires exactly one **unconditional** discard for every formal admitted by
   `core_parameters`, including used parameters, and explicitly forbids a
   second use analysis. This head adds `binding_has_use`, emits only for unused
   bindings, and makes the fixture require used parameters *not* to be
   discarded. That is a different product decision, not an implementation of
   the ready issue. Implement the existing unconditional contract, or stop and
   re-refine the issue before code; do not close #1548 with the current design.
2. The issue requires a new lasting `task unused-parameter` wired into verify,
   strict O0 and O2 builds, representatives for every bounded declarator branch,
   and mutation guards for the general prologue plus first/middle/last and
   callable/aggregate positions. Reusing `unused-function` here provides none
   of those load-bearing proofs and currently asserts the contrary design.
3. The canonical Stage 2 pair is not serialized. #1243, #1508, and #1537 still
   have active pair claims; #1545/#1546 remain `pr-open` in #1557, and #1563 is
   open. #1548 itself says to claim only after that lane releases. Text hunks may
   auto-merge, but `bootstrap/manifest.json`, `SHA256SUMS`, and release evidence
   are shared generated authorities, not evidence of safe parallel ownership.

I synchronized #1548's stale `ready` body/label to `in-progress` because its
live claim and open PR otherwise violate the backlog state contract. I did not
change or impersonate your claim; please move your own claim to `pr-open` only
when the serialized, contract-matching head is published.

This P1 change is not part of the active release-blocking queue. Do not merge it
across the current FREEZE. After pair handoff: rebase on exact main, implement
the ready contract and focused mutation gate, regenerate shared authorities
from the final tree, and obtain green exact-head CI.

hjosugi:
## Contract audit: BLOCK at exact `8c65c0882545998dd5d53222642a4673e2f7fdc3`

The current implementation is mechanically green but implements the opposite of #1548’s load-bearing contract and must not be rebased/merged as-is.

- #1548 requires one unconditional `(void)k_b<ID>;` for **every** formal admitted by `core_parameters`, including used parameters, and explicitly forbids a second use analysis. This PR adds `binding_has_use`, emits only for unused bindings, and asserts that used parameters are not discarded.
- The issue requires a lasting `task unused-parameter`, O0+O2 strict builds, every admitted declarator branch, first/middle/last positions, and pointer/callable/aggregate mutation guards. Reusing `unused-function` with Int/O2 cases does not satisfy that matrix or its refusal proof.
- “Changes emitted C of no program that compiles today” is false: current main has used borrowed-Bytes helpers that receive the unconditional Bytes discard; this patch removes those bytes.
- The corrected compiler/Taskfile/new-fixture change necessarily moves the canonical pair measurement basis. Absence of the renamed helper from the old ledgers does not waive remeasurement.

After v0.10.1 THAW and a tracker-visible canonical-pair handoff, reuse this PR only after rewriting it to unconditional discard with no `binding_has_use`, adding the dedicated O0/O2 declarator/mutation gate, regenerating the four pair authority digests, Taskfile/task-help/census, driver/input pins, exact pair ledger and mirror-derived prose, and the Taskfile/manifest release-evidence index digest. The claims manifest/projected claim pages and unresolved-calls ledger should remain unchanged absent a separate justification.

Until the current owner updates or releases its claim, no second implementation owner should overlap it.

PR #1579, opened 2026-08-21
tooling: record the shell verdict data sweep (#1554)

### LINKED ISSUE #1554: A shell verdict returned as an exit status is a silent split between dash and bash
## Metadata

- State: in-progress
- Kind: test-quality
- Area: Tooling / shell harnesses
- Priority: P2
- Size: S
- Blocked by: none
- Audited against: `origin/main@a858e0fa6f0060c4ecc2d0299852f4282014a9b5`

## Problem

**A shell helper that returns a value greater than 1 as *data*, consumed inside
`$( )` under `set -e`, is a silent split between shells.** `dash` kills the
command substitution's subshell the moment the helper returns non-zero;
`bash` and `zsh` do not. CI runs `sh`, which is `dash` on the runner and
`bash` on a typical developer machine.

So the gate passes locally, forever, and fails in CI **on the first day the data
reaches the value that means "undecidable"** — not when the code is written.

This is not hypothetical; it shipped and was found the hard way in
`stdlib/check-capabilities.sh` (PR #1512).

`citation_liveness` answered with an exit status — 0 live, 1 all-closed, 2
undecidable — and `validate_liveness` was called as
`decided=$(validate_liveness "$MATRIX") || { ... }`. The first row whose
citation was newer than the committed backlog snapshot returned 2, `set -e`
killed the substitution before any row was judged, and the gate printed:

```
capabilities: a planned or deferred row cites only closed issues
capabilities: a state that describes finished work reads as "someone is on this"
```

with nothing between the last two lines. **The failure path re-scans to name the
offending row and named none** — because the re-scan reached the same function
through `cmd && continue`, a tested context where `-e` does not apply. A gate
failing on a row its own re-scan cannot find is the signature.

The divergence, measured on this machine:

```sh
set -e; inner() { return 2; }; outer() { inner; printf '14\n'; }
x=$(outer) || printf 'outer-failed\n'; printf 'x=[%s]\n' "$x"
# bash: x=[14]     zsh: x=[14]     dash aborts
```

It needed **both halves** to stay hidden: the shell difference, and a value that
had never occurred because no capability row had ever cited an issue newer than
the snapshot. It survived twelve mutation self-tests and an existing positive
case, all of which reached the function through a tested context.

## Scope

Sweep `tooling/` and the shell harnesses it drives for the same shape:

- a function whose non-zero return is **data** rather than failure;
- consumed inside `$( )`, or as a bare command whose status is read by
  `case $?` / `test $?`;
- in a file that runs under `set -e`.

For each, either return the verdict on **stdout** — which is what #1512 did, and
which removes the shell dependence entirely rather than testing it — or move the
call into a tested context and say why in a comment.

Out of scope: `bin/kofun`, `bootstrap/**`, and `tests/**` outside the harnesses
`tooling/` drives. Those are separate sweeps if this one finds enough to justify
them.

## Acceptance criteria

- [ ] Every occurrence of the shape in `tooling/` is listed, with file and line,
      including the ones judged safe and why.
- [ ] Each unsafe one is fixed by printing the verdict rather than returning it,
      or by a tested context with a comment naming this issue.
- [ ] If the sweep finds an unsafe site, at least one fixed site has a test in
      the **shape of the real call site** — `$(...)` guarded by `||` — proving the
      pass does not abort. A test that reaches the function only through a tested
      context passes on the broken version and is worse than none; #1512's
      `newer_than_snapshot` is the model. If the sweep finds no unsafe site, record
      that result and keep #1512's real-shape test as the reference proof rather
      than introducing a synthetic defect merely to duplicate it.
- [ ] The sweep records what it did **not** cover, so a partial sweep is not read
      as a total one.

## Validation

| --- | --- | --- |
| the swept harnesses | `task forbidden-requirements-census gate-reachability task-help capabilities` | PASS |

## Dependencies

Blocked by: none. Related to #1512, where the instance was found and fixed.

hjosugi:
## Summary

- record the exhaustive #1554 sweep of all five tracked `tooling/**/*.sh` files and both direct caller gates
- show that the sole status-as-data helper, kotest's `run_unit`, is already safe in a POSIX tested context
- list every reviewed near-match and the explicit exclusions, including the #1512 real-call reference proof
- bind the safe call-site comment to #1554 without changing runner behavior

The sweep found zero unsafe sites. The issue acceptance now says that when the result is zero, #1512's existing `$(...) ||` regression remains the reference instead of manufacturing a synthetic defect.

## Validation

- `task forbidden-requirements-census gate-reachability task-help capabilities`
- `sh -n` and Dash 0.5.13.5 `-n` over the seven audited/reference scripts
- actual kotest no-match status-3 path under `/bin/sh` (bash), Dash, and zsh: all print `Tests  0 passed (0 total, 0 suites)` and exit 0
- known reproducer: bash/zsh produce `x=[14]`; Dash produces `outer-failed` and `x=[]`
- `git diff --check`

Full `task verify` remains required on the final rebased head during the exclusive release validation window.

Closes #1554

### DIFF
--- a/tooling/SHELL_VERDICT_AUDIT.md
+++ b/tooling/SHELL_VERDICT_AUDIT.md
@@ -1,0 +1,60 @@
+# Shell verdict audit
+
+This is the #1554 sweep of shell status values that can be confused with
+failure by `set -e`. It was measured at
+`2d3aef9215b4a930827dff997418f7ebceb9cb61`.
+
+## Scope
+
+The sweep covers all five tracked `tooling/**/*.sh` files:
+
+- `tooling/compile-census/check.sh`
+- `tooling/forbidden-requirements/fixtures/mention-and-invoke.sh`
+- `tooling/kotest/run.sh`
+- `tooling/lsp/build-semantic-bundle.sh`
+- `tooling/machine-dependent/check.sh`
+
+It also checks the two shell gates that directly execute those scripts,
+`tests/lsp/check.sh` and `tests/stdlib/kotest/check.sh`. They are callers, not
+harnesses launched by `tooling/`. The kotest runner launches `bin/kofun`, but
+that shell entrypoint is an explicit exclusion in #1554.
+`tests/tooling/verify-object-reuse/check.sh` mentions the LSP builder only as
+text inspected by `grep`, so it is not an execution path.
+
+The sweep does not cover `bin/kofun`, `bootstrap/**`, unrelated `tests/**`, or
+shell files outside these direct paths. The forbidden-requirements file is a
+detector fixture, not a gate; it remains in the inventory so its command
+substitution cannot be mistaken for an omission.
+
+## Status used as data
+
+| producer | status | consumer | verdict |
+| --- | --- | --- | --- |
+| `tooling/kotest/run.sh:318-397` (`run_unit`) | `3` at line 336 means no selected tests; `2` at lines 344, 374, and 382 classifies a build error; the compiled suite's status is captured at lines 385-388 and returned at line 396 | the sole call is `if run_unit ...; then ... else unit_status=$?; fi` at lines 431-435, followed by the status `case` at line 436 | safe: the command is in a POSIX tested context, so `set -e` cannot consume the data status; the adjacent comment names #1554 |
+
+There are no unguarded occurrences and therefore no unsafe site to rewrite.
+
+## Near matches reviewed
+
+| site | why it is not the defect |
+| --- | --- |
+| the four tooling `ROOT=$(...)` initializers | output is path data, but a nonzero status is a real initialization failure |
+| `tooling/lsp/build-semantic-bundle.sh:19-22,48` | Node prints a path and `uname` prints a platform; a nonzero status is failure, not a third verdict |
+| `tooling/forbidden-requirements/fixtures/mention-and-invoke.sh:43-45` | Node prints a version and failure is failure; this file is included here, though the forbidden-requirements production census excludes its detector-fixture directory |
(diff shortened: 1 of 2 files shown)


hjosugi:
## Exhaustiveness audit: BLOCK at `a3eaad26e2b70000a5ca11b1e620c4081cd4a142`

The current sweep is otherwise well reasoned, but it is not exhaustive over #1554’s stated `tooling/` scope and introduces an unrelated measured-claim contradiction.

1. The tree has six tracked POSIX-shell entrypoints under `tooling`, not five `**/*.sh`. The missing extensionless executable is `tooling/lsp/kofun-lsp` (`#!/bin/sh`, `set -eu`, ROOT substitution, `exec node`). It is safe, but must be inventoried. Update five→six and four ROOT substitutions→five. Also include `tests/lsp/visibility.sh` in caller-gate scope: `tests/lsp/check.sh` invokes it and it passes the real wrapper to JS.
2. New `tooling/SHELL_VERDICT_AUDIT.md` falsifies `docs/ROADMAP.md`’s measured row `external security audit | absent`, whose command is `git ls-files *AUDIT*` with no matches. Prefer renaming this document to `tooling/SHELL_VERDICT_SWEEP.md`; widening the ROADMAP probe would unnecessarily touch a release-evidence digest and expand scope.
3. After v0.10.1 THAW, rebase once onto the settled post-#1587/#1580/#1574 main, re-enumerate by tracked shebang rather than extension, refresh the measured-at SHA/counts, and add the new exact `pr-open` claim head.

Revalidate syntax and actual no-match behavior across the six tooling entrypoints plus LSP caller scripts under the available shells; include `task kotest` and preferably `task lsp` with the existing census/reachability/capability/repository/release checks, then quiet verify and exact-head CI. This needs no pair/bootstrap/claim/evidence update if the preferred file rename is used.

Because the PR currently auto-closes #1554 with a live claim, either change `Closes` to `Addresses` for the safe merge/backlog sequence or append terminal `merged` immediately and rerun the same main SHA after state settles.

Issue #1584, opened 2026-08-21
The two halves spell the same decisions separately, and nothing compares them


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation issue
- Kind: test-quality
- Area: Bootstrap / Stage 2 pair
- Priority: P1
- Size: M
- Parent: #1401
- Blocked by: none
- Audited against: `origin/main@36acd513`

## Problem

The two halves of the pair spell the same *decisions* separately — a type set,
a table size, a seed constant, an argument count — and **nothing compares
them**. Six defects of exactly this shape were found in one week.

| # | the decision | how they differed |
| --- | --- | --- |
| #1570 | the annotated-binding type exclusion | C half excluded `List[Int]` via a `list_int` flag; Kofun half omitted the term |
| #1546 | the body symbol counter's seed | C half's running counter absorbed `fail`; Kofun half's literal `count + 17` did not |
| #1537 | the declared-parameter table size | C half stopped at 8; Kofun half walked all of them |
| #1571 | an argument count at one call site | C half passes three; Kofun half passed two |
| #1545 | the selfhost builtin name table | C half's 17-name table omits `to_text`; the Kofun half computes differently |
| #1508 | two callee names | present in one half, defined nowhere |

**Five of the six are literal values sitting in source in both files.** Not
behaviour, not control flow — constants and sets that a human kept in step by
hand and eventually did not.

## Why the three existing gates cannot see them

The stage 2 Kofun half is checked three ways, all static, because **nothing
executes it**: `bootstrap/manifest.json:41` names `bootstrap/stage1/compiler.kofun`
as `canonical_source`, so the selfhost chain never touches stage 2's.

| gate | proves | misses |
| --- | --- | --- |
| `check.sh` round-trip | the text lexes, parses and reprints | every value in it |
| `stage2-pair-calls` (#1509) | every call names something that exists | its arity, and every constant |
| `stage2-pair-mirror` (#1538) | every ledger function has a counterpart | what the counterpart *decides* |

And #1408's undefended-branch ledger cannot help with any of them: the C half
takes its branch, correctly, every time. That is now **six** defects the ledger
could not have surfaced, against one it did (#1537 was found by reading a
function the ledger pointed at).

## What is wanted

A fourth static parity check: **the paired decisions the two halves must
spell identically are enumerated, and a gate fails in both directions.**

The shape that already works here twice is a ledger plus a scan —
[...]

PR #1538, opened 2026-08-20
pair-coverage: which ledger functions the Kofun half mirrors at all (#1514)

hjosugi:
Closes #1514 (child of #1401).

## The finding

**A quarter of #1401's work-list is not what it looks like.**
`tests/pair-coverage/undefended.tsv` lists 1911 untaken branches in 354
functions of `bootstrap/stage2/compiler.c`, and #1401 reads them as places a
divergence between the two halves of the pair can hide. That reading assumes
both halves implement the function.

```sh
python3 - <<'PY'
kofun = open("bootstrap/stage2/compiler.kofun").read()
rows = [l.split("\t") for l in open("tests/pair-coverage/undefended.tsv")
        if l.strip() and not l.startswith("#")]
absent = [r for r in rows if r[2].strip() not in kofun]
print(len(absent), "of", len(rows), "functions;",
      sum(int(r[0]) for r in absent), "of", sum(int(r[0]) for r in rows), "branches")
PY
# 86 of 354 functions; 473 of 1911 untaken branches
```

Reading all 86 splits them, and the split is the point:

| | functions | untaken branches |
| --- | ---: | ---: |
| same work, different name | 64 | 330 |
| genuinely C-only | 22 | 143 |

Plus three whose names occur only in prose, giving **25 C-only functions holding
153 branches — 8% of the ledger.** The other 92% is a real pair surface.

## What this adds

`tests/pair-coverage/mirror.tsv`: one row per ledger function, saying whether
the Kofun half does the work and under what name, and `task stage2-pair-mirror`
to hold it exact.

The renamed 64 are two systematic transliterations the ledger's names cannot
show — C's `sh_*` selfhost-HIR emitter is Kofun's `selfhost_*`, C's `sl_*` C11
lowering is Kofun's `selfhost_c11_*` — plus a family inlined into its caller.
**Every one of those rows carries a mark rather than a resemblance**: an
identical `E2S…` code with identical message text, an identical emitted record
field order, in several places a word-for-word identical doc comment.

The C-only 25 are `compiler.ensure_move` (71 branches), the
`Stage2AuthorityContext` embedder API (53), memo/capacity/allocator plumbing
(21), host-level file and identifier handling (7), and `ends_with` (1), which is
not a cluster but the defect #1508 recorded.

**Three of those clusters are already documented in `compiler.kofun`'s own
prose** — "lives in the C seed only" (`:17128`), "a seed-only capability with no
[...]

PR #1585, opened 2026-08-21
stage2: the Kofun half admits a List[Int] annotation and then refused it (#1570)

### LINKED ISSUE #1570: The Kofun half admits a List[Int] annotation and then refuses the binding as a malformed enum
## Metadata

- State: in-progress
- Kind: implementation
- Area: Bootstrap / Stage 2 pair
- Priority: P1
- Size: S
- Parent: #1401
- Blocked by: none
- Audited against: `main@2d3aef9215b4a930827dff997418f7ebceb9cb61`

## Goal

Make the Kofun half of the pair accept `let xs: List[Int] = …`, which its own
annotation check already admits and its dispatch then refuses.

`compiler.kofun`'s `lower_body` lists `List[Int]` among the annotation types it
accepts:

```
compiler.kofun:14492                if binding_type != "Int" &&
compiler.kofun:14493                   binding_type != "Int?" &&
compiler.kofun:14494                   binding_type != "Text" &&
compiler.kofun:14495                   binding_type != "List[Int]" &&      <- admitted
compiler.kofun:14496                   binding_type != "Decimal" && …
compiler.kofun:14502                    return "error[E2S32]: unknown supported aggregate type `" + …
```

and then omits it from the dispatch's exclusion set sixty lines later:

```
compiler.kofun:14558            } else if binding_type != "Int" &&
compiler.kofun:14559               binding_type != "Text" &&
compiler.kofun:14560               binding_type != "Decimal" &&
compiler.kofun:14561               binding_type != "Float" &&
compiler.kofun:14562               binding_type != "DecimalResult" {       <- List[Int] not excluded
compiler.kofun:14563                if record_declaration_start(source, binding_type) >= 0 { …record… }
compiler.kofun:14613                    if mutable { …"concrete enum bindings are immutable"… }
compiler.kofun:14617                    if token_kind(source, value_start) != "identifier" {
compiler.kofun:14618                        …"concrete enum initializer must name a constructor"…
```

So an annotated `List[Int]` binding falls through the record test and into the
**concrete enum ladder**, where a list literal is not an identifier.

**The C half compiles it**, which is what makes this a divergence rather than a
shared limitation:

```sh
cc -std=c11 -O1 -w bootstrap/stage2/compiler.c -o /tmp/stage2
printf 'fn main() -> Int {\n    let xs: List[Int] = [1, 2, 3]\n    print(len(xs))\n    return 0\n}\n' > /tmp/al.kofun
/tmp/stage2 /tmp/al.kofun /tmp/al.c /tmp/al.ir /tmp/al.tokens
# expected current result: exit 0, and /tmp/al.c contains
#   KofunIntListValue k_b0 = kofun_list_int_value(…{UINT64_C(3), {INT64_C(1), INT64_C(2), INT64_C(3)}});
```

The C half keeps the two checks in step with a `list_int` flag
(`compiler.c:21384-21395`), so the annotated binding never reaches the enum
ladder at all.

**Three sub-forms, all divergent:**

| written | C half | Kofun half |
| --- | --- | --- |
| `let xs: List[Int] = [1,2,3]` | emits the list value | `E2S32 concrete enum initializer must name a constructor` |
| `let mut xs: List[Int] = [1,2,3]` | emits the list value — `E2S144` explicitly exempts `List[Int]` | `E2S32 concrete enum bindings are immutable in this Core slice` |
| `let a: List[Int] = make_list()` | `KofunIntListValue k_b1 = kofun_fn_make_list();` | routed into `emit_enum_value(…, "List[Int]")` — an enum carrier for a list |

## How much source this covers

```sh
git grep -c 'let \(mut \)\?[a-z_]*: List\[Int\] *=' -- '*.kofun' | awk -F: '{n+=$2} END {print n" in "NR" files"}'
# expected current result: 122 in 43 files
```

Including `bootstrap/native/encoder.kofun`. **Every one of them would be refused
by the mirror**, which is the sharpest statement of what "the Kofun half is
never executed" currently costs: a hundred and twenty-two bindings in the tree
depend on a rule only one half implements.

## Why nothing catches it

The stage 2 Kofun half is neither `canonical_source` nor `trusted_seed` in
`bootstrap/manifest.json`, so nothing compiles it and nothing runs the result.
The C half is right, so every gate that reads emitted C is green. And #1408's
undefended-branch ledger cannot help — the branch is *taken*, by the C half,
correctly.

## Scope

In scope: the missing `List[Int]` term in the dispatch exclusion, and whatever
the enum/record ladder needs so the three forms above lower the same way in both
halves.

Out of scope: `Int?`, which is also absent from the second condition but is
handled by a dedicated branch before it — worth a look while in there, and not
this issue's claim.

## Acceptance criteria

- [ ] `let xs: List[Int] = [1, 2, 3]`, its `mut` form, and the call form lower
      the same way in both halves.
- [ ] The two conditions that must agree are derived from one statement, or a
      gate compares them; a second list of type names is how this happened.
- [ ] `task list-int-values` and `task stage2` pass.

## Validation

| --- | --- | --- |
| the three forms | `task list-int-values` | PASS |
| the pair is synchronized | `task stage2 && task selfhost-fixed-point` | PASS |

## Serialization

This changes the canonical Stage 2 pair. Its owner must announce before claiming
and must not run pair gates concurrently with another pair change.

## Dependencies

Blocked by: none.

Found while classifying `lower_body` for #1401.

hjosugi:
Closes #1570.

**Not for merge until `kofun-5f` announces THAW.** The release is mid-tag.

## One term

`lower_body`'s annotation check admits `List[Int]`; the dispatch sixty lines
below omitted it from the exclusion set. So an annotated list binding fell past
the record test into the **concrete-enum ladder**:

| written | C half | Kofun half, before |
| --- | --- | --- |
| `let xs: List[Int] = [1, 2, 3]` | emits the list value | `E2S32 concrete enum initializer must name a constructor` |
| `let mut xs: List[Int] = …` | emits the list value | `E2S32 concrete enum bindings are immutable` |
| `let a: List[Int] = f()` | `KofunIntListValue k_b1 = kofun_fn_f();` | `emit_enum_value(…, "List[Int]")` — an enum carrier for a list |

Excluding the type routes the annotated form into the general initializer path,
which **already lowers `List[Int]` correctly** — it calls `emit_list_int_value`
and gives the binding a `KofunIntListValue` carrier. That is the same
destination the C half reaches via its `list_int` flag
(`compiler.c:21463-21467`). The enum/record ladder needs nothing.

**122 bindings in 43 tracked files** are written this way, including
`bootstrap/native/encoder.kofun`.

## No fixture, and the measurement is the reason

```sh
grep -rl 'let \(mut \)\?[a-z_]*: List\[Int\] *=' tests/ --include='*.kofun' | wc -l
# 34
```

Thirty-four fixture files already use the annotated form and between them cover
all three sub-forms — `let mut copied: List[Int] = returned.samples` and
`let empty_again: List[Int] = identity(empty)` among them. They run through the
**C half**, which was already right, so a thirty-fifth would pass before this
change and after it. That is a fixture that cannot fail.

## And no fixture *can* exercise this half

`bootstrap/manifest.json:41` names **stage 1's** Kofun source as
`canonical_source`. The selfhost chain `S → C1 → A1 → C2 → A2` never touches
stage 2's. That file is checked three ways, all static, and none of them
evaluates a condition:

| gate | proves |
| --- | --- |
| `check.sh:592` round-trip | the text lexes, parses and reprints |
| `stage2-pair-calls` (#1509) | every call names something that exists |
| `stage2-pair-mirror` (#1538) | every ledger function has a counterpart |

So "both halves lower this the same way" is a claim about two texts. The gate
that would hold them together is filed as **#1584** — six defects this week
share the shape of *a decision the two halves spell separately*.

## Verification

`task check`, `task stage2`, `task stage2-pair-calls`, `task stage2-pair-mirror`,
`task list-int-signatures`, `task list-int-values` — all green. The last two
confirm the C half is untouched, which it is: only `compiler.kofun` changed
apart from the digests.

`preflight` clean apart from the pre-existing `inputs.tsv` drift (#1408).

### DIFF
--- a/bootstrap/stage2/SHA256SUMS
+++ b/bootstrap/stage2/SHA256SUMS
@@ -1,2 +1,2 @@
-02a60013c862c7359a708db5718e415c2f979aa582aa52211127d27e9e4d175d  bootstrap/stage2/compiler.kofun
+5ae6d8585c892e792e43f4abf51b5458c554148783e50b901ece96f2db9888b6  bootstrap/stage2/compiler.kofun
 da353011417529ea8e3c6c4de1cec219f64b90fb7c54d15b6255414bc2644f77  bootstrap/stage2/compiler.c
--- a/bootstrap/stage2/compiler.kofun
+++ b/bootstrap/stage2/compiler.kofun
@@ -14621,8 +14621,30 @@ fn lower_body(
                 " = " + value + ";\n" +
                 "    if (kofun_failed) return " + failure_result + ";\n"
                 cursor = skip_trivia(source, value_end)
+            # #1570. `List[Int]` belongs in this exclusion, and its absence
+            # was a divergence rather than a limitation: the annotation check
+            # sixty lines above admits it, and the C half keeps the two in step
+            # with a `list_int` flag that short-circuits this whole ladder
+            # (`compiler.c:21463-21467`). Without the term here an annotated
+            # list binding fell past the record test into the concrete-enum
+            # ladder, where a list literal is not an identifier -- so this half
+            # refused `let xs: List[Int] = [1, 2, 3]` as
+            # `concrete enum initializer must name a constructor`, and its
+            # `mut` form as `concrete enum bindings are immutable`.
+            #
+            # Excluding it routes the annotated form into the general
+            # initializer path below, which already lowers `List[Int]`
+            # correctly: it calls `emit_list_int_value` and gives the binding a
+            # `KofunIntListValue` carrier. That is the same destination the
+            # C half reaches, by the same reasoning, which is why one term is
+            # the whole fix.
+            #
+            # 122 bindings in 43 tracked files are written this way, including
+            # `bootstrap/native/encoder.kofun`. Every one would have been
+            # refused by this half.
             } else if binding_type != "Int" &&
                binding_type != "Text" &&
+               binding_type != "List[Int]" &&
                binding_type != "Decimal" &&
                binding_type != "Float" &&
                binding_type != "DecimalResult" {


hjosugi:
## Exact-head review at `c4f87bde`: BLOCK on #1570 AC2

The functional change is correct: the new `binding_type != "List[Int]"` term routes all three annotated-list forms into the existing `emit_list_int_value` / `KofunIntListValue` path. The committed Kofun digest also matches exactly, and no manifest or release-evidence update belongs to this patch.

The lasting proof is missing. #1570 requires the annotation-admission and enum-exclusion decisions to be derived from one statement **or compared by a gate**. This PR does neither. Reverting the new term leaves every existing task/fixture green because all 34 fixtures exercise only the already-correct C half. The PR's non-executability explanation is valid, but it makes a static defect-sensitive proof more important; it does not waive the checkbox.

#1584 is still `needs-detail` with no gate or implementation, so filing it cannot satisfy the current `Closes #1570`. A narrow fix fits existing infrastructure: extend `tests/stage2/list-int-signatures/check-pair.mjs` / `task list-int-signatures` with semantic anchors for the C `list_int` short-circuit and Kofun exclusion, plus a mutation self-test that removes the new last `List[Int]` term and requires a named failure. That script already has semantic anchors, `replaceLast`, and mutation self-tests.

Also synchronize the canonical claim from `active` to `pr-open` with this exact PR/head. Because this takes the pair, either explicitly park `c4f87bde` with no edit/rebase/pair gates until the release queue hands it back, or enter a tracker-visible serial order. Do not merge it into the current release corrective queue merely to clear the resource; #1570 is not a tag blocker. After handback, rebase once, add the lasting proof, regenerate pair SHA once, and require new exact-head CI.

hjosugi:
## Exact-head static-proof audit: BLOCK at `dd55714bf00ba910be5412ae25d1643a9b49aab8`

The one-term Kofun fix is functionally correct, but #1570 AC2 remains unproved. Current green gates are not defect-sensitive to this edit: deleting the new last `binding_type != "List[Int]" &&` term still leaves the existing fixtures, call ledger, mirror, round-trip, and CI green.

Before merge after v0.10.1 THAW:

1. Extend `tests/stage2/list-int-signatures/check-pair.mjs` with uniquely contextual semantic anchors for the Kofun annotation admission, the Kofun enum/record exclusion including List[Int], and the matching C `declared_type == List[Int]` / `list_int=true` / `!list_int` short-circuit. Generic token counts are insufficient.
2. Add a named mutation self-test using the existing exact `replaceLast` mechanism to remove the target second Kofun occurrence and require refusal. Prefer a symmetric C mutation as well because the check publishes a two-half decision. No new `.kofun` fixture is needed.
3. Rebase once onto the actual post-release main, refresh any stale count/line prose, rerun the focused list-int/stage2/pair-call/mirror/fixed-point gates, quiet verify, and exact-head CI. #1584 may remain the broader six-decision follow-up; it need not be completed for this narrow proof.
4. This changes tracked `compiler.kofun`, so the c005 pair measurement cannot authorize the rebased head even though `inputs.tsv` already names that file. Serialize with the canonical pair queue and measure once on the final integrated pair basis.
5. Change `Closes #1570` to `Addresses #1570` until merge/main CI, then append terminal `merged` for the existing owner and close manually.

No `compiler.c`, manifest, release claim, or generated release-evidence edit is required for this PR itself; recompute the pair SHA after final rebase and rerun the release pack as validation.

hjosugi:
## Handing off — this is ready to merge on green, nothing else is needed

My session is ending. State of this PR, so whoever picks it up does not
reconstruct it:

- **Rebased onto `4e66b21c`** — `main` after #1595 repaired the
  `stage2-pair-mirror` breakage. No conflicts.
- **Green locally** before the push, and
  `sh tests/preflight/check.sh` reports
  **`PASS: every structural obligation is satisfied`** — including the pinned
  input set, which I had missed on my previous PR and which is what turned
  `main` red.
- **CI running.** Its earlier failure was `stage2-pair-mirror` only, caused by
  #1548's rename outliving a ledger row — not by anything here. Confirmed from
  the run log, not assumed.

**If CI is green: merge it, then set the claim on its issue to `merged`.**

**If CI is red: check whether another PR on `main` broke it before suspecting
this change.** That was the cause twice today. The terminal line to look for is
`task: Failed to run task "<name>"` — if the failing task is one this PR does
not touch, it is almost certainly inherited.

Worktree and branch cleanup after merge:

```sh
git worktree remove --force <path> && git branch -D <branch>
git push origin --delete <branch>
```

hjosugi:
**Correction to my handoff comment above: `main` has moved and this needs a rebase before merging.**

When I wrote that comment `main` was `4e66b21c` and this branch was rebased onto it. Since then two commits landed — #1595 (the `stage2-pair-mirror` repair) and #1593 (the pair-coverage re-measurement) — and `main` is now `3a2a87b4`.

So the instruction "merge on green" is no longer sufficient on its own:

1. Rebase onto `3a2a87b4`.
2. Regenerate the derived files rather than resolving them by hand — `bootstrap/stage2/SHA256SUMS`, `trusted_seed_sha256` in `bootstrap/manifest.json`, and `artifacts/release-evidence/index.json`. The recipe is in the handoff; `git checkout --ours` on those three, then `task release-evidence`.
3. Re-run `sh tests/preflight/check.sh`. It reported `PASS: every structural obligation is satisfied` against `4e66b21c`; that result does not carry across a rebase, because #1593 rewrote the pair-coverage ledgers this PR's basis is checked against.
4. Then merge on green.

**The reason I am flagging it rather than leaving the earlier comment to stand:** a stale "ready to merge on green" is worse than no instruction, because it reads as a decision someone already made carefully. It was careful when written and it is stale now, and the difference is not visible from the comment itself.

The change in this PR is unaffected — nothing in #1595 or #1593 touches what it fixes.

hjosugi:
**Retracting my previous comment: this does NOT need a rebase. Ready to merge on green, as originally stated.**

I said `main` moving to `3a2a87b4` invalidated the `preflight PASS` here. That was wrong, and wrong in the direction that costs someone work. Measured rather than assumed, by kofun-60 and then independently by me:

```
git merge-tree --write-tree origin/main <branch>   → clean, no conflicts
merged inputs.tsv: 1299 lines − 11 comments = 1288 pinned
actual *.kofun in merged tree                      = 1288
present-but-unpinned                               = (empty)
typed_return_owner.kofun pinned                    = exactly 1
stage2-pair-mirror  359 / 359
stage2-pair-calls   543 / 544, 5545 arities
```

**The reason is the rule this PR's sibling incident produced.** `tests/pair-coverage/{undefended,mirror}.tsv` are keyed on `compiler.c` function *names*. Neither of my PRs adds or removes a function in either half — `compiler.c` here is +33/−4 entirely inside existing functions, and `compiler.kofun` has zero `fn` additions or removals:

```
compiler.c   added/removed function definitions: 0
compiler.kofun added/removed fn:                 0
```

So there is no row that can go stale. #1593 rewrote the ledgers' **contents**; the keys are what a PR can collide with, and those are untouched.

**Why I am writing this out rather than quietly deleting the previous comment.** I reasoned "#1593 rewrote the ledgers this PR is checked against, therefore the PASS doesn't survive" — plausible, and I posted it without measuring. A stale go-ahead is worse than no instruction; so is a false stop, and a false stop is the one that looks responsible. Both comments stay up so the next reader can see which one was measured.

**Merge on green. No rebase, no regeneration.**

hjosugi:
**Rebased onto `cf1999c2` and force-pushed. #1575 is merged; this was the last one.**

My previous comment said "no rebase, no regeneration". That was measured and true against `3a2a87b4` — and it stopped being true the moment **I merged #1575**, which moved `main` under this branch. Third time today an instruction was accurate when written and stale within the hour, and this time I caused the staleness myself.

**One conflict, in a derived file, resolved by regeneration rather than by taking a side:**

```
bootstrap/stage2/SHA256SUMS   (the only conflicting path)
  ours   (main, with #1575)  compiler.kofun 02a60013…
  theirs (this branch)       compiler.kofun 942711ec…
  correct value              compiler.kofun 5ae6d858…
```

**Neither side's digest was right**, which is the whole argument for regenerating: this branch's `compiler.kofun` change on top of #1575's makes a file whose hash neither parent ever saw. `compiler.c` is unchanged here, so `da353011…` carries through and `trusted_seed_sha256` in `bootstrap/manifest.json` needed no edit — verified rather than assumed. `bin/kofun-digest` cross-checked against `sha256sum`.

**Verified after rebase, before push:**

```
task release-evidence        PASS  (pack already current, no diff)
sh tests/preflight/check.sh  PASS: every structural obligation is satisfied
                                   pinned input set matches the tree (1288 files)
task stage2-pair-mirror      PASS  6 of 6
task stage2-pair-calls       PASS  7 of 7
task list-int-signatures     PASS
task list-int-values         PASS
```

Merging on green.

PR #1595, opened 2026-08-21
pair-coverage: main is red — a ledger row outlived the function it named (#1548, #1408)

hjosugi:
## `main` is red and this makes it green

CI has been failing on `main` since **8fefeb40** (run at 12:03Z). `stage2-pair-mirror` is in `verify`, and it refuses:

```
FAIL: pair mirror: tests/pair-coverage/mirror.tsv no longer describes the tree:
    bytes_parameter_discards is same-name and compiler.kofun defines no such fn
```

**Nothing is wrong with the compiler pair.** #1548 renamed `bytes_parameter_discards` → `unused_parameter_discards`, generalising it past Bytes, and added `binding_has_use`. Both are defined in *both* halves. What went stale is the record of them.

## Why it happened, since the same shape will recur

| when | what |
| --- | --- |
| `c005fc31` | ledger measured — `bytes_parameter_discards` exists |
| `83e2be26` | ledger refreshed from that measurement |
| #1572 merges `8c65c088` | #1548 renames the function |

#1548 was green against a `main` that **predated** the refresh. Neither change is wrong on its own; merging the second broke the first. A PR is tested against the `main` it forked from, not the one it lands on — and the pair ledger is a derived artifact that any `compiler.c` rename can invalidate from a distance.

## Three edits. No measurement.

- **`undefended.tsv`** — the row `4  4  bytes_parameter_discards` is removed. A row naming a function `compiler.c` does not have is not an unwatched function, it is a stale record. `mirror.tsv` must carry one verdict per ledger row, which is what turned this from a quiet inaccuracy into a red gate.
- **`mirror.tsv`** — its verdict goes with it. 361 → 360 rows. The 24 C-only and 67 other-name counts are unchanged, because the removed row was `same-name`.
- **`inputs.tsv`** — pins `tests/stage2/unused-function/unused_parameter.kofun`, which #1548 added and never pinned. `measure.sh` refuses to start against an unpinned tree, so the ledger was **unmeasurable** as well as red.

## The measured totals are left alone

`1963 / 1873 / 361` is what that run found, and it stays what that run found. A new header block states instead that:

- the rows now sum to **1959 / 1869 / 360** — verified by summing them, not asserted;
- the four-branch difference is **exactly the deleted row**;
- `unused_parameter_discards` and `binding_has_use` have **no row at all** until someone measures again.

That last point matters for whoever runs the next measurement: the totals will **rise**, the gate reports a rise as a regression, and that is the safe direction. Naming the gap in the file beats papering over it with an estimate — and beats letting the next measurer argue with the gate about a number nobody wrote down.

## Verification

| check | result |
| --- | --- |
| `stage2-pair-mirror` | **PASS** — 360 of 360 verdicts; 6 of 6 proof cases |
| `stage2-pair-calls` | **PASS** — 543 of 544 targets, 5545 arities, 7 of 7 proof cases |
| `inputs.tsv` vs `git ls-files '*.kofun'` | exact, both directions |
| before the fix, on clean `main` | reproduces the failure above |

## Note for reviewers

This is deliberately the *minimum* that makes `main` green, split out of #1593 so it can land immediately. #1593 re-measures the ledger properly and will be rebased on top of this.


hjosugi:
**I broke this. #1548 is mine, the rename is mine, and the unpinned fixture is
mine.** Reviewed independently against `main` rather than taking the report:

```sh
git show origin/main:tests/pair-coverage/mirror.tsv | grep -c bytes_parameter_discards   # 1
git show origin/main:bootstrap/stage2/compiler.kofun | grep -c '^fn bytes_parameter_discards'   # 0
git show origin/main:bootstrap/stage2/compiler.kofun | grep -cE '^fn (unused_parameter_discards|binding_has_use)'   # 2
git show origin/main:tests/pair-coverage/inputs.tsv | grep -c unused_parameter.kofun   # 0
```

So: the row names a function that no longer exists, both new functions are
present in both halves, and the fixture I added was never pinned. Exactly as
described, and the pair itself is fine.

**The fix is right and I would not change a line of it.**

- The stale row is **deleted**, not re-measured. A row naming a function
  `compiler.c` does not have is a stale record, not an unwatched function — and
  deleting it is the only honest edit available without re-running.
- The header totals are **left as measured**, with a `POST-MEASUREMENT
  CORRECTION` block saying the rows now sum to 1959/1869/360, that the
  difference is precisely the removed row, and that nothing was re-measured to
  produce it. That is the distinction that keeps the ledger meaning what it
  says: measured numbers stay measured, and an arithmetic consequence of a
  deletion is labelled as one.
- The two new functions get **no row**, deliberately, so the next measurement's
  totals rise for a reason that is not a regression — and the block says so in
  advance. That sentence is worth more than the fix, because a totals increase
  with no explanation is exactly how a ledger stops being trusted.

**On the incident shape, since you asked whether it earns a rule in
`docs/CONCURRENT_AGENTS.md`: yes, and it is distinct from the seven.**

The others are all *two agents acting at once*. This one is **one agent acting
correctly against a `main` that changed underneath it**:

| `c005fc31` | ledger measured — old name exists |
| `83e2be26` | ledger refreshed from that measurement |
| `8c65c088` | #1548 renames the function |

Both changes were correct and green in isolation. **A PR is tested against the
`main` it forked from, not the one it lands on**, and a derived artifact keyed
on identifiers in `compiler.c` can be invalidated from a distance by a rename
that never touches it. No amount of coordination between sessions prevents that
— it needs either a gate that fails the *renaming* PR, or the knowledge that
renaming a `compiler.c` function is a ledger edit.

The second is the cheap one and it belongs beside the four unowned resources
already listed: **`tests/pair-coverage/{undefended,mirror}.tsv` are keyed on
`compiler.c` function names — renaming one is an edit to both.** I did not know
that, and the list I helped write did not say it.

**Merging as soon as CI clears**, and not waiting for you. Thank you for
splitting it out of #1593 so it could land immediately rather than behind a
measurement.

PR #1604, opened 2026-08-22
digest: the last call site still reaching for GNU sha256sum (#1213)

hjosugi:
Found while emptying the last three branches after `v0.11.0-seed`. One of them,
`recovered/native-sha256-1213`, is a recovered never-landed second
implementation of #1213; checking whether it still held anything `main` lacked
is what turned this up. It did not — `main` had landed the better
implementation — but **neither of them finished the job**.

## The defect

```
.github/workflows/native-hosts.yml:87
actual=$(sha256sum "$IMAGE" | awk '{ print $1 }')
```

#1213 replaced GNU `sha256sum` with `bin/kofun-digest`, and the tree states
that as done in four places: the census header, `check.mjs`,
`package/manager.sh`, and `tests/digest/check.sh`'s own opening comment, which
calls `sha256sum` "a third [implementation]" in the past tense.

This call site is not an incidental digest. It decides whether each of the six
emitted native images matches `bootstrap/native/SHA256SUMS`, and its verdict
becomes the native-host evidence a release binds — so the evidence anchoring a
published claim was accepted or rejected by a tool this project does not ship,
does not test, and whose absence outside a GNU userland #1213 exists to
survive.

## The fix, and the reason there is a second half

- The call site runs `./bin/kofun-digest`, which builds from
  `bootstrap/stage2/sha256.c` with the C11 compiler the bootstrap already
  requires and prints GNU's exact output format — the `awk` beside it and every
  committed digest are untouched. Both matrix runners are GitHub-hosted
  `ubuntu-24.04` images with a C compiler.
- `task digest` now scans the tree for the case. **Nothing was looking**, which
  is the only reason one call site outlived the migration.

The scan reuses the census's `commandPosition` rather than growing a second
copy of that judgement: it already knows `$(sha256sum` is an invocation while
`SHA256SUMS` and a name containing the token are not, and #1535 widened it for
wrapper `--` terminators last week. Comments are stripped first — roughly half
this tree's occurrences of the string are prose explaining #1213, including the
new file's own header, and a rule that flagged those would be silenced within
the day.

## It refuses the case it exists for

Both directions are proven before it reports on the tree: three invocation
shapes it must catch, three mentions it must not, and a refusal to pass if the
scan reached no files at all. Against the exact defect, reintroduced:

```
FAIL: digest: .github/workflows/native-hosts.yml invokes GNU `sha256sum`
  ($(sha256sum); compute the digest with `bin/kofun-digest`, which prints the
  same format (#1213)
FAIL: digest: a call site still invokes GNU sha256sum
$ echo $?
1
```

and after the fix:

```
PASS: the sha256sum scan catches 3 invocation shapes and refuses 3 mentions
PASS: none of 436 runnable tracked files invokes GNU `sha256sum`
```

## Bookkeeping this change owed

- `census.tsv` regenerated with `--count`: `tests/digest/check.sh` now invokes
  `node` 5 times, not 3. Self-test PASS, both directions.
- The evidence pack's digest for `native-hosts.yml` rebound.
- No new `drivers.tsv` pin: this extends `task digest`, which is already in
  `verify` and already pinned.

## Validation

- `task digest` PASS · `task gate-reachability` PASS (167 of 185 reachable, 18
  recorded) · `node tooling/forbidden-requirements/self-test.mjs` PASS ·
  `sh tests/release/check-claims.sh` PASS
- full local `task verify` running; CI at the exact head is the gate

### DIFF
--- a/.github/workflows/native-hosts.yml
+++ b/.github/workflows/native-hosts.yml
@@ -84,7 +84,13 @@ jobs:
           expected=$(awk -v path="$CHECKSUM_PATH" '$2 == path { print $1 }' \
             bootstrap/native/SHA256SUMS)
           test -n "$expected"
-          actual=$(sha256sum "$IMAGE" | awk '{ print $1 }')
+          # The repository's own SHA-256, not the host's (#1213). This
+          # comparison decides whether the six-host evidence a release binds is
+          # accepted, so it must not be settled by a tool the project neither
+          # ships nor tests. `bin/kofun-digest` builds from
+          # `bootstrap/stage2/sha256.c` with the C11 compiler the bootstrap
+          # already requires and prints GNU's format, so the `awk` is unchanged.
+          actual=$(./bin/kofun-digest "$IMAGE" | awk '{ print $1 }')
           test "$actual" = "$expected"
           chmod +x "$IMAGE"
           set +e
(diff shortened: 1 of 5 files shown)


hjosugi:
Superseded by #1607, same branch under an accurate name.

Renaming the head branch to `agent/native-host-digest-tools-1213` closed this PR instead of retargeting it, so #1607 carries the work. It is this change plus the half that was missing here: the macOS lane of the same workflow verified its two images with `shasum -a 256` 130 lines below the `sha256sum` this PR fixed, and the gate only refused the GNU spelling. Both lanes now use `bin/kofun-digest`, and the scan refuses all three spellings `package/manager.sh` records having tried before #1213 replaced that chain. The Windows lane's `Get-FileHash` is deliberately out of scope and is filed as #1606.

Nothing was reviewed here, so nothing is lost with the number.

PR #1608, opened 2026-08-23
census: one definition of which files can invoke something (#1213)

hjosugi:
A cleanup pass over what #1607 just landed, before it has time to become
load-bearing.

## What was wrong with it

`tests/digest/no-host-digest-tools.mjs` shipped with its own copy of the
census's tree walk — `git ls-files`, skip the fixture directory, keep what has
a runnable extension or a shebang. Reviewing it against `check.mjs` a day later,
the copy had already lost a branch:

```js
if (!named && body.indexOf(String.fromCharCode(0)) !== -1) continue
```

That line is not decoration. `check.mjs` carries a measurement beside it: a NUL
rules out a *candidate* discovered by shebang sniffing, which has to read every
tracked file including binary fixtures, but it must not rule out a file whose
extension already says it is a script — `tests/interop/bindgen-c/check-report.mjs`
embeds four NUL bytes in a fixture string, and skipping it dropped a real
Node.js source file from the census.

So two scanners asking the same question in the same words had already stopped
agreeing about which files the tree contains, one week after the second one was
written. That is the argument for this change; the line count is incidental.

## What changed

`tooling/forbidden-requirements/universe.mjs` now holds `trackedFiles`,
`EXCLUDED`, `RUNNABLE`, and `universe()`, and both callers import it. The digest
gate loses 40 lines and gains the NUL rule. `check.mjs` loses the definition and
keeps everything else.

**Not `detect.mjs`**, which would have been the shorter import. Its own header
says it is separated from the checker so the self-test can drive the detectors
over fixtures *without scanning the repository*, and a module that shells out to
`git ls-files` on import takes that property away. The prose form of the set
stays there as `FILE_SET`, beside the per-detector predicates and matching its
sibling in `tooling/machine-dependent/`, because `--predicates` prints them
together — a count is only meaningful with its predicate attached.
`universe.mjs` records where it went and why.

## What was deliberately left alone

- `tests/rfc/validate-registry.mjs` and `tests/release/validate-claims.mjs` each
  define a three-line `trackedFiles()`. Neither wants this file set — no shebang
  sniffing, no fixture exclusion — and a shared name meaning a different set in
  each caller is worse than three short functions.
- `tooling/machine-dependent/` keeps its own `FILE_SET` over `*.sh *.mjs *.js`:
  a different question, correctly asked separately.
- Precompiling the three `commandPosition` regexes once instead of per file. The
  gate runs in 0.20s over 438 files; the change would buy milliseconds and cost
  a cache to keep honest.
- The altitude question this raises and does not answer: the tree now has two
  mechanisms for "a tool the build must not require" — the contract's
  `forbidden_core_build_requirements`, censused in both directions, and this
  gate. Folding `sha256sum` and `shasum` into the contract would delete the gate
  entirely, but `forbidden_core_build_requirements` is normative in RFC-0018, so
  that is an RFC amendment and a decision, not a refactor. Worth filing if it
  looks right.

## Validation

- `task digest` PASS — 5 invocation shapes caught, 5 mentions refused, tree
  clean at 438 files rather than 437, which is `universe.mjs` entering the set
  it defines
- census checker PASS on 675 rows with the new `node source` row;
  `node tooling/forbidden-requirements/self-test.mjs` PASS, including the
  fixture-exclusion assertion that keeps the one deliberate hole from widening
- `task gate-reachability` PASS · full local `task verify` in flight, reported
  below

### DIFF
--- a/tests/digest/no-host-digest-tools.mjs
+++ b/tests/digest/no-host-digest-tools.mjs
@@ -22,11 +22,15 @@
  * dgst`: the digest subcommand is the forbidden one, and refusing every use of
  * the tool would be a rule about the wrong thing.
  *
- * The detector is the census's `commandPosition`, reused rather than
- * re-written: it already knows that `$(sha256sum` is an invocation and that
- * `SHA256SUMS` and `nodejs-flavour`-shaped tokens are not, and #1535 widened
- * it for wrapper `--` terminators. A second, private copy of that judgement
- * would drift from it.
+ * Both halves of the scan are the census's, reused rather than re-written. The
+ * detector is `commandPosition`, which already knows that `$(sha256sum` is an
+ * invocation and that `SHA256SUMS` and `nodejs-flavour`-shaped tokens are not,
+ * and which #1535 widened for wrapper `--` terminators; the file set is
+ * `universe()`. This gate first shipped with its own copy of the second one,
+ * and the copy was already wrong within the week -- it had dropped the NUL rule
+ * that keeps a binary fixture from being sniffed as a script, so the two
+ * scanners disagreed about which files the tree even contains. A private copy
+ * of either judgement drifts from it.
  *
  * One digest in the tree is deliberately outside this rule, and it is filed
  * rather than left silent: the Windows lane of the same workflow computes its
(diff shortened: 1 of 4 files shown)


hjosugi:
Full local `task verify` on this branch: **exit 0**, 1627 PASS lines, no FAIL — the same count as the run on #1607's tree, which is the expected result for a change that moves a definition without changing what it decides.

CI at this head is green as well (Kofun verification 14m12s, native host evidence binding, release evidence pack).

The one measurable difference is the file count the digest gate reports: 437 → 438, which is `universe.mjs` itself entering the set it defines.

PR #1609, opened 2026-08-23
release: the procedure's blocks cannot report PASS while failing (#1603)

### LINKED ISSUE #1603: The release procedure's verification blocks report PASS when their tests fail
## Metadata

- State: ready
- Kind: documentation / test-quality
- Area: Release procedure
- Priority: P2
- Size: S
- Blocked by: none.
- Audited against: `main@41251110cffb36abed1c61a13a31ddae798de441`

## Goal

Make every verification block in `docs/RELEASING.md` fail closed no matter what
shell it is pasted into, so a failed check cannot print `PASS`.

Measured while cutting `v0.11.0-seed` from `main@41251110`, and it is what
found this: step 7's block printed

```
PASS: remote tag v0.11.0-seed = 41251110cffb36abed1c61a13a31ddae798de441
```

while **no such tag existed on the remote**. Two independent causes stacked:

1. **`set -eu` was not in effect.** The block ran in a non-interactive zsh
   invoked through an `eval` wrapper. Reproduced directly:

   ```
   $ set -eu; false; echo "REACHED-AFTER-FALSE"
   REACHED-AFTER-FALSE
   ```

   Every `test` in the block after the failure was therefore advisory, and the
   final `printf` ran regardless.

2. **zsh read `:r` as a history modifier.** The refspec
   `"refs/tags/$tag:refs/tags/$tag"` expanded to
   `refs/tags/v0.11efs/tags/v0.11.0-seed`, so `git push` failed with
   `src refspec ... does not match any` — the failure the block then covered.

Confirmed by querying the remote directly rather than believing the output:

```
$ git ls-remote release-target 'refs/tags/v0.11*'
(no output)
```

The tag was then pushed under POSIX `sh` and every step 6 and 7 assertion
re-run under `sh` with `set -eu`, exiting 0. The published release is correct;
what is wrong is that the procedure could not tell me it had failed.

## Why this matters more than a shell footgun

`docs/RELEASING.md` already says to run the whole procedure in one POSIX shell
beginning with `set -eu`, so the operator error is real and is mine. But the
document is otherwise written entirely against false greens — step 1 warns in
its own words that a pipeline ending in `tail` reports `tail`'s status and
makes a failing run read as green. This is that same defect inside the file
that warns about it: the penalty for not following its shell instruction is a
convincing `PASS` rather than an error.

## Definition of done

- Each verification block ends by proving its own precondition, e.g. a leading
  guard that exits nonzero when `set -e` is not active, or `test ... ||
  { printf 'FAIL: ...' >&2; exit 1; }` at every assertion so the block does not
  depend on `errexit` at all.
- No block interpolates a variable immediately before `:`; use `${tag}` so a
  `:r`-style modifier cannot be parsed out of it.
- A self-test that runs the blocks' guard under a shell with `errexit`
  disabled and requires a nonzero exit.

## Out of scope

Changing the procedure's steps or their order. This is about the blocks
reporting truthfully, not about what they check.

hjosugi:
Closes #1603.

## The defect, re-measured before fixing it

Cutting `v0.11.0-seed`, step 7's block printed

```
PASS: remote tag v0.11.0-seed = 41251110cffb36abed1c61a13a31ddae798de441
```

while **no such tag existed on the remote**. `docs/RELEASING.md` is written
entirely against false greens — step 1 warns in its own words that a pipeline
ending in `tail` reports `tail`'s status and makes a failing run read as green —
so this was that defect inside the file that warns about it.

**The `:r` half reproduces exactly.** Non-interactive zsh, on this machine:

```
$ zsh -c 'tag=v0.11.0-seed; printf "%s\n" "refs/tags/$tag:refs/tags/$tag"'
refs/tags/v0.11efs/tags/v0.11.0-seed
$ zsh -c 'tag=v0.11.0-seed; printf "%s\n" "refs/tags/${tag}:refs/tags/${tag}"'
refs/tags/v0.11.0-seed:refs/tags/v0.11.0-seed
```

So `git push` was handed a refspec naming a tag nobody created, and the failure
it returned was covered by the advisory checks around it.

**The errexit half does not reproduce as written, and that changed the fix.**
`set -eu; false; echo REACHED` exits at `false` in sh, bash and zsh here —
non-interactively, interactively (`-i`), through `eval`, and through `-c`. What
that rules out is a shell that honours the flag and runs on anyway. What it
leaves is a wrapper starting **a fresh shell per command**, so the `set -eu`
typed once at the top of the procedure was never in the shell where the check
ran. That is why the guard goes at the top of *every block* rather than once at
the top of the procedure.

## The fix

**`release/fail-closed.sh`** — sourced, not run, because it has to inspect and
exit the shell the procedure is actually in. It reads `$-` rather than probing
behaviour, and that is not a shortcut:

> `( set -e; false; true )` returns **0 in all six configurations measured** —
> sh, bash and zsh, errexit on and off — because POSIX has the shell ignore
> `-e` inside a compound command whose status is consumed by `||`, `&&` or `if`,
> which is the only way to read that status without the failure taking the shell
> down. A probe that cannot fail is worse than no probe.

The same file defines `fail`, so every check in every block ends
`|| fail '<what was being proved>'` and no block depends on errexit at all. The
two halves cover each other: the guard catches the shell, the failure actions
stop a block that reaches a wrong answer even where the guard never ran.

**`docs/RELEASING.md`** — all seven blocks rewritten to that shape, and both tag
refspecs now use `${tag}`.

**`task release-procedure`** (`tests/release/procedure-blocks.mjs`, in `verify`)
— it proves the document rather than reading it:

1. extracts the seven `sh` blocks — five are indented inside a numbered step, so
   a rule anchored at `` ^``` `` finds two of them;
2. requires each block to open with the guard, before anything else runs;
3. requires every `test`, `git`, `gh` and `task` in them to carry a failure
   action;
4. refuses `$name:` anywhere in a block;
5. parses every block in each available shell;
6. **runs the guard**: refused with neither `-e` nor `-u` and saying why on
   stderr, refused with `-e` but no `-u`, silent under `set -eu`, and a failed
   check stopping the block — in sh, bash and zsh.

Its own rule is proved in both directions first: 4 unguarded statements it must
catch, 8 guarded ones it must accept.

## Against the defect

The document as it stands on `main@ba1b38db`:

```
FAIL: release procedure: the block at docs/RELEASING.md:69 does not source release/fail-closed.sh
FAIL: release procedure: docs/RELEASING.md:69: `test "$fetch_urls" = "$release_url"` has no failure action
FAIL: release procedure: docs/RELEASING.md:202: `$tag:` interpolates before a colon, which zsh reads as a
  history modifier -- write `${tag}` (this is how `refs/tags/$tag:refs/tags/$tag` became
  `refs/tags/v0.11efs/tags/v0.11.0-seed`)
... 63 findings
```

and after:

```
PASS: the block rule catches 4 unguarded statements and accepts 8 guarded ones
PASS: 7 docs/RELEASING.md blocks source release/fail-closed.sh first, and all 91 of their statements fail closed
PASS: all 7 blocks parse in sh, bash, zsh
PASS: release/fail-closed.sh refuses a shell without `set -eu` and stops a failed check, in 3 shell(s): sh, bash, zsh
```

A mangled block is caught too — deleting one closing quote in step 7 produces
`docs/RELEASING.md:242 is not valid sh: unexpected EOF while looking for
matching '"'` in all three shells.

## What it deliberately does not enforce

Stated in the gate's header rather than left to be discovered: a command that is
neither `test` nor `git`/`gh`/`task`, and an assignment whose `$(...)` does not
close on its own line, are exempt. `actual_assets=$(for asset in ...; do` closes
three lines later, and a line-wise rule would either demand a failure action
inside a `for` body or need a shell parser to know better. The document writes
those with `|| fail` anyway; the gate does not police them.

## Bookkeeping

- census regenerated with `--count`: two new files, and `Taskfile.yml` moves
  from 7 `node` invocations to 8
- `release-procedure` added to `verify` and to its help group
- evidence pack rebound — `Taskfile.yml` is one of its declared inputs

## Validation

`task release-procedure` PASS (all four sections) · `task task-help` PASS ·
`task gate-reachability` PASS at 168 of 186 · `task repository-check` PASS ·
`task backlog` PASS · census checker and self-test PASS · `task release-claims`
PASS.

The full local `task verify` is being re-run and reported below. The first
attempt was spoiled by orphaned build processes from a run I had killed —
`cp: ... build/stage2/kofun-stage2: Text file busy` inside `examples`, which is
contention and not a result. Recording that rather than quietly rerunning is
what step 1 of the procedure this PR is about asks for.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -1246,6 +1246,17 @@ tasks:
     cmds:
       - "sh tests/release/check-claims.sh"
 
+  # The procedure is a document, and a document cannot be run -- so the checks
+  # inside it were trusted rather than checked, and one of them printed PASS for
+  # a tag that did not exist (#1603). This extracts every `sh` block from
+  # `docs/RELEASING.md`, requires each to open with the fail-closed guard and
+  # every check in it to carry its own failure action, and then *runs* the guard
+  # in each available shell instead of only finding it.
+  release-procedure:
+    desc: "Prove docs/RELEASING.md's blocks cannot report PASS while failing"
+    cmds:
+      - "node tests/release/procedure-blocks.mjs"
+
   # The release lane: regenerate the reviewable pack. `release-claims` then fails
   # until the regenerated pack is committed, so the two cannot diverge.
   release-evidence:
@@ -1442,4 +1453,5 @@ tasks:
             assertions \
             backlog \
             release-claims \
+            release-procedure \
             rfc-registry
--- a/artifacts/release-evidence/index.json
+++ b/artifacts/release-evidence/index.json
@@ -785,7 +785,7 @@
   ],
   "evidence_digests": {
     ".github/workflows/native-hosts.yml": "06bee8496d78319c7102ea89c83e270baa6ecbd6039cc3486e1573819488f3e2",
-    "Taskfile.yml": "00638a3bebed08b931c6219b99e36be462c9a03eeb00b0020a4b5971bb428314",
+    "Taskfile.yml": "df9047aa4993ba1d1bd20d1339249863aa77b5313ab460f9409a93e1ff04cb8b",
     "bootstrap/native/README.md": "6c470b203c3dbc327c6b601ab958c754e6f82be78c255d442307ad0c7691bc21",
     "bootstrap/native/check-macho64-signed.sh": "6fc51ddfec599ea85b8b06a5c16d68b6fdc994621cdb8ca4267e20be38141393",
     "bootstrap/native/check-macho64.sh": "d6bd52e3f6756aa607a2dc44a0117defeaaa55e64fa3f4221986da6406251e15",
--- a/docs/RELEASING.md
+++ b/docs/RELEASING.md
@@ -63,24 +63,49 @@ Run the whole procedure in one POSIX shell, beginning with `set -eu`; every
 command block below assumes that same fail-closed shell. Use a clean checkout
 of `main`, with the working tree clean. Bind every fetch and push to the
 intended repository explicitly; do not assume that a remote named `origin` is
-`kofun-lang/kofun`:
+`kofun-lang/kofun`.
+
+**That instruction is now checked rather than trusted, in both directions.**
+Cutting `v0.11.0-seed`, step 7's block printed `PASS: remote tag ...` while no
+such tag existed on the remote: it had been pasted into a shell where `set -eu`
+was not in effect, so every `test` in it was advisory and the summary ran
+anyway. So each block opens by sourcing `release/fail-closed.sh` — from the
+checkout root, where the whole procedure runs — which refuses a shell missing
+`e` or `u` and defines `fail`, and every check in every block ends
+`|| fail '<what was being proved>'`. The guard catches the shell; the failure
+actions mean a block that reaches a wrong answer stops there even if the guard
+never ran. `task release-procedure` proves both halves, and runs the guard
+rather than only looking for it (#1603).
+
+Read the exit status of each block, not the last line it printed. Step 1 makes
+the same point about a pipeline ending in `tail`; it applies to the blocks
+themselves:
 
 ```sh
 set -eu
+. ./release/fail-closed.sh || exit 1
 release_repo=kofun-lang/kofun
 release_remote=release-target
 release_url=ssh://git@github.com/kofun-lang/kofun
 if ! git remote get-url "$release_remote" >/dev/null 2>&1; then
-    git remote add "$release_remote" "$release_url"
+    git remote add "$release_remote" "$release_url" ||
+        fail "adding the $release_remote remote"
 fi
-fetch_urls=$(git remote get-url --all "$release_remote")
-push_urls=$(git remote get-url --push --all "$release_remote")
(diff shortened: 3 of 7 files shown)


hjosugi:
Full local `task verify` on a quiet tree: **1631 PASS lines, no FAIL**, and no `Text file busy` — the contention that spoiled the first attempt is gone. The count is 1627 → 1631 against `main`, which is `release-procedure`'s four sections arriving.

CI at this head is green as well: Kofun verification 24m34s, native host evidence binding, release evidence pack.

Both directions of the new gate, for the record:

```
$ git stash && node tests/release/procedure-blocks.mjs 2>&1 | grep -c '^FAIL'
63
$ node tests/release/procedure-blocks.mjs
PASS: the block rule catches 4 unguarded statements and accepts 8 guarded ones
PASS: 7 docs/RELEASING.md blocks source release/fail-closed.sh first, and all 91 of their statements fail closed
PASS: all 7 blocks parse in sh, bash, zsh
PASS: release/fail-closed.sh refuses a shell without `set -eu` and stops a failed check, in 3 shell(s): sh, bash, zsh
```

PR #1610, opened 2026-08-23
release: the block rule checks by default rather than by name (#1603)

hjosugi:
The refactor pass over #1609, which found one thing worth fixing and it is not
cosmetic.

## The gate had a silent gap

`tests/release/procedure-blocks.mjs` asked "does this statement carry a failure
action?" of four command names — `test`, `git`, `gh`, `task` — and of nothing
else. `docs/RELEASING.md` already contained three statements outside that list:

```sh
cd "$asset_dir"
"$digest_tool" -c "$source_sums"
printf '%s\n' 0.3.50-seed >VERSION
```

The first enters the directory every following check reads. The second is the
digest verification that the whole of step 7 exists to perform. The third is how
the version is set. The gate reported the file clean while walking past all
three — and would have walked past the next verb the procedure grows.

That is the same failure mode as the defect #1603 was filed for: a check that
reports success over ground it never looked at.

## The rule inverts

Everything a block runs is checked. The exemptions are things that cannot report
a wrong answer:

| exempt | why |
|---|---|
| shell keywords and the closing halves of constructs | `fi`, `done`, `)` run nothing |
| an assignment that runs nothing | `x=literal`, `x=${y%%[[:space:]]*}` cannot fail |
| `printf`/`echo` **without** a redirection | output is not a check — but a redirection is: a full disk while writing `VERSION` is exactly the failure this document used to print PASS through |
| an assignment whose `$(...)` does not close on its own statement | `actual_assets=$(for asset in ...; do` is three lines from its `done`; demanding a failure action inside a `for` body would be demanding the wrong thing |

An allowlist of checked names stops covering the file the moment the file grows
a verb. An allowlist of exemptions has to be argued for one entry at a time, and
each of the four carries its argument in the source.

**`docs/RELEASING.md` needed no change.** It already passes the stricter rule —
the document was written to the standard the gate was not yet enforcing, which
is the useful half of this measurement.

## Cases

4 → 8 catches, 8 → 12 acceptances. The new catches are the three statements
above plus a `curl` that no allowlist would have contained; the new acceptances
are `set -eu`, `fi`, `)`, and a plain string assignment.

```
PASS: the block rule catches 8 unguarded statements and accepts 12 guarded ones
PASS: 7 docs/RELEASING.md blocks source release/fail-closed.sh first, and all 91 of their statements fail closed
PASS: all 7 blocks parse in sh, bash, zsh
PASS: release/fail-closed.sh refuses a shell without `set -eu` and stops a failed check, in 3 shell(s): sh, bash, zsh
```

Also drops `export` from the three functions, which nothing imports — the cases
that exercise them are in the same file, and an export with no importer reads as
an interface someone may rely on.

## Validation

`task release-procedure` PASS in all four sections · census unchanged, so no new
row and no count moved · `task release-claims` PASS · `task gate-reachability`
PASS. Full local `task verify` running and reported below; CI at this head is
the gate.

### DIFF
--- a/tests/release/procedure-blocks.mjs
+++ b/tests/release/procedure-blocks.mjs
@@ -21,24 +21,29 @@
  *   1. every `sh` block sources `release/fail-closed.sh` before it runs
  *      anything else, so a shell that would swallow failures is refused at the
  *      first line rather than at the summary;
- *   2. every `test` and every top-level `git`, `gh` or `task` in those blocks
- *      carries its own failure action, because the guard alone cannot help a
- *      wrapper that starts a fresh shell per command -- which is what the
- *      measurement below says actually happened;
+ *   2. everything those blocks run carries its own failure action, because the
+ *      guard alone cannot help a wrapper that starts a fresh shell per command
+ *      -- which is what the measurement below says actually happened;
  *   3. no `$name` is interpolated immediately before a colon;
  *   4. every block parses -- nothing else in the tree asks whether this
  *      document's shell is shell, and a mangled quote would otherwise be found
  *      by the operator mid-release;
  *   5. the guard is executed, in both directions, rather than merely located.
  *
- * WHAT IT DOES NOT ENFORCE, said here rather than left to be discovered: a
- * command that is neither `test` nor one of those three names, and an
- * assignment whose `$(...)` does not close on its own line, are not required to
- * carry a failure action. `actual_assets=$(for asset in ...; do` opens a
- * command substitution that closes three lines later, and a rule that read it
- * line by line would either demand a failure action in the middle of a `for`
- * body or need a shell parser to know better. The document writes those with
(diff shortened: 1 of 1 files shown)


hjosugi:
Full local `task verify` on this branch: **1631 PASS lines, no FAIL**, same as the run on #1609's tree — expected, since the document did not have to change to satisfy the stricter rule.

`task release-procedure` alone, both directions:

```
PASS: the block rule catches 8 unguarded statements and accepts 12 guarded ones
PASS: 7 docs/RELEASING.md blocks source release/fail-closed.sh first, and all 91 of their statements fail closed
PASS: all 7 blocks parse in sh, bash, zsh
PASS: release/fail-closed.sh refuses a shell without `set -eu` and stops a failed check, in 3 shell(s): sh, bash, zsh
```

PR #1611, opened 2026-08-23
census: read shell command position instead of matching it (#1552)

### LINKED ISSUE #1552: The shell census treats wrapper prose and mixed wrapper chains as commands
## Metadata

- State: ready
- Kind: test-quality
- Area: Tooling / forbidden-requirements census
- Priority: P2
- Size: M
- Parent: #1451
- Blocked by: none. #1535 landed in #1574 (`main@b3e58d0d`).
- Audited against: `main@a858e0fa6f0060c4ecc2d0299852f4282014a9b5`

## Goal

Replace the shared shell detector's unanchored wrapper shortcut with one bounded,
quote-aware command-position model that composes executable wrappers without
confusing shell builtins with external executables.

This owns the broader grammar deliberately excluded from #1535. That issue adds
the narrow single-wrapper `--` spelling first; this issue removes the legacy
false positives and makes mixed wrapper chains explicit after #1535 lands.

Measured on exact clean `main@a858e0fa6f0060c4ecc2d0299852f4282014a9b5`:

```sh
node --input-type=module <<'NODE'
import { commandPosition } from './tooling/forbidden-requirements/detect.mjs'

for (const [name, body, token] of [
  ['prose wrapper', 'echo env node', 'node'],
  ['quoted wrapper', "printf '%s\\n' 'command node'", 'node'],
  ['command/env chain', 'command env -- node', 'node'],
  ['env/xargs chain', 'env xargs -- npm', 'npm'],
  ['xargs/env chain', 'xargs env -- npm', 'npm'],
  ['env cannot invoke builtin command', 'env -- command node', 'node'],
  ['xargs cannot invoke builtin command', 'xargs -- command node', 'node'],
]) {
  console.log(name, (body.match(commandPosition(token)) ?? []).length)
}
NODE
```

Observed results:

```text
prose wrapper 1
quoted wrapper 1
command/env chain 0
env/xargs chain 0
xargs/env chain 0
env cannot invoke builtin command 1
xargs cannot invoke builtin command 1
```

The first two are invented invocations. The next three are real invocations
that are missed. The final two are also invented: external `env` and `xargs`
cannot resolve the shell builtin `command`, but the unanchored inner
`command node` branch counts them.

The live census has one measured manifestation. The Node invoke detector finds
two hits in `tests/conformance/backends/wasm32-node.sh`:

```text
command -v node
exec node
```

The second is line 24 inside a single-quoted generated-script literal, not an
invocation by this tracked shell process. A correct repair is therefore expected
to change that file's Node invoke count from 2 to 1 while retaining its row for
the real `command -v node` use. Re-measure this expectation against the post-#1535
main before moving the issue to `ready`.

No duplicate was found in open or closed issues or pull requests. #1500 and
#1528 are adjacent detector corrections; #1535 owns only the immediate,
single-wrapper terminator spelling.

### In scope

- replace the unanchored ordinary shell wrapper shortcut with a bounded lexical
  command-position model;
- keep ordinary no-terminator wrapper detection and the #1535 terminator cases;
- compose wrapper states so shell `command` wrappers may precede external
  `env|xargs` wrappers, while an external wrapper cannot turn `command` into a
  shell builtin;
- distinguish operators and wrapper words in executable shell text from those
  inside single, double, and ANSI-C quoted prose;
- treat escaped newlines as continuations rather than new command positions;
- handle command substitutions and `case` pattern boundaries either exactly or
  fail closed for an explicitly documented unknown region;
- add direct positive/negative fixtures for the measured matrix above; and
- regenerate the census only for measured semantic count changes.

### Out of scope

- a complete POSIX/Bash parser, runtime expansion, glob resolution, or dynamic
  command names;
- guessing long wrapper options or options that consume a following argument;
- changing JavaScript child-process or YAML command detection;
- changing requirement classifications, the native-toolchain contract, or CI
  workflow behavior; and
- touching the canonical Stage 2 pair or release claims.

## Acceptance criteria

- [ ] `echo env node` and quoted `command node` prose count zero.
- [ ] `command env -- node`, `env xargs -- npm`, and `xargs env -- npm` each
      count exactly once.
- [ ] `env -- command node` and `xargs -- command node` count zero.
- [ ] ANSI-C quoted prose, escaped operators/newlines, command substitutions,
      and a `case` pattern fixture have explicit load-bearing expectations.
- [ ] All existing ordinary-wrapper controls and #1535 terminator controls stay
      green.
- [ ] The `wasm32-node.sh` Node invoke count changes from 2 to 1, its real
      `command -v node` row remains, and every other census change is explained.
- [ ] No JavaScript/YAML detector, Stage 2 pair file, contract profile, workflow,
      or release claim changes.

## Validation

| --- | --- | --- |
| focused lexical and wrapper-state cases | `node tooling/forbidden-requirements/self-test.mjs` | every measured positive and near miss has the exact count |
| live tree vs ledger | `node tooling/forbidden-requirements/check.mjs` | exact agreement in both directions |
| reviewable counts | `node tooling/forbidden-requirements/check.mjs --count` | only measured semantic changes |
| focused gate | `task forbidden-requirements-census` | PASS |
| repository | `task verify` | PASS |

## Dependencies

Blocked by: #1535.

Both issues edit `tooling/forbidden-requirements/detect.mjs` and its direct
self-test. Re-measure current behavior and the census after #1535 closes, then
promote this issue to `ready` if its premise still holds.

hjosugi:
Closes #1552.

## The matrix, re-measured on today's main

Every row of the issue's evidence still held at `main@951a2c79`:

```
prose wrapper                        1   invented
quoted wrapper                       1   invented
command/env chain                    0   missed
env/xargs chain                      0   missed
xargs/env chain                      0   missed
env cannot invoke builtin command    1   invented
xargs cannot invoke builtin command  1   invented
```

The first two are the unanchored `wrapper\s+` shortcut matching the word `env`
wherever it appears — including inside a single-quoted string that is the *text
of a script this one generates*. The middle three are wrapper chains the pattern
had no way to compose. The last two are the shortcut in the other direction:
external `env` and `xargs` cannot resolve the shell builtin `command`, so what
runs there is `command`, and `node` is its argument.

## What replaces it

The text is tokenised once, and command position becomes a property of where a
word sits rather than of what precedes it in the string.

- **Quoting**: single and ANSI-C quoted regions are opaque and stay inside the
  word containing them; double quotes are *not* opaque, because `"$(rustc -Vv)"`
  runs rustc — reading the whole span as text lost three real invocations in
  `examples/rust-shim/benchmark.sh`.
- **Wrapper chains** compose in the order the shell can resolve: a shell wrapper
  may precede an external one, never the reverse.
- **`case`** gets the three states it actually has — reading its subject,
  reading a pattern, running a body — because patterns sit exactly where a
  command would.
- **Not modelled, by decision**: long options, options that consume the
  following argument, runtime expansion, dynamic command names. Each returns
  "no invocation here" rather than a guess, and each has a case.

## Two defects the old rule survived only by keeping no state

**`withoutComments` was not quote-aware.** It cut at any `#` after whitespace,
including inside a string:

```sh
assert_grep "row $item does not name issue #$owner in its own header"
```

which leaves an unterminated quote. Line by line the damage stopped at the line.
To a scanner it swallowed the rest of the file — every `rustc` invocation in
`tests/usability/check.sh` after line 122 went uncounted. Comment ranges now come
from the tokeniser, so there is one implementation of what a quote is.

**Here-document bodies were read as shell.** One apostrophe in English prose
inside one opens a quoted region running to the next apostrophe, wherever that
is. Bodies are skipped now — which is also the right answer for the case that
started this issue: a heredoc holding the text of a script is not this script
running it.

## Every census row that moved, and why

| file | change | why |
|---|---|---|
| `tests/conformance/backends/wasm32-node.sh` | node 2 → 1 | the issue's named case: `'exec node …'` is the text of a generated script. Its real `command -v node` row remains |
| `stdlib/date_time/tests/verify.sh` | node 3 → 2 | `differential='… (node is not installed)'` — `(` opened a command position inside a quoted assignment |
| `tests/release/check-claims.sh` | go-task row removed | ``"… \`task release-evidence\`"`` — escaped backticks are text, not a substitution |
| `bootstrap/selfhost/check-compiler-driver.sh` | cc 12 → 11 | `for candidate in ${KOFUN_SANITIZER_CC:-} clang gcc cc` — `}` opened a command position, so a `for` list element counted as an invocation |
| `tests/interop/bindgen-c/check-sanitizers.sh` | cc 6 → 5 | same shape |
| `tests/rfc/check-registry.sh` | node 10 → **11** | **a use the old rule could not see**: it allowed no indentation between the line start and an assignment, so `        NAME="$x" node …` matched nothing |

The last row is the direction that matters. Five are false positives leaving; the
sixth is a real dependency that was never recorded, and a ledger that only ever
shrinks is not being read in both directions.

## API

`commandPosition` becomes `matchCommands(body, pattern, dialect)` — it returns
the matches rather than a regular expression, because the shell half no longer
has one to return. JavaScript and YAML keep their patterns unchanged, as the
issue scopes them out.

The one consumer outside this directory is `tests/digest/no-host-digest-tools.mjs`
(#1213), which passes the **two-word** pattern `openssl[ \t]+dgst`. A model
assuming a pattern matches a single word would silently stop refusing
`openssl dgst`, so both directions of that case are pinned in the self-test.

## Cases

24 command-position cases, each owning one fact rather than folding into an
aggregate — three kinds of quoting, wrapper chains in both compositions, `case`
patterns against `case` bodies, here-document bodies against what follows them,
escaped operators and escaped newlines, `${x}` expansion against `{ x; }`
grouping, and two-word spellings in both directions. All 18 existing wrapper and
terminator controls stay green.

## Validation

`node tooling/forbidden-requirements/self-test.mjs` PASS · checker PASS in both
directions over 676 rows · `task digest` PASS · `task gate-reachability` PASS ·
full local `task verify` **1632 PASS, no FAIL, exit 0**.

No JavaScript or YAML detector, Stage 2 pair file, contract profile, workflow, or
release claim changed.

### DIFF
--- a/tests/digest/no-host-digest-tools.mjs
+++ b/tests/digest/no-host-digest-tools.mjs
@@ -23,7 +23,7 @@
  * the tool would be a rule about the wrong thing.
  *
  * Both halves of the scan are the census's, reused rather than re-written. The
- * detector is `commandPosition`, which already knows that `$(sha256sum` is an
+ * detector is `matchCommands`, which already knows that `$(sha256sum` is an
  * invocation and that `SHA256SUMS` and `nodejs-flavour`-shaped tokens are not,
  * and which #1535 widened for wrapper `--` terminators; the file set is
  * `universe()`. This gate first shipped with its own copy of the second one,
@@ -35,14 +35,14 @@
  * One digest in the tree is deliberately outside this rule, and it is filed
  * rather than left silent: the Windows lane of the same workflow computes its
  * digest with PowerShell's built-in `Get-FileHash`, which this scan cannot see
- * because `commandPosition` reads shell, JavaScript, and YAML `run:` position
+ * because `matchCommands` reads shell, JavaScript, and YAML `run:` position
  * and not PowerShell. That is also a different dependency — a built-in of the
  * host being tested rather than a userland tool someone installs — and whether
  * `bin/kofun-digest` can run there at all is unmeasured. #1606 carries both
  * questions, so a passing run here means five lanes, not six.
  */
 
-import { commandPosition, dialectOf, withoutComments } from '../../tooling/forbidden-requirements/detect.mjs'
+import { dialectOf, matchCommands, withoutComments } from '../../tooling/forbidden-requirements/detect.mjs'
 import { universe } from '../../tooling/forbidden-requirements/universe.mjs'
 
 /* `pattern` is a regular-expression fragment; `spelling` is what a message says. */
@@ -64,7 +64,7 @@ function invocationsIn(path, body) {
     const source = withoutComments(path, body)
     const dialect = dialectOf(path)
     return TOOLS.flatMap(({ spelling, pattern }) =>
-        (source.match(commandPosition(pattern, dialect)) ?? []).map((text) => ({ spelling, text })),
+        matchCommands(source, pattern, dialect).map((text) => ({ spelling, text })),
     )
 }
 
--- a/tooling/forbidden-requirements/census.tsv
+++ b/tooling/forbidden-requirements/census.tsv
@@ -99,7 +99,7 @@ cc	invoke	3	required-today	required	bootstrap/c_abi/check.sh
 cc	invoke	4	required-today	required	bootstrap/native/check.sh
 cc	invoke	1	required-today	required	bootstrap/selfhost/build-a1-a2.sh
 cc	invoke	5	required-today	required	bootstrap/selfhost/c11/check-c11.sh
-cc	invoke	12	required-today	required	bootstrap/selfhost/check-compiler-driver.sh
+cc	invoke	11	required-today	required	bootstrap/selfhost/check-compiler-driver.sh
 cc	invoke	2	required-today	required	bootstrap/selfhost/check-diverse-double-compilation-refusals.sh
 cc	invoke	2	required-today	required	bootstrap/selfhost/check-driver-diagnostics.sh
 cc	invoke	6	required-today	required	bootstrap/selfhost/check-profile.sh
@@ -193,7 +193,7 @@ cc	invoke	1	required-today	required	tests/fuzz/value_if.sh
 cc	invoke	2	required-today	required	tests/fuzz/visibility-artifacts.sh
 cc	invoke	1	required-today	optional	tests/http/check.sh
 cc	invoke	2	required-today	required	tests/interfaces/visibility-filtering.sh
-cc	invoke	6	required-today	required	tests/interop/bindgen-c/check-sanitizers.sh
+cc	invoke	5	required-today	required	tests/interop/bindgen-c/check-sanitizers.sh
 cc	invoke	4	required-today	required	tests/interop/bindgen-c/check.sh
 cc	invoke	1	required-today	required	tests/interop/bindgen-c/import-boundary/run.sh
 cc	invoke	2	required-today	required	tests/module-constants/check.sh
@@ -272,7 +272,7 @@ node	invoke	7	required-today	required	spec/wasi-command-projection-v1/check.sh
 node	invoke	5	required-today	required	spec/wasi-host-matrix-v1/check.sh
 node	invoke	15	required-today	required	spec/wasm-host-abi-v1/check.sh
 node	invoke	6	required-today	required	spec/wasm-host-profile-v1/check.sh
-node	invoke	3	required-today	optional	stdlib/date_time/tests/verify.sh
+node	invoke	2	required-today	optional	stdlib/date_time/tests/verify.sh
 node	invoke	4	required-today	required	stdlib/decimal/tests/verify.sh
 node	invoke	8	required-today	required	tests/artifact-qualification/check.sh
 node	invoke	1	required-today	required	tests/backlog/check-stamps.sh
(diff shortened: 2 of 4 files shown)


hjosugi:
Amended with the refactor pass over this change, before review rather than after:

- **Cost, measured and stated.** Reading costs more than matching: the checker goes 0.9s → 2.6s and `task digest`'s scan 0.20s → 0.86s over the same tree. Most of what that could have been is already gone — the callers ask fourteen detectors about one file in a row, so a **single-entry token cache** takes the checker from 4.7s to 2.6s. Two passes remain per file, one to find comment ranges and one over the stripped text; collapsing those would mean threading tokens through an interface four callers share, for about a second and a half of wall time.
- Unified the two `esac` branches in the `case` machine into one, and dropped a dead line from `addToWord`.

One thing this change surfaces and does not fix, worth its own issue if it looks right: `tooling/machine-dependent/detect.mjs` has its **own** `withoutComments`, with the same `#`-inside-a-string defect this PR fixes here. Its file set is `*.sh *.mjs *.js`, so the same shell quoting applies. Fixing it means measuring that ledger's counts, which is a separate change with its own explanations.

Issue #1628, opened 2026-08-29
Three tracked programs Stage 2 accepts emit C that does not compile


hjosugi:
## Metadata

- State: needs-detail
- Tracker role: curated implementation issue
- Kind: implementation
- Area: Compiler / Stage 2 emission
- Priority: P1
- Size: M
- Parent: #1539
- Blocked by: none
- Audited against: `origin/main@3d0e2a2f69667acb5914a82d46ef357b14703c07`

## Problem

#1539 states the property: *if `bin/kofun build` does not refuse a program with
an `error[...]` diagnostic, then the C it wrote must compile.* It also predicted
that checking it over the tracked corpus **"finds nothing new, because those
files already build in some gate."**

**That prediction is false, and this issue is the three counterexamples.**

## Current behavior and evidence

Measured on `3d0e2a2f69667acb5914a82d46ef357b14703c07` over all 1,320 tracked
`*.kofun`: 358 build, 940 are refused by name with `error[`, and 22 reach the
host C compiler. Nineteen of the 22 are fixtures owned by another driver — the
native, wasm and selfhost-A1 corpora, plus the Stage 1 seed, which need their
own translation units and are the wrong invocation rather than a defect.

**Three are not.** These are ordinary tracked programs the default
`bin/kofun build` path owns:

```sh
bin/kofun build tests/conformance/inference/hm-levels/recursion.kofun -o /tmp/o
# recursion.c:235:11: error: 'k_b1' undeclared (first use in this function)
# recursion.c:212:16: error: 'kofun_lambda_1' defined but not used [-Werror=unused-function]

bin/kofun build tests/conformance/syntax/issues_48_60/token-spans.kofun -o /tmp/o
# token-spans.c:227:13: error: unused variable 'k_b0' [-Werror=unused-variable]

bin/kofun build tests/typed-sidecar/fixtures/stage2_events.kofun -o /tmp/o
# stage2_events.c:240:20: error: unused variable 'k_b3' [-Werror=unused-variable]
```

The first is the sharp one and is not a warning-flag artifact: **the emitter
wrote a reference to a variable it never declared**, and `kofun_lambda_1` is
emitted and never called. That is the same class as #1622's `lambda_scope` — a
name read where nothing binds it — but on the emission side rather than in the
Kofun half's own text.

The other two are `-Werror=unused-variable` on a binding the emitter created and
did not use. Whether the fix is to stop emitting the dead binding or to stop
[...]

Issue #1622, opened 2026-08-25
#1401 region: hold the Kofun half to bindings that exist, not just calls


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation child
- Kind: test-quality
- Area: Bootstrap / Quality
- Priority: P1
- Size: M
- Parent: #1401
- Blocked by: none
- Audited against: `aa1445d58f3d7d7737854a9eea5d8bf5f092ef22`

## Goal

One gate, the sibling of `tests/pair-coverage/calls.sh`: **every name read in
`bootstrap/stage2/compiler.kofun` resolves to something that binds it**, with a
ledger of the exceptions that fails in both directions.

## Current behavior and evidence

`tests/pair-coverage/calls.sh` resolves every *call target* in the Kofun half
and records the one that does not in `unresolved-calls.tsv`. Its own header
states what it does not cover, and a bare read is on that list by omission: it
reads "only lowercase-initial call targets", which is what `fn` names and
builtins are.

Nothing else covers it either. `bootstrap/stage2/check.sh` round-trips the file
and byte-compares the identity projection, which resolves no names. The gates
that compare the halves — `selfhost-generations`, `selfhost-fixed-point` —
compare emitted C, and only the C half emits.

The hole is not hypothetical. Measured on
`aa1445d58f3d7d7737854a9eea5d8bf5f092ef22`, which is `main` before PR #1621:

```sh
grep -n '\blambda_scope\b' bootstrap/stage2/compiler.kofun
# 13541:                        lambda_scope,
# 13556:                    lambda_scope + "|" + token_text(source, name) +
```

Two reads, no binding anywhere in 23,676 lines. The C half declares it at the
top of the same lambda walk in `build_scope_hir_mode`. It survived because the
only thing that resolves names in this file — `kofun-stage2 --compile-outcome`
— refuses 12,635 lines earlier on `E2S35: lexical use limit is 256 per
function` (#1483), so nothing ever reached line 13541.

## Why it matters

#1483 will eventually make `--compile-outcome` succeed on the Kofun half, and
that *is* the total check. This issue is worth doing before then for the reason
`calls.sh` was: it costs seconds instead of minutes, it runs on every pull
request, and it names the defect rather than reporting a refusal 12,000 lines
away from it. When #1483 lands, this gate keeps the property from regressing
without re-running a four-minute self-compile.

## Scope

- Read every identifier token in the Kofun half that is a *read*, not a call
[...]

Issue #1596, opened 2026-08-21
Four derived artifacts went stale the same way in one day, and none of the causing changes noticed


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated planning issue
- Kind: spec-design
- Area: Tooling / derived artifacts
- Priority: P2
- Size: M
- Blocked by: none
- Audited against: `origin/main@0d3423cedeb5837d176485e932411be4f17e5d09`

## Problem

**Four times in one day, a derived artifact went stale because its basis moved
underneath it, and in every case the thing that noticed was a slow gate or a
red `main` rather than the change that caused it.**

These were treated as four separate gate bugs. They are one shape.

| artifact | what invalidated it | what noticed | when |
| --- | --- | --- | --- |
| `tests/pair-coverage/inputs.tsv` | new `*.kofun` files from #1315, #1503, #1321, and two stdlib modules | `task preflight`, after #1523 shipped it; **nothing in `verify` reaches it** | drift reached 22 files before anyone measured |
| `tests/pair-coverage/drivers.tsv` | #1321 added `task bytes-mutation` to `verify` and never pinned it | `measure.sh` refusing to start | at the top of a 2.5-hour measurement |
| `artifacts/release-evidence/index.json` | any edit to a pinned file — `Taskfile.yml`, `stdlib/tests/verify.sh`, `bootstrap/manifest.json`, `spec/**` | the 9-second CI lane, on the **next** PR as often as the causing one | at least four times today |
| the pair ledger + mirror map | #1548 renamed `bytes_parameter_discards` → `unused_parameter_discards` | `stage2-pair-mirror`, by **turning `main` red** | #1595 is the fix |

## The shape, stated once

A derived artifact records a fact about the tree at some commit. Its basis — the
set of inputs it was derived from — is **not** checked against the tree by the
thing that changes the tree. So:

1. a change edits an input;
2. the change is green, because the artifact's gate is slow, outside `verify`,
   or not reached by that PR's file set;
3. the artifact is now describing a tree that no longer exists;
4. the cost lands on **somebody else** — the next PR, the next measurement, or
   whoever pushes to `main` next.

The last row is the sharpest instance and shows why this is not a discipline
problem. **#1548 was green against the `main` it forked from**, and
`83e2be26` refreshed the ledger after that fork. Neither PR was wrong; the pair
of them was. A PR is tested against the `main` it forked from, not the one it
lands on, and no amount of care by either author closes that.

## Proposed direction

[...]

PR #1629, opened 2026-08-29
emission: check that every program Stage 2 accepts emits C that compiles (#1539)

### LINKED ISSUE #1539: Nothing checks that a program Stage 2 accepts emits C that compiles

hjosugi:
Closes #1539. Found #1628.

The property, stated in one place so a future gate can cite it rather than
restate it:

> **If `bin/kofun build` does not refuse a program with an `error[...]`
> diagnostic, then the C it wrote must compile.**

Nothing stated it and nothing checked it. Every gate that compiles emitted C
does so over hand-written fixtures; the grammar fuzzer generates *malformed*
programs and never invokes a host C compiler; `selfhost-generations` and
`selfhost-fixed-point` compare the two halves' output, so a defect in both is
invisible by construction; and none of this class is an untaken branch, so
`stage2-pair-coverage` cannot bound it.

## #1539 predicted this would find nothing

Its own words: shape A "finds nothing new, because those files already build in
some gate." Measured on `3d0e2a2f` over all **1,320** tracked `*.kofun`:

```
358 built    940 refused by name    22 neither    249s serially
```

Nineteen of the 22 are the wrong invocation rather than a defect — native, wasm
and selfhost-A1 fixtures whose owning gates supply their own translation units,
plus the Stage 1 seed, whose emitted C includes `kofun_unicode.c`.

**Three are not**, and they are ordinary programs the default driver owns:

| file | what `cc` says |
| --- | --- |
| `tests/conformance/inference/hm-levels/recursion.kofun` | `error: 'k_b1' undeclared (first use in this function)`; `kofun_lambda_1` emitted and never called |
| `tests/conformance/syntax/issues_48_60/token-spans.kofun` | `error: unused variable 'k_b0' [-Werror=unused-variable]` |
| `tests/typed-sidecar/fixtures/stage2_events.kofun` | `error: unused variable 'k_b3' [-Werror=unused-variable]` |

The first is not a warning-flag artifact: **the emitter wrote a reference to a
variable it never declared** — the same class as #1622's `lambda_scope`, on the
emission side. Filed as **#1628** and recorded as `emission-defect` rows, so this
lands green without anyone pretending they are fixed.

Each of the three is exercised today by a gate that reads something other than a
compiled binary — a token span, a typed sidecar projection, an inference trace.
That is this issue's argument, now with instances instead of an argument.

## Acceptance criteria

- [x] **The property is stated in one place** — the gate header, cited by the
      ledger rather than restated in it.
- [x] **A gate checks it over the tracked corpus, shown to fail by
      reintroducing a defect** — `--prove` runs the real classifier over two
      real tracked files, one that builds and one that violates.
- [x] **The generated half's design question is answered before implementing
      it**: it is an axis on the existing fuzz generator, not a third one, and
      it is *not* in this change. The measurement supplies what that design was
      missing — a per-program cost of **0.19 s** — so its bound can be argued in
      seconds rather than in cases, which is the problem `grammar.sh`'s own
      comment says any answer here inherits.
- [x] **The gate distinguishes "refused by name" from "host compiler said no"**
      — that distinction *is* the classifier, and the two are the same exit
      status today.

## Affordability — what `needs-detail` was actually asking

249 s against a `verify` that runs ~17 minutes, so the task sits **outside
`verify`** with a row in `tooling/gate-reachability/unreachable.tsv` stating
that cost. Parallelising was considered and rejected: `bin/kofun build` writes
`<stem>.c` into one shared `build/bootstrap-stage1/`, so concurrent builds race
on output paths.

## Validation

| Check | Result |
| --- | --- |
| `task emission-compiles` | `PASS: 1298 of 1320 ... 22 recorded`; `8 of 8 cases behaved as required` |
| `task stage2-pair-decisions` | PASS (landed in #1627) |
| `sh tests/preflight/check.sh` | PASS |
| `node tooling/gate-reachability/check.mjs` | PASS |
| `node tooling/task-help.mjs --check` | PASS |
| `node tooling/forbidden-requirements/check.mjs` | PASS |
| `sh tests/release/check-claims.sh` | PASS |

The must-not-fire case is proved **first**: a gate that refused everything would
satisfy all seven refusals and be useless. The denominator is proved too — an
empty corpus refuses rather than reading as green.

## Pair lane

No Stage 2 pair file is written, so the lane held by the live claim on #1516 is
untouched.

https://claude.ai/code/session_01S7UbJQZTYMc8HbpGwuxuUb


hjosugi:
## Fixup pushed: the fifth obligation

The first head was refused by `Kofun verification` for an obligation the other
four did not include:

```
FAIL: assertion budget: tests/emission/check.sh has 1 silent assertions and no budget entry
```

The line read the corpus with stderr thrown away and no handler, which under
`set -e` aborts with an empty stderr — and it was silencing the one condition
this gate must never swallow, an unreadable corpus. **Fixed rather than
budgeted**: the budget's own header says the plan is to drive every count to
zero (#814), so landing a new non-zero row would add slack to a file whose
purpose is to remove it.

```
PASS: every script is at its recorded silent-assertion budget (0 remaining, 56 at zero)
```

Worth recording separately: `task preflight` reported three of this task's
obligations in one run and **not** this one — its checks do not include the
assertion budget, so a new shell script still costs one CI lane to discover.
That is the same argument #1596 made about `drivers.tsv`.

## One thing a reviewer should know about this PR's CI

`emission-compiles` is deliberately **outside `task verify`** (249 s), so the
`Kofun verification` lane does not run it. Its evidence is the local run at this
head, stated rather than implied:

```
$ sh tests/emission/check.sh
PASS: 1298 of 1320 tracked program(s) build or are refused by name;
      22 recorded in tests/emission/exceptions.tsv

$ sh tests/emission/check.sh --prove
8 of 8 cases behaved as required
```

Issue #1513, opened 2026-08-20
Stage 2 pair: the C half escapes non-ASCII identifiers and the Kofun half does not


hjosugi:
## Metadata

- State: in-progress
- Tracker role: curated implementation issue
- Kind: implementation
- Area: Bootstrap / Stage 2 pair
- Priority: P1
- Size: M
- Parent: #1401
- Blocked by: none
- Audited against: `origin/main@41e7464d5ad0edb8741b25aa948e531a02e75ef2` — re-measured 2026-09-21; the decision gate passes: `19 one-property mutations each fail at its owning boundary`
- Normative profile: `spec/stage2-pair-host-primitives-v1/PROFILE.md`
- Decision merged by: #1617

## Goal

Make both halves of the canonical Stage 2 pair emit identical C for non-ASCII identifiers and refuse every path alias that names the input before any write.

## Recorded decision

Option A is selected. The compiler receives two compiler-private host operations:

- `stage2_unicode_scalar_at` decodes one validated UTF-8 scalar for the identifier escaper;
- `stage2_same_file` follows the final symlink and compares `st_dev` plus `st_ino`, returning a typed lookup refusal.

Neither operation is callable by ordinary user code. This does not create a program-visible byte API; #1499 remains that owner. The exact visibility, authority, result, and refusal rules are fixed by the normative profile.

## Implementation scope

- Add both private operations to the C and Kofun compiler halves.
- Route all enumerated identifier declaration/reference/parameter/call funnels through the same `k_uXXXXXX` spelling.
- Refuse same-text, hard-link, symlink, and normalized aliases before changing the input.
- Update the pair checksums, bootstrap manifest, and release evidence in one serialized change.
- Do not widen the public Core builtin surface or narrow Unicode syntax.

## Acceptance criteria

- [x] Both halves emit byte-identical C for the full non-ASCII identifier funnel inventory.
- [x] Removing scalar handling from either half makes the pair gate fail.
- [x] Both halves refuse every input alias before a write, with lookup-error precedence covered.
- [x] Removing either identity check makes the pair gate fail.
- [x] `task unicode`, `task stage2`, and `task selfhost-fixed-point` pass.
- [x] Pair digests, manifest, and evidence are regenerated by their owning commands.

## Validation

- Decision contract: `task stage2-pair-host-primitives-decision`
- Implementation gate: an executable two-sided pair fixture covering every semantic funnel and alias class
[...]

Issue #1312, opened 2026-08-11
Benchmark report v1: implement the canonical Bytes codec and negative corpus in Kofun


hjosugi:
**Completed by PR #1646 at `46284ad2984600a69175c255023d49dcedde0543`; all acceptance criteria verified.**

## Metadata

- State: in-progress
- Parent: #847
- Kind: implementation
- Area: Core Standard Library / Benchmarking
- Priority: P2
- Size: M
- Integration dependency: PR #1635 (the lexical-use-budget slice of #1483); this codec exceeds the previous 256-use limit. #1322 is merged as PR #1641.
- Audited baseline: `origin/main@04003d9b153a09967723a0313016b7a78e55d2cd`

## Goal

Encode and decode exact canonical `kofun.bench-report/v1` Bytes in production
Kofun from/to #1311's validated report outcome, consuming the completed shared
Managed Bytes[65536] profile. This child is a pure Bytes/model codec; it does
not open, rename, publish, or reserve a filesystem name.

## Exact codec profile

- Encode #1310's fixed order/escaping/integer/framing transactionally for a
  caller-provided proven-unique Bytes destination. Work remains private until
  complete. Success installs exactly one validated <=16,384-byte buffer. Every
  error or preexisting BR011 input outcome is a typed non-success, exposes no
  report bytes as success, and preserves the destination's exact prior length/
  capacity/pointer/bytes; no accepted prefix becomes visible. The codec reads
  no ambient or in-flight cancellation token; BR011 is decided before encoder
  invocation and maps directly to the no-bytes plan.
- Wire input over 16,384 bytes is BR004 before parser allocation. A valid
  canonical output is at most 16,384 bytes inside the 65,536-byte shared Bytes
  ceiling, so Bytes-capacity failure is unreachable in this codec. OOM is the
  only reachable allocation failure.
- Decode accepts one read-borrowed complete Bytes and returns the explicit
  49-field success/error outcome. It preserves raw order and every identity,
  budget, measurement, summary, counter availability/value, digest, host, and
  noise field. No borrowed view escapes.
- Failure precedence is exact: wire size BR004; BOM/UTF-8/JSON syntax/trailing
  BR001; complete decoded-key depth scan with excessive depth BR004 superseding
  duplicates; duplicate BR002; schema/scalar/cross-invariant BR003..BR007; then
  canonical-byte mismatch BR002.
- Parser limits are <=16,384 bytes and container depth 16 below root. Samples
[...]

PR #1635, opened 2026-09-21
stage2: raise the lexical use budget from 256 to 4096 (#1483)

hjosugi:
## Summary

Part of #1483 — the part #1513 does not gate. Does **not** close it.

The one bound `compiler.kofun` violates is `E2S35 lexical use limit is 256 per
function`, in six functions (peaks 315 / 377 / 495 / 785 / 1345 / 286, measured
2026-08-25 on the issue). This raises that budget to 4096 in both halves of the
pair and records why, which is acceptance criterion 5 for the only bound that
needed a decision.

## The decision: raise, not restructure

**What the bound is for.** It backs no array — `use_count` is a plain counter
gating how many `use|` records the scope-HIR walk appends to a growable
`Buffer`. What it caps is **time**: each use re-scans the function from
`function_open` through the classification helpers (`lambda_scope_open`,
`parent_block_open`, `const_parameter_of_type`, …), so the walk is quadratic in
a function's uses. Measured with gprof on the fixture shape:

| uses per function | scope-HIR walk |
|---:|---:|
| 256 | 0.05s |
| 4096 | 12s |

That is the cost the issue asked a raised bound to name. No memory changes.

**Why not split.** Splitting six functions — `lower_body` into six pieces — to
fit a budget that backs nothing but that walk would be an arbitrary cut by
construction, in both halves, under all the pair gates. Any of those functions
may deserve decomposition on its own merits; that is a design change to the
pair and not this bound's reason.

**Why 4096.** Three times the measured peak (1345), and the `USE_LIMIT` that
`bootstrap/stage2/hm_levels_frontend.c` already chose for the same count. The
other three lexical budgets (depth 32, scopes 256, bindings 256) are untouched;
nothing exceeds them.

## What it measures

With the bound gone, the committed compiler — not an instrumented scratch build
— refuses its own source at #1513's byte:

```
$ ./kofun-stage2 --compile-outcome bootstrap/stage2/compiler.kofun out.c out.ir out.tokens
error[E2S16]: unknown Core function `is_xid_start` at byte 754
exit=1     (608s wall at -O2)
```

Criteria 2–4 of #1483 stay behind #1513, and every stage after scope-HIR
construction stays *unreached*, not satisfied.

## The commits, in order

1. **`stage2: raise the lexical use budget from 256 to 4096`** — `256 → 4096`
   at the three guard sites in each half, one comment beside `use_count`
   verbatim in both recording the above; two rows in
   `tests/pair-coverage/decisions.tsv` hold the guard and its message text to
   the same number; `tests/conformance/modules/lexical-scopes` moves its use
   boundary to 4096/4097 (now ~40s for the two cases, which its README names);
   digests, manifest, evidence regenerated.
2. **`pair-coverage: the corpus timeout must outlast the pair's own Kofun
   half`** — a cost the first version did not name. `compiler.kofun` is a
   pinned corpus input for the two-compiler ledger, and under its `-O0
   --coverage` build it ran 371s when the old bound refused it at line 906 and
   runs **1080s** now that the whole scope-HIR stage precedes the `is_xid_start`
   refusal. The 600s per-input limit timed it out and the harness refused to
   report, as written. `measure.sh` moves to 2400 with both numbers beside it;
   `tooling/machine-dependent/ledger.tsv` re-pins the bound as a `backstop`.
3. **`lexical-scopes: reach the resolved-use refusal on purpose`** — the first
   full measurement read `build_scope_hir_mode` 27 → 28: the resolved-use
   `> 4096` guard had gone untaken, because `compiler.kofun` tripping it at
   line 906 was the only thing that ever reached it. `generate_uses` reaches
   only the unresolved-assignment site. One binding, 4096 resolved assignments,
   a read as the 4097th use (~12s) reaches it deliberately.
4. **`pair-coverage: re-measure the ledger at 78e75b41`** — two full runs, 214
   min for the pair. The header separates what the two #1483 runs isolate
   (that one row, back to 27, every other row identical between them) from
   what a single measurement over the span since `80e45fc7` cannot attribute
   (five rows entered, two left, three changed — by name, the September
   positional-move and Bytes-identity commits that did not re-measure).
5. Merge of `origin/main` at `42e2c29a` (#1515), which landed mid-run and is
   **not in the ledger's basis** — said so in the header. #1640 deferred the
   ledger to "whoever lands last"; that is the run after this one, and the
   next `check.sh` will read its three new functions. Digest conflicts
   resolved by regenerating.
6. **`pair-coverage: the mirror map follows the re-measured ledger`** — five
   `same-name` rows in, two out.

## Verification

Local, at `78e75b41` (the measured tree) unless noted:

- `task verify`: exit 0 at `7beeebec`; CI green at `d1a9dc8f` and `78e75b41`
- `stage2-pair-coverage`: `PASS: the Stage 2 pair ledger is exact: 1987 untaken
  under gcc, 1892 under clang, across 370 function(s)`, and `prove.sh` 10/10
- `task stage2`, lexical-scopes fixture, `stage2-pair-calls` / `-bindings` /
  `-mirror` / `-decisions` with `--prove`, `machine-dependent-bounds`,
  `preflight`, `check-claims`, `repository-check`, the eight cheap pair-reading
  gates: PASS
- At the merged head `52958106`: seed builds clean, digests verify, all four
  pair ledgers + `machine-dependent` + `preflight` + `check-claims` PASS;
  `task verify` running locally alongside CI


hjosugi:
Read-only integration review of #1635 at `52958106`, against main `1dfe3934`: compiler sources auto-merge, but six derived/ledger files conflict. Main already has a newer two-host ledger at `91d353f2` (372 functions, 2003/1909), so taking this PR's `undefended.tsv` / `mirror.tsv` would discard #1513 evidence. Preserve main's `xid-scalar-decoding` decision plus the two new use-budget rows, regenerate checksums/manifest/release evidence, and remeasure the final integrated ledger under the serialized pair owner. Two small review follow-ups: the new comments incorrectly call the depth budget 256 (it is 32), and the preserved-pattern-candidate guard still lacks a deliberate 4096/4097 boundary fixture. No competing pair work or heavy run started; `codex-takeover-20260923`'s #1641 → #1635 order is respected.

hjosugi:
FYI for the #1635 evidence handoff: the frozen `eff76476f8` measurement remains valid for its committed **173-driver / 1344-input** basis. Please retain that exact commit/basis in the refreshed ledger header and keep the raw receipt. No additional full coverage run is requested.

Later integration changes the existing benchmark model/corpus; #1312 adds `benchmark-report-codec` and `codec.kofun`; #1220 changes the compiler and adds `concurrency-hir` (generated cases, no new tracked `.kofun`). Reconcile the current pins and normal required gates, and state that these later changes are outside the frozen coverage basis. The refreshed baseline should not be relabelled as final-head coverage merely because its rows were merged.

The explicit #1220/#1312 validation lists do not require a fresh complete coverage run for each PR; preflight validates current list membership and ordinary CI excludes the expensive coverage task. Your existing frozen-baseline plan is compatible with that boundary. The independent read-only audit records the literal references at `audit-scoped-hir-1220-20260923/build/pair-coverage-integration-audit/REPORT.md`.

